PR #1, opened 2021-10-22
feature/ DAO-390 

sepehr2github:
- react-i18next added
- configuration files added


packages/web-app/react-i18next.d.ts
```diff
@@ -0,0 +1,13 @@
+import { resources, defaultNS } from './i18n.config';
+// react-i18next versions lower than 11.11.0
```
githubdoramon: If we are using the higher one, doesn't make sense to have the module declared twice right? You can remove this part

packages/web-app/tsconfig.json
```diff
@@ -15,6 +15,6 @@
     "noEmit": true,
     "jsx": "react"
   },
-  "include": ["src"],
+  "include": ["src", "react-i18next.d.ts"],
```
githubdoramon, line 18: Why is this needed?

packages/web-app/i18n.config.ts
```diff
@@ -0,0 +1,17 @@
...
+import { initReactI18next } from 'react-i18next'
+
+export const resources = {
+  en: {
+    translation: en
+  }
+} as const
+
+i18n.use(initReactI18next).init({
+  lng: 'en',
```
githubdoramon: I remember that 18next have a fallback property (so it uses the fallback translation if it does find the key. Is this still a thing?

githubdoramon (changes requested):
@sepehr2github also important to add tests for validating translation keys

packages/web-app/i18n.config.ts
sepehr2github: Yes, We can define fallbacks too. but because we didn't decide on some details yet. I just did basic configs.

packages/web-app/react-i18next.d.ts
sepehr2github: Sorry, About this one!. It's happened by mistake. I removed that but I don't know why It was on my PR

WissenIstNacht:
I would suggest a different folder structure here (also used by i18next creators in their examples):

```bash
src/
├─ i18n/
│  ├─ en/
│  │  ├─ translation.json
│  ├─ de/
│  │  ├─ translation.json
```

I would also be ok with `i18n` being called `locales`.

PR #20, opened 2021-10-28
Edit eslint and prettierrc based on Google's TypeScript style guide DAO-424 

sepehr2github:
config eslint
config prettierrc

### DIFF
--- a/packages/web-app/.eslintrc
+++ b/packages/web-app/.eslintrc
@@ -15,7 +15,6 @@
     "ecmaVersion": 11,
     "sourceType": "module",
     "ecmaFeatures": {
-      // Allows for the parsing of JSX
       "jsx": true
     }
   },
--- a/packages/web-app/.eslintrc
+++ b/packages/web-app/.eslintrc
@@ -52,9 +52,6 @@
     {
       "files": ["**/*.ts", "**/*.tsx"],
       "parser": "@typescript-eslint/parser",
-      "extends": [
-        "plugin:@typescript-eslint/recommended"
-      ],
       "rules": {
         "@typescript-eslint/no-non-null-assertion": "off",
         "@typescript-eslint/no-use-before-define": "off",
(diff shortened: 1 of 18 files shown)


githubdoramon (approved):

packages/web-app/.eslintrc
```diff
@@ -10,6 +10,15 @@
       "version": "detect"
     }
   },
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": 11,
+    "sourceType": "module",
+    "ecmaFeatures": {
+      // Allows for the parsing of JSX
```
nivida: This inline comment is not really required because it is self-describing :)

packages/web-app/.prettierrc
```diff
@@ -1,6 +1,6 @@
 {
-  "trailingComma": "none",
-  "semi": false,
+  "bracketSpacing": false,
   "singleQuote": true,
-  "printWidth": 80
-}
+  "trailingComma": "es5",
+  "arrowParens": "avoid"
+}
```
nivida: EOF

nivida (changes requested):

Fabricevladimir (approved):

packages/web-app/.prettierrc
WissenIstNacht: While we're at it: 

- `trailingComma` is the same as default setting and could therefore be removed.
- @nivida AFAIK Prettier does not have a EOF setting. Instead, its behavior is to always add one line at the end of file.

WissenIstNacht: Also, is the `arrowParens` based on the google style guide? (I couldn't find information on this in the style guide)

sepehr2github: > Also, is the `arrowParens` based on the google style guide? (I couldn't find information on this in the style guide)

reference:
https://github.com/google/gts/blob/main/.prettierrc.json

PR #22, opened 2021-11-03
Feature/components DAO-421 

Fabricevladimir:
- TSDX react-storybook template used


packages/ui-components/.storybook/preview.js
```diff
@@ -1,9 +1,5 @@
...
-  actions: { argTypesRegex: "^on[A-Z].*" },
-  controls: {
-    matchers: {
-      color: /(background|color)$/i,
-      date: /Date$/,
-    },
-  },
-}
\ No newline at end of file
+  actions: {argTypesRegex: '^on.*'},
```
WissenIstNacht: Could you elaborate on this change?
Would this not match too many (potentially unwanted) strings?

WissenIstNacht (approved):

packages/ui-components/example/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
```
WissenIstNacht: Do we need an explicit `example` in this package? Isn't storybook in itself already an example?

packages/ui-components/.storybook/preview.js
Fabricevladimir: This wasn't an intentional change. It does match too many yes

packages/ui-components/example/index.html
Fabricevladimir: Hmm. This does seem unnecessary.

sepehr2github (approved):

PR #27, opened 2021-11-08
Feature/ DAO-434 

sepehr2github:
- Popover added to the storybook
- Radix UI library added

### DIFF
--- a/packages/ui-components/src/components/popover/popover.tsx
+++ b/packages/ui-components/src/components/popover/popover.tsx
@@ -67,11 +67,12 @@ export const Popover: React.FC<PopoverProps> = ({
 
 type StyledContentProps = {
   width: number | string;
+  style?: CSSProperties | undefined;
 };
 
 const StyledContent = styled(Content).attrs(
-  ({width = 300}: StyledContentProps) => {
-    const style: CSSProperties = {
+  ({width = 300, style}: StyledContentProps) => {
+    const currentStyle: CSSProperties = style || {
       width,
       background: '#FFFFFF',
       boxShadow: '0px 0px 4px rgba(50, 63, 75, 0.16)',
@@ -80,6 +81,6 @@ const StyledContent = styled(Content).attrs(
       margin: 8,
     };
 
-    return {style};
+    return {style: currentStyle};
   }
 )<StyledContentProps>``;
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,84 @@
...
+   */
+  content: ReactNode;
+  /**
+   * Content Width
+   */
+  width: number | string;
+}
+
+/**
+ * Primary UI component for user interaction
```
Fabricevladimir: I suggest updating this comment.

packages/ui-components/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,84 @@
...
+  );
+};
+
+type StyledContentProps = {
+  width: number | string;
+};
+
+const StyledContent = styled(Content).attrs(
+  ({width = 300}: StyledContentProps) => {
+    const style: CSSProperties = {
```
Fabricevladimir: Could you refactor this to use tailwind classes instead of react styles?

packages/ui-components/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,84 @@
...
+   */
+  open?: boolean;
+  /**
+   * Event handler called when the open state of the popover changes.
+   */
+  onOpenChange?: (open: boolean) => void;
+  /**
+   * The preferred popover position
+   */
+  position: {
```
Fabricevladimir: Is it possible to flatten position? The rationale being that nested we lose the comments for side and align as well as the type of controls that storybook would generate.

WissenIstNacht (commented):
Hey nice work so far, looks good.

I have a general question/remark about this PR, though. It relates to the "ownership" or "responsibilities" of some of the components. Let me try to explain:

The Popover component accepts `children` and `content.` As far as I understand, the `children` are used for the triggering logic, while the `content` is what is actually displayed in the popover. Cool idea, so far so good. At the same time you define a `StyledContent`, which acts as the white popover bubble. And the you add the content into there. What I want to get at here is that on the one hand you allow for customization (i.e. give ownership/responsibility over the content to the caller of the `Popover` component), yet you "force" that content to be inside this `StyledComponent` (thereby giving ownership/responsibility over the content to the `Popover` component itself). This could get awkward if someone passes a div because they think, for example, that they can change the background. Intuitively I would either let the caller handle every aspect of the content (including styling), or I would do everything in the component itself, limiting the caller to just calling it and maybe specify some text.

Let me know what you think!

WissenIstNacht (approved):

Fabricevladimir (approved):

WissenIstNacht (approved):

Fabricevladimir (approved):

PR #29, opened 2021-11-08
Chore / DAO-437 

Fabricevladimir:
- Re-exported testing-library/react with custom `render` function that includes a Router and internalization provider
- Changed path to the test-utils in tsconfig so utilities can be referenced as follows: ```import {screen, render} from 'test-utils'```

### DIFF
--- a/packages/web-app/jest.config.js
+++ b/packages/web-app/jest.config.js
@@ -4,5 +4,10 @@ module.exports = {
   collectCoverage: true,
   collectCoverageFrom: ['src/**/*.ts(x)'],
   setupFilesAfterEnv: ['<rootDir>/.jest/setup.ts'],
-  modulePaths: ['<rootDir>/src/', '<rootDir>/.jest']
-}
+  modulePaths: ['<rootDir>/src/', '<rootDir>/.jest'],
+  moduleDirectories: [
+    'node_modules', // add the directory with the test-utils.js file, for example:
+    'utils', // a utility folder
+    __dirname, // the root directory
+  ],
+};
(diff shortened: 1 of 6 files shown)


packages/web-app/jest.config.js
```diff
@@ -4,5 +4,10 @@ module.exports = {
   collectCoverage: true,
   collectCoverageFrom: ['src/**/*.ts(x)'],
   setupFilesAfterEnv: ['<rootDir>/.jest/setup.ts'],
-  modulePaths: ['<rootDir>/src/', '<rootDir>/.jest']
-}
+  modulePaths: ['<rootDir>/src/', '<rootDir>/.jest'],
+  moduleDirectories: [
+    'node_modules', // add the directory with the test-utils.js file, for example:
```
WissenIstNacht: Thanks for leaving a comment. Unfortunately I don't understand it. Could you explain why `node_modules` is here?

WissenIstNacht (approved):
Thanks for doing this, it will certainly come in handy.

packages/web-app/jest.config.js
Fabricevladimir: Setting `moduleDirectories` overrides Jest's default behavior when it's looking for modules/packages. By adding in `node_modules` in the array, we're instructing Jest to also look into that folder when it's trying to resolve a module/package. You can take a look at the corresponding documentation [here](https://jestjs.io/docs/next/configuration#moduledirectories-arraystring).

WissenIstNacht: Thanks a lot for the explanation 👌🏼

sepehr2github (approved):

sepehr2github:
Please solve the conflicts then merge

PR #31, opened 2021-11-09
[DAO-380] [cicd] adds pipelines for push checks and deployment

mathewmeconry:


.github/workflows/webapp-deploy.yml
```diff
@@ -0,0 +1,66 @@
+name: WebApp Deploy
+on:
+  push:
+    paths:
+      - 'packages/web-app/**'
+    branches:
```
yuetloo, line 6: do we need to trigger this job whenever ui-components changes, since we build it here in this job...

.github/workflows/webapp-deploy.yml
```diff
@@ -0,0 +1,66 @@
...
+        working-directory: ${{env.working-directory}}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build ui-components
+        run: |
+          cd ../ui-components
+          yarn
+          yarn build
+          yarn link
+      - name: Install depdendencies
```
yuetloo, line 40: If we use `lerna` in this monorepo, we can simplify the build steps here, i.e. no need to do `yarn link` manually, lerna will generate the link for dependent libraries, hoisting dependent libraries also speed up build time.  Not sure what the web team think about using lerna.  @WissenIstNacht , any comment on this?

yuetloo (approved):
LGTM in general, just a couple of minor comments which can be addressed later if needed.

.gitignore
```diff
@@ -2,7 +2,8 @@
 node_modules
 
 # testing
-/coverage
+**/coverage
```
WissenIstNacht: only having `coverage` does the trick 👍🏼 (all other PRs already have that in place)

.github/workflows/webapp-deploy.yml
WissenIstNacht: Never set up lerna, so can't really comment on that 😅 

But in theory the linking step is not necessary. If there is a `build` in the `ui-components` package, the `web-app` package should be able to pick up on it. Recall that `ui-components` is actually installed as a dependency in the `web-app` package.

I guess what I'm trying to say is that yarn should take care of all the linking in a workspaces environment. Also, we currently don't have plans to install Lerna.

mathewmeconry: according to https://aragonassociation.atlassian.net/browse/DAO-381 we won't publish the ui-component lib yet. therefore I need to link the package manually to the web-app pacakge

.github/workflows/webapp-deploy.yml
mathewmeconry: the best would be not, because the ui-component should have its own workflow to build and publish it. and usually you don't rebuild your app when a dependency is changed.
Unfortunately Ramon/you won't publish the ui lib yet, therefore I had to do the manual linking later in the work flow

.gitignore
mathewmeconry: reverted

PR #32, opened 2021-11-09
Feature/basic-buttons-DAO-439

WissenIstNacht:
Added buttons according to design system.


packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
-   */
-  backgroundColor?: string;
-  /**
-   * How large should the button be?
-   */
-  size?: 'small' | 'medium' | 'large';
-  /**
-   * Button contents
-   */
+  mode: 'primary' | 'secondary' | 'tertiary' | 'ghost';
```
Fabricevladimir, line 8: Can we get some comments on some of these guys? Functionality for example I'd like to know a bit more about at a glance.

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
+  size: 'small' | 'default';
+  functionality: 'add' | 'open' | 'normal';
```
Fabricevladimir: Is it possible to generalize this? I'd like to be able to add whatever icon I want and maybe have the prop determine where to put it. Something like `<Button icon={abc} right/>` or `<Button iconRight={abc}/>`. What do you think?

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
-  size = 'medium',
+  mode = 'primary',
+  size = 'default',
+  functionality = 'normal',
   label,
-  primary = false,
-  backgroundColor,
-  ...props
+  onClick,
+  disable,
```
Fabricevladimir: Typo? I think it was expecting `disabled` 😄

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
 
-type StyledButtonProps = {
-  backgroundColor?: string;
-  primary: boolean;
-  size: ButtonProps['size'];
+// Content Components ======================================================
+
+type ContentProps = {
+  label: ButtonProps['label'];
+  functionality: ButtonProps['functionality'];
```
Fabricevladimir: Ha! I like this. This is us making TS do the work for a change!

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
+};
+
+/**
+ * Provides the content for the button. The content depends on the Button's
+ * functionality prop. If the functionality is 'add', or 'open', the content
+ * will be a label and an icon. If the functionality is 'normal', the content
+ * will be just a label.
+ */
+const Content: React.FC<ContentProps> = ({label, functionality}) => {
```
Fabricevladimir: Hmm this seems like it could have been achieved with TW + SC, especially if we generalize the button to take any icon. Like change the `order` of the flex box items based on whether we want the icon on the right or left.. Thoughts?

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
+  if (functionality === 'add') {
```
Fabricevladimir: I think this could be simplified to 2 if statements since we're actually returning the JSX.

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
+    return (
+      <FlexDiv className="flex">
+        <AddIcon />
+        {label}
+      </FlexDiv>
+    );
+  } else if (functionality === 'open') {
+    return (
+      <FlexDiv className="flex">
```
Fabricevladimir: We can remove the `className=flex` since FlexDiv already has `display: flex;` applied

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
+};
 
-    const style: any = {
-      ...(size ? variantSizeStyles[size] : variantSizeStyles.default),
-      backgroundColor,
-    };
+/**
+ * Extends the SizedButton element with the desired styling
+ */
+const StyledButton = styled(SizedButton).attrs(({mode}: StyledButtonProps) => {
```
Fabricevladimir: I like the composition! For whatever reason I keep forgetting you can extend an already defined styled component 😅

packages/ui-components/src/components/button/button.tsx
```diff
@@ -2,72 +2,119 @@ import React, {ButtonHTMLAttributes} from 'react';
...
+  let className;
 
-    return {className, style};
+  if (mode === 'primary') {
```
Fabricevladimir: Might consider adding a default so that if the user passes in something random the button is still styled properly. That said, I don't know if the typings will allow them to 🤔

packages/ui-components/stories/Button.stories.tsx
```diff
@@ -5,32 +5,15 @@ import {Button, ButtonProps} from '../src';
...
-Large.args = {
-  size: 'large',
-  label: 'Button',
-};
-
-export const Small = Template.bind({});
-Small.args = {
-  size: 'small',
-  label: 'Button',
+export const AllProps = Template.bind({});
```
Fabricevladimir: Do you think we should have the different states of the button as stories or is that too much? I feel like right now, it's a bit of a hassle to go through all the props in Storybooks canvas layout. That said, the controls are laid out well so...

Fabricevladimir (changes requested):

packages/ui-components/stories/Button.stories.tsx
```diff
@@ -5,32 +5,15 @@ import {Button, ButtonProps} from '../src';
...
+AllProps.args = {
+  label: 'Button Text',
+  disabled: false,
+  onClick: () => {
+    console.log('hi');
```
sepehr2github: Maybe It's better to use Alert('hi') instead of the console.log. I this case user can see the onClick event without checking console

sepehr2github (commented):
Thank you. It looks awesome and you did a great job. I think Fabrice mentioned everything.

packages/ui-components/src/components/button/button.tsx
WissenIstNacht: Since the buttons were grouped together in the designs, I tried to get them all into one interface. That being said, we absolutely can. And since other buttons depend on this, we probably should.

Fabricevladimir: Yeah I'm guessing this is going to be an iterative process too based on the availability of different types of buttons.

packages/ui-components/src/components/button/button.tsx
WissenIstNacht: no, you can't just pass anything. :)

packages/ui-components/stories/Button.stories.tsx
WissenIstNacht: I've been asking myself the same thing. For the time being I'd say we don't invest time in something that's already there. As you said, the controls already allow to see the props in action. If we want to refine it and add proper docs and stuff we always do that later.

packages/ui-components/stories/Button.stories.tsx
Fabricevladimir: @sepehr2github Wasn't this supposed to show up in the action tab in storybook?

sepehr2github (approved):

Fabricevladimir (approved):

PR #36, opened 2021-11-11
Feature/icon DAO-446 

sepehr2github:
Icons added!

Notes: 
1) We still need to add some defined sizes like (small, medium, etc...).
2) Let me know if there is any problem with Exported Svgs


packages/ui-components/src/components/icons/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+export {IconCheckboxSelected} from './components/icon_checkbox_selected';
+export {IconCheckmark} from './components/icon_checkmark';
+export {IconChevronDown} from './components/icon_chevron_down';
+export {IconChevronUp} from './components/icon_chevron_up';
+export {IconChevronRight} from './components/icon_chevron_right';
+export {IconChevronLeft} from './components/icon_chevron_left';
+export {IconClose} from './components/icon_close';
+export {IconCopy} from './components/icon_copy';
+export {IconFavoriteDefault} from './components/icon_favorite_default';
+export {IconFavoriteSelected} from './components/icon_favorite_selected';
```
WissenIstNacht: case typo :) But please change the name of the file to lowercase and _not_ the name of the import to uppercase.

WissenIstNacht (changes requested):
Hey there, thanks for the tedious work. I have a few complaints, though 😅 

Basically, I disagree with the nested `components` folder that you have in both the `icons` and `stories` folder.
- For the one in `icons` I don't really see the point. If you ever open the icons folder you probably intend to look at the icons. So IMO there's not much point in hiding them another level further down 😋 
- For the the one in `stories` I see why you did it. However, I would not call this a `components` folder, as it doesn't contain any components, just import/exports. Also, I would not put those files and corresponding folder into `stories` at all.

I think there's a way to solve both issues at once. You could group the items in `icons` into two subfolders: `interface` and `module`. That way, you have (still) have a subfolder in icons. Then you could basically move the `ìnterface-x-list.tsx` files to those subfolders.

Let me know if I missed something :)

packages/ui-components/src/components/icons/index.tsx
sepehr2github: That's wired. I fixed that before on my VS Code but it didn't apply!!!!

Fabricevladimir (approved):
Looks great, man! Once the changes requested by @WissenIstNacht have been added you can go ahead and merge.

Fabricevladimir (approved):

WissenIstNacht (approved):

PR #37, opened 2021-11-12
Feature/menu buttons dao 442

WissenIstNacht:
This PR:

- Added button with following variations: 
    - No text, just icon
    - Closed menu button (with menu icon)
    - Opened menu button (with cross icon)
    - Small/default sizing
- Fixed some tests
- Modified some svgs to take color and to correct sizing


packages/ui-components/src/components/button/iconButton.tsx
```diff
@@ -46,6 +46,6 @@ type StyledIconProps = {
   iconSrc: IconButtonProps['iconSrc'];
 };
 
-const StyledIcon = styled.img.attrs(({iconSrc}: StyledIconProps) => {
+export const StyledIcon = styled.img.attrs(({iconSrc}: StyledIconProps) => {
   return {src: iconSrc, className: 'w-3 h-3'};
```
sepehr2github: Why did you export StyledIcon? So we can use it as the base of the other Icons button?

sepehr2github (approved):

packages/ui-components/src/components/button/menuButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+export const MenuButton = ({
+  size = 'default',
+  isMobile,
+  isOpen,
+  onClick,
+}: MenuButtonProps) => {
+  if (isMobile) {
+    return (
+      <StyledButton onClick={onClick} size={'default'}>
+        <IconMenu />
```
Fabricevladimir, line 35: Should we also display a close for mobile? Currently we don't actually see this because the bottomsheet covers up the menu, but say a user goes with a 'traditional' mobile that still shows the nav and toggle button, this will only display the open icon.

packages/ui-components/src/components/button/menuButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+      </StyledButton>
+    );
+  }
+
+  if (isOpen) {
+    return (
+      <OpenButton onClick={onClick} size={size}>
+        <FlexDiv side={'left'}>
+          <IconClose className="fill-current text-primary-500" />
```
Fabricevladimir: You can remove the class names I believe because the svgs define fill and should also take the text color of the parent `OpenButton` which I noticed was primary-500 as well

packages/ui-components/stories/MenuButton.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+import {MenuButton, MenuButtonProps} from '../src';
+
+export default {
+  title: 'Components/Buttons/Menu',
+  component: MenuButton,
+} as Meta;
+
+const Template: Story<MenuButtonProps> = args => <MenuButton {...args} />;
+
+export const AllProps = Template.bind({});
```
Fabricevladimir: Can you add a couple stories for the controls, just to be explicit?

packages/ui-components/stories/MenuButton.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+AllProps.args = {
+  isMobile: false,
+  isOpen: false,
+  onClick: () => {
```
Fabricevladimir, line 16: This overrides the default behavior of the action appearing on the action tab on Storybook. Is this intentional?

Fabricevladimir (changes requested):
Looking good!

packages/ui-components/src/components/button/menuButton.tsx
WissenIstNacht: Had the same thought. AFAIK we don't really have the designs for it yet.

packages/ui-components/stories/MenuButton.stories.tsx
WissenIstNacht: Sepehr suggested adding something that was more visible/interactive than the logs on the actions tab on a previous PR. I found the idea good, so I went ahead and did it.

packages/ui-components/src/components/button/iconButton.tsx
WissenIstNacht: I was probably using it at some point but ultimately decided against it. Must have forgotten to remove the export. Thanks for pointing it out 👍🏼

Fabricevladimir (approved):

PR #41, opened 2021-11-17
Chore: Icons & typography

WissenIstNacht:
Changes:

- Imported Manrope font 
- Added 'add' and 'close' icons to the respective buttons.
- Improved size/position of certain icons


packages/ui-components/src/components/button/menuButton.tsx
```diff
@@ -41,7 +41,11 @@ export const MenuButton = ({
     return (
       <OpenButton onClick={onClick} size={size}>
         <FlexDiv side={'left'}>
-          <IconClose />
+          <IconClose
+            width={12}
+            height={12}
+            className="fill-current text-primary-500"
```
Fabricevladimir: Should `fill-current` be passed down? Also, the SVG should automatically take on the text color as well.

packages/ui-components/src/styles/tailwind.css
```diff
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Manrope:wght@200;300;400;500;600;700;800&display=swap');
```
sepehr2github, line 1: I think It's better to use a local font instead of using a remote link. What do you think?

sepehr2github (commented):
Thanks.
Looks Great!

Fabricevladimir (approved):
Looks great!

packages/ui-components/src/styles/tailwind.css
WissenIstNacht: Agreed! Can I suggest waiting for the design tokens to come in and tackle this in a corresponding ticket?  (use this as a temporary fix so we don't have to demo serif fonts 😋)

sepehr2github (approved):

PR #47, opened 2021-11-22
Feature: Dao Card

WissenIstNacht:
This PR 

- implements,
- adds a storybook entry,
- tests
- and documents

the `DaoCard` component.

### DIFF
--- a/packages/ui-components/src/utils/addresses.tsx
+++ b/packages/ui-components/src/utils/addresses.tsx
@@ -1,17 +1,19 @@
+export type Address = string;
+
 // get truncated address
-export function BeautifyLabel(label: string | null) {
-    if (label === null) return '';
-    if(IsAddress(label))
-      return (
-        label.substring(0, 5) +
-        '...' +
-        label.substring(label.length - 4, label.length)
-      );
-    else return label
-  }
-  
-  // check label type
-  export function IsAddress(address: string | null) {
-    const re = /0x[a-fA-F0-9]{40}/g;
-    return Boolean(address?.match(re));
-  }
+export function shortenAddress(address: Address | null) {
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/utils/addresses.tsx
```diff
@@ -1,17 +1,21 @@
...
\ No newline at end of file
+export function shortenAddress(address: Address | null) {
+  if (address === null) return '';
+  if (IsAddress(address))
+    return (
+      address.substring(0, 5) +
+      '...' +
+      address.substring(address.length - 4, address.length)
+    );
+  else return address;
```
sepehr2github, line 14: I selected "label" instead of address because sometimes It can be an ENS! But Ens is a kind of address. I guess :)))

sepehr2github (changes requested):
Great Work. Thanks
Can you add a wide prop for the Dao card?

packages/ui-components/src/utils/addresses.tsx
WissenIstNacht: Ok! So then if you do `shortenAddress("blaaaaaaaa.blaaaaaaa.aragon.ens")` you would get `'blaa...ens`?

If that's true, then I can simply add the `EnsAddress` type to the argument, no problem.

Fabricevladimir (approved):
Looks good to me!

RakeshUP (approved):

sepehr2github (approved):

packages/ui-components/src/utils/addresses.tsx
sepehr2github: hmm... No, I think it should be 'blaaaaa....'

PR #51, opened 2021-11-23
Feature/Badge component DAO-466

RakeshUP:
- Badge component added
- New color palettes added in tailwind config


packages/ui-components/src/components/badge/badge.tsx
```diff
@@ -0,0 +1,32 @@
...
+  } else if (colorScheme === 'red') {
+    colorCode = 'bg-critical-200 text-critical-800';
+  } else {
+    colorCode = 'bg-ui-50 text-ui-600';
+  }
+
+  const className: string = `text-xs p-0.5 rounded ${colorCode}`;
+
+  return { className };
+}) <any>``;
```
WissenIstNacht:
```suggestion
}) <StyledBadgeProps>``;
```

packages/ui-components/src/components/badge/badge.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import styled from 'styled-components';
+
+export type BadgeProps = {
+  /** Changes a badge's color scheme */
+  colorScheme?: 'default' | 'green' | 'red';
```
WissenIstNacht, line 6: Personally, I would have named the types after the function rather than the color (i.e., "success", "critical", etc.). Imo, this would more nicely reflect the underlying color scheme defined by the design system.
I realize it can be awkward to call the badge with the `critical` props in some situation. But overall I think consistency is more important. 

Let me know if you dis-/agree :)

WissenIstNacht (changes requested):
Thanks for your first PR, looking good! :)

packages/ui-components/src/components/badge/badge.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
Fabricevladimir, line 1: Would you mind adding a does it render test for the component?

Fabricevladimir (dismissed):
Good work, good work!

packages/ui-components/src/components/badge/index.tsx
```diff
@@ -0,0 +1 @@
+export * from './badge';
```
sepehr2github, line 1: Based on prettier in all ts/tsx files, It's better to have an empty line at the end of the files.
Because there is no tsx here It's better to use .ts file instead of tsx

sepehr2github (commented):
Well Done! Thanks

packages/ui-components/src/components/badge/badge.tsx
RakeshUP: I am having the same thought too. And in this particular sense of the badge representing the token price change, it also makes sense to name "green" as "positive" and "red" as "negative". I was not able to decide on the convention, so left it up as the color name itself

WissenIstNacht (approved):

Fabricevladimir (approved):

PR #57, opened 2021-11-25
Feature/EnsName DAO-483

sepehr2github:
- Ens Name added
- Ens Avatar added
- add IsConnect attribute to wallet button

Note1: I couldn't check the ens name/avatar because there are no ens belonging to me! I need your help to know how to test that
Note2: About passing IsConnect prop I'm not sure It's the best way to implement it or not! because we should implement the same condition in the label prop too So let me know if you have a better idea 
(For example, passing two different labels for connecting and not connecting status).


Fabricevladimir (approved):

RakeshUP (approved):
Looks good, but didn't test. We'll merge it into develop and people with ENS names will help us

PR #58, opened 2021-11-25
Feature/end of navbar - DAO-431 

Fabricevladimir:
Nabar
- refactored to use ui-components
- populate mobile nav menu
- added breadcrumbs for all paths save dashboard ("/") and not found routes

menuButton
- focus and active state
- show text on tablet up

daoCard
- onClick applied to card when switcher not present
- option to include switcher
- updated corresponding story
- switchLabel property for translation

daoSelector
- Changed avatar to large
- Temporarily changed text to bold until proper typography has been added

walletButton
- added padding on tablet above

Packages installed
babel-preset-vite - For import.meta in test environment
react-router-breadcrumbs-hoc - For breadcrumb generation


packages/ui-components/src/components/popover/popover.tsx
```diff
@@ -77,10 +84,10 @@ const StyledContent = styled(Content).attrs(
...
       borderRadius: 12,
-      padding: '20px 16px',
-      margin: 8,
     };
 
     return {style: currentStyle};
   }
-)<StyledContentProps>``;
+)<StyledContentProps>`
+  ${({side}) => side && contentMargins[side]}
```
WissenIstNacht, line 92: No need to address this, just curious: Couldn't this have been passed as className in StyledContent?

packages/ui-components/src/components/menuItem/menuItem.tsx
```diff
@@ -37,9 +37,11 @@ export const MenuItem: React.FC<MenuItemProps> = ({
 
 type ContainerProp = {isSelected: boolean};
 const Container = styled.button.attrs(({isSelected}: ContainerProp) => ({
-  className: `flex items-center px-1.5 py-2 space-x-1.5 ${
-    isSelected ? 'text-primary-500 bg-ui-0' : 'text-ui-600'
-  } focus:outline-none focus:ring-2 focus:ring-primary-500 active:text-primary-500 hover:text-ui-800 focus:bg-ui-0 rounded-xl`,
+  // focus:outline-none focus:ring-2 focus:ring-primary-500 TODO: investigate outline
```
WissenIstNacht: still necessary?

packages/ui-components/src/components/daoCard/daoCard.tsx
```diff
@@ -11,11 +11,15 @@ export type DaoCardProps = {
...
+  /** Determines if switch button is displayed and sets the card to clickable */
+  includeSwitch?: boolean;
   /**
    * Allows the Dao Card component grow horizontally, thereby moving the switcher
    * button right.
    * */
-  wide: boolean;
+  wide?: boolean;
+  /** Label for switch button */
+  switchLabel: string;
```
WissenIstNacht: Thanks for catching this. I often forget about the locales.

packages/ui-components/src/components/button/iconOnlyButton.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
```
WissenIstNacht, line 1: Thanks for the addition!

packages/web-app/src/components/breadcrumbMenuDropdown/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  selected,
+  onMenuItemClick,
+}) => {
+  return (
+    <Container>
+      <NavLinksContainer>
+        <NavLinks
+          selected={selected}
+          onClick={onMenuItemClick}
+          isMobile={true}
```
WissenIstNacht, line 18: Confused about this one. Does this (keeping it constant true) work because this is only relevant when the viewport is mobile anyway?

WissenIstNacht (approved):
Last but very much not least. Thanks and kudos on that one. There's a couple of lines of code I'll have to get more familiar with. But as far as I can tell, everything works as expected.

packages/web-app/src/components/breadcrumbMenuDropdown/index.tsx
Fabricevladimir: This is a naming issue. The `isMobile` prop is perhaps incorrect because what that checks is whether an `actionItem` or a `menuItem` component is rendered. A possible solution would be passing in the rendered component instead of delegating the process to `Navlinks`. So in place of the `isMobile` (which is factually incorrect) we would have a `component` prop that would take a `actionItem` or a `menuItem`. Thoughts?

packages/ui-components/src/components/menuItem/menuItem.tsx
Fabricevladimir: The investigation? Yes!

packages/ui-components/src/components/popover/popover.tsx
Fabricevladimir: Yes, I suppose so. I'll take a look at that.

packages/ui-components/src/components/menuItem/menuItem.tsx
WissenIstNacht: lol didn't see the call for investigation at end of the line (wonder why... 😅)

packages/web-app/src/components/breadcrumbMenuDropdown/index.tsx
WissenIstNacht: I'd have to think about it more. Let's postpone that, though.

packages/ui-components/src/components/actionItem/actionItem.tsx
```diff
@@ -60,9 +58,10 @@ export const ActionItem: React.FC<ActionItemProps> = ({
 
 type ContainerProps = {isSelected: boolean; wide: boolean};
 const Container = styled.button.attrs(({isSelected, wide}: ContainerProps) => ({
+  // TODO: Investigate outline - focus:outline-none focus:ring-2 focus:ring-primary-500
   className: `${wide && 'w-full'} rounded-xl py-1.5 px-2 text-base ${
     isSelected ? 'text-primary-500 bg-primary-50' : 'text-ui-600 bg-ui-0'
-  } focus:outline-none focus:ring-2 focus:ring-primary-500 hover:text-primary-500 active:bg-primary-50 disabled:text-ui-300 disabled:bg-ui-0`,
+  } hover:text-primary-500 active:bg-primary-50 disabled:text-ui-300 disabled:bg-ui-0`,
```
sepehr2github: If needed can you do the same for wallet button

sepehr2github (approved):
Well Done Fabrice :))))))

PR #59, opened 2021-11-25
feature/Token Card Component DAO-465

RakeshUP:
Token card component implemented

### DIFF
--- a/packages/ui-components/__tests__/token.test.tsx
+++ b/packages/ui-components/__tests__/token.test.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { TokenCard } from '../src';
+
+describe('TokenCard', () => {
+  function setup(args: any) {
+    render(<TokenCard {...args} />);
+    return screen.getByTestId('daoCard');
+  }
+
+  test('should render without crashing', () => {
+    const element = setup({});
+    expect(element).toBeInTheDocument;
+  });
+});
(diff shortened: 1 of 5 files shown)


packages/ui-components/__tests__/token.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import { TokenCard } from '../src';
+
+describe('TokenCard', () => {
+  function setup(args: any) {
+    render(<TokenCard {...args} />);
+    return screen.getByTestId('daoCard');
```
Fabricevladimir: This test isn't passing. I suspect it might be because of the test id being `daoCard` rather than `tokenCard`.
While it's not part of this task, if you could take a look at the Badge test as well, it would be very appreciated!

Fabricevladimir (changes requested):
Nice work!

packages/ui-components/__tests__/token.test.tsx
WissenIstNacht: There's a fix for the badge test that on one of my PRs 👍🏼

WissenIstNacht (changes requested):
Hey the current card looks great!

"Small" issue though (pun intended 😉): You might have missed the mobile version of it. It's not very different, but for example it it hides the grey label and shrinks avatar and typography.

RakeshUP:
@WissenIstNacht Added the mobile version of the component. I missed checking the design system. Thanks for fixing the test case!

WissenIstNacht:
All good 👌🏼

I like the `ToggleMobileVisibility` component. Will factor it out next time I need something like this :)

WissenIstNacht (approved):

Fabricevladimir (approved):
I had missed the mobile as well 😅 . Nice work

PR #60, opened 2021-11-26
Feature: Typography dao 486

WissenIstNacht:
This PR:

- Installs and configures the fluid type plugin for TW to ui-components and web-app.
- Adds an additional text-size configuraiton to TW.
- Updates the the text sizes in ui-components.
- fixes the failing badge test
- Refactors the (internal) button props

### DIFF
--- a/packages/ui-components/src/components/button/addButton.tsx
+++ b/packages/ui-components/src/components/button/addButton.tsx
@@ -6,14 +6,13 @@ import {FlexDiv} from './iconButton';
 
 /** Button to add stuff. Comes with a plus sign icon  */
 export const AddButton: React.FC<ButtonProps> = ({
+  label,
   mode = 'primary',
   size = 'default',
-  disabled = false,
-  label,
-  onClick,
+  ...props
 }) => {
   return (
-    <StyledButton mode={mode} onClick={onClick} size={size} disabled={disabled}>
+    <StyledButton mode={mode} size={size} {...props}>
       <FlexDiv side={'left'}>
         <IconAdd />
         <p>{label}</p>
--- a/packages/ui-components/src/components/button/button.tsx
+++ b/packages/ui-components/src/components/button/button.tsx
@@ -1,15 +1,19 @@
-import React, {ButtonHTMLAttributes} from 'react';
+import React from 'react';
 import styled from 'styled-components';
 
 // Simple Button ===============================================================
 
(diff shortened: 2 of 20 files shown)


packages/ui-components/src/components/button/button.tsx
```diff
@@ -1,15 +1,19 @@
-import React, {ButtonHTMLAttributes} from 'react';
+import React from 'react';
 import styled from 'styled-components';
 
 // Simple Button ===============================================================
 
-export type ButtonProps = ButtonHTMLAttributes<HTMLButtonElement> & {
```
sepehr2github, line 6: On other buttons type, you passed ...props. I believe It would be great if we can keep this for supporting all prop types

sepehr2github (approved):
It was one of the most important tasks on our table. Good job!

packages/ui-components/src/components/button/addButton.tsx
```diff
@@ -6,14 +6,13 @@ import {FlexDiv} from './iconButton';
...
   mode = 'primary',
   size = 'default',
-  disabled = false,
-  label,
-  onClick,
+  ...props
 }) => {
   return (
-    <StyledButton mode={mode} onClick={onClick} size={size} disabled={disabled}>
+    <StyledButton mode={mode} size={size} {...props}>
```
Fabricevladimir, line 15: Are we opening up the api and allowing user defined styles?

Fabricevladimir (approved):
Looks great!

WissenIstNacht:
Addressing the concerns around the Button type:

For some reason `styledComponent`s do not like being passed `{...props}`. Maybe I'm doing something wrong, but I just can't get it to work. This means that so far we've have to explicitly pass props we want the `StyledButton` (and, by extension, the `SizedButton`) to have. So far, these were `onClick` and `disabled`. 

This means that, so far, we pretty much didn't make use of all the types we inherited by `ButtonHTMLAttributes<HTMLButtonElement>` (as we didn't pass them along to the styled components). This is why I have removed the inheritance.

With that inheritance being gone, though, passing `{...props}` around doesn't open the API w.r.t to the `Button` component. Any value passed in {...props} must eventually agree with `ButtonProps`, which now consists of exactly 5 values.

Note that the API doesn't change w.r.t. `StyledButton`. It has always been possible to pass something like `style={{background: 'red'}}` to it and will still be possible now.

PR #61, opened 2021-11-26
Feature: overview section wrapper dao 467

WissenIstNacht:
This PR

- implements the token section wrapper
- implements the transfer section wrapper
- embeds both on the current finance page as an example


RakeshUP (approved):
Works well!

I see a few files were linted in this PR, we should do a whole project lint when the PR queue is nearly empty, so as to minimize merge conflicts. Till then we'll try to cope with lint errors!

WissenIstNacht:
Very good point, thanks for bringing it up (I always forget).

We initially agreed on formatting and linting configs. But somehow it seems like not everyone follows them (or at least not automatically). Let's sync in a meeting on this to find out where we're at.

packages/ui-components/src/components/icons/interface/index.ts
```diff
@@ -9,6 +9,7 @@ export {IconChevronRight} from './icon_chevron_right';
 export {IconChevronLeft} from './icon_chevron_left';
 export {IconClose} from './icon_close';
 export {IconCopy} from './icon_copy';
+export {IconLinkExternal} from './icon_link_external';
```
sepehr2github, line 12: Can you remove this one IconLinkExternal has already been exported in line 22

sepehr2github (changes requested):
Great as always! Thank you. :)

packages/ui-components/src/components/icons/interface/index.ts
WissenIstNacht: On my screen the one on line 22 is no longer there 😅 

Explanation: I started with this PR before any `ExternalLinkIcon` was exported. After Fabrice was done with his PR I rebased this one. Since he had also exported that icon on his branch there are now two exports. So I deleted one of them. But I can change which one I delete if you want ;)

sepehr2github: No, It's ok hahaha. I just want to make sure It works fine. Thank you :))))

sepehr2github (approved):

PR #66, opened 2021-11-30
Feature/transferListItem - DAO-470 

Fabricevladimir:
Added transferListItem
- only taking into account MVP with transfer types of Deposit and Withdraw


packages/ui-components/src/components/transferListItem/transferListItem.tsx
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Date transfer was executed or a loading indication if transfer is still pending
+   */
+  transferDate: string;
+  transferType: 'Deposit' | 'Withdraw';
+  usdValue: string;
+  onClick?: () => void;
+};
+
+const Icons: {[key: string]: JSX.Element} = {
```
WissenIstNacht, line 31: Nice factoring, but why not have it as an actual (modules-scoped) react component?

WissenIstNacht (approved):
Thanks a lot, looks good 👌🏼

packages/ui-components/src/components/transferListItem/transferListItem.tsx
Fabricevladimir: As opposed to a plain object? I didn't extract it solely because I don't necessarily see any sort of re-usability for it. That's also why I didn't make it a component. What are your arguments for making it a component?

WissenIstNacht: My argument is we're using react 😂

Jokes aside: I haven't seen anyone do this and I was just curious what your thoughts were.

One thing I'm not sure about in this approach is how react handles something like this internally. And/Or how this fits into react's re-rendering process. Is it treated like a component? If not, is it constantly re-rendered if parent props change? 

I honestly have no idea. And I doubt it's really going to matter at this scale. No need to worry about this for now 👍🏼

Fabricevladimir: It shouldn't treat it any differently than an object of strings for example like we do with the style objects. More specifically it should treat it as an object of functions (`JSX.Element`). But I can look into it later because that is an interesting point. Usually, yes, it would either be a `renderAvatar` function or a separate component.

RakeshUP (approved):

packages/ui-components/src/components/transferListItem/transferListItem.tsx
RakeshUP: When the parent props change, all child nodes are tried to re-render. Only reconciliation step differs(DOM doesn't update if nothing in the child has changed)

It shouldn't matter if it's an object of components or a React component with if-else-if conditionals. 
As for the above one, because the components inside the object are pure, any performance differences shouldn't be there

WissenIstNacht: Thank you for the answer, good to know! 👌🏼

PR #67, opened 2021-12-01
Feature/fetch today token data - DAO-506 

Fabricevladimir:
Added the following hooks:
- `useIsMounted` - to determine whether component is mounted before setting the state to avoid memory leak warning
- `useInterval` - run useEffect callback at specified interval 
- `usePollTokens` - combination of previous hooks that fetches the **current** USD value for given token list


packages/web-app/src/pages/finance.tsx
```diff
@@ -1,26 +1,105 @@
...
 
 import {
   TokenSectionWrapper,
   TransferSectionWrapper,
 } from 'components/sectionWrapper';
+import usePollTokens from 'hooks/usePollTokens';
+
+const TEMP_FETCH_INTERVAL = 300000;
+
+const TOKENS = [
```
sepehr2github: I Think It's better to put all contacts inside a file on the contacts folder

packages/ui-components/src/components/tokenCard/token.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled from 'styled-components';
-import { Badge } from '../badge';
+import {Badge} from '../badge';
 
 // TODO: change types accordingly
 export type TokenCardProps = {
```
sepehr2github, line 10: On line 8 there is a type warning --> Type 'string | null' is not assignable to type 'string' for tokenSymbolURL

sepehr2github (commented):
Well Done! It's a big move on our roadmap. Thank you

packages/web-app/src/pages/finance.tsx
```diff
@@ -1,26 +1,105 @@
...
-          <ColoredDiv />
-          <ColoredDiv />
-          <ColoredDiv />
+          {prices &&
+            TOKENS.map(token => {
+              return (
+                <TokenCard
+                  key={token.name}
+                  tokenName={token.name}
+                  tokenSymbolURL={token.imgUrl}
```
WissenIstNacht: As Sepehr mentioned over in `ui-components` this field may not take a null value. I agree with that approach. It keeps presentational components as dumb as possible, thereby separating responsibility.
This means the data must be checked (and any default values assigned) before passing it to the component. 

On a different, but related note. I missed this when reviewing @RakeshUP's PR. On the `TokenCard` component: Even with non-null props and default values there's still a chance that something  goes wrong when fetching the token icon. Could you add a default symbol/icon/thing to the code base's assets and call it `onError` on the img component?

packages/web-app/src/hooks/useInterval.tsx
```diff
@@ -0,0 +1,29 @@
+import {useEffect, useLayoutEffect, useRef} from 'react';
+
+const DEFAULT_INTERVAL = 300000; //interval in ms
+
+/**
+ * Hook running callback at specified interval
+ * @param callback Function to run after each interval. Defaults to 60000ms
```
WissenIstNacht: 300000?

packages/web-app/src/hooks/usePollTokens.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const isMounted = useIsMounted();
+  const [prices, setPrices] = useState<TokenPrices>({});
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+
+  const fetchPrices = useCallback(
+    async (tokens: string[]) => {
+      setIsLoading(true);
+      for (let index = 0; index < tokens.length; index++) {
+        const address = tokens[index];
+        const price = await fetchTokenUsdPrice(address);
```
WissenIstNacht: Is there a technical reason for not using `Promise.all()` here instead of awaiting prices in sequence?

packages/web-app/src/pages/finance.tsx
```diff
@@ -1,26 +1,105 @@
...
+                      ? usdFormatter.format(
+                          Number(prices[token.address]) * token.count
+                        )
+                      : t('finance.unknownUSDValue')
+                  }
+                  changeDuringInterval="+$150,002.3"
+                  percentageChangeDuringInterval="+ 0.01%"
+                />
+              );
+            })}
```
WissenIstNacht: I know the designs aren't there yet, but could you add a simple component that says "No token information available" if `prices` is null?

WissenIstNacht (changes requested):
Nice, works well!

Left a couple of minor remarks. Also, let me know if I misunderstood something about the sequential use od async/await, as this may have a sizeable impact on performance.

packages/ui-components/src/components/tokenCard/token.tsx
Fabricevladimir: @WissenIstNacht I want to address part of your comment here as well.
The issue with this is that there are tokens have null for their icons because newly created ones don't have an icon. That said, there is a placeholder for these tokens in the designs, and I would prefer having the component itself render the placeholder as opposed to passing the placeholder icon (I think it's just a rounded element?) for the following reasons:

1. We can check if the icon is a broken icon (img onError) and display the placeholder only if the placeholder exists in the TokenCard itself.
2. Dev experience: anyone using the token card will have to create a placeholder icon if their token has no icon when it could have worked out of the box. 
2. (Semantics really lol) Passing the placeholder icon into the token card is passing 'falsy' data because that icon is not the token's icon in actual fact. 

I don't think that checking if a source is given makes the component less dumb because it doesn't need to hold state to do that.

packages/web-app/src/pages/finance.tsx
Fabricevladimir: I'm keeping `TOKENS` on this page to show that the token data will not be imported from a different module but either passed as props or queried inside  the Finance page itself

packages/web-app/src/hooks/useInterval.tsx
Fabricevladimir: No, 60000 lol

packages/web-app/src/pages/finance.tsx
Fabricevladimir: Wait, you at least have the DAO token no? So there should never be a case where there are no tokens to display...

WissenIstNacht: Ok, but isn't the presence of this token also dependent on fetching data (e.g. a subgraph call)? What happens if that fails?

Fabricevladimir: This is a design/UX issue I think. The user should be aware that there was an error fetching their token and not that they have **no** tokens because one says there is an abnormal state and the other, oops your tokens are all gone. I can add a placeholder for it, but I think a better approach is necessary.

WissenIstNacht: Fair enough, hadn't thought about that. :)

WissenIstNacht (approved):
Thanks for all the changes. Looks good to me!

sepehr2github (approved):

PR #72, opened 2021-12-05
transfer sections added

sepehr2github:
- Just all transfer sections added

### DIFF
--- a/packages/web-app/src/pages/transfers.tsx
+++ b/packages/web-app/src/pages/transfers.tsx
@@ -1,7 +1,50 @@
 import React from 'react';
+import styled from 'styled-components';
+import {useTranslation} from 'react-i18next';
+
+import {TransferSectionWrapper} from 'components/sectionWrapper';
 
 const Transfers: React.FC = () => {
-  return <h1>All Transfers Page</h1>;
+  const {t} = useTranslation();
+
+  return (
+    <div className={'m-auto mt-4 w-8/12'}>
+      <SectionContainer>
+        <TransferSectionWrapper title={t('allTransfer.thisWeek')}>
+          <div className="my-2 space-y-2 border-solid">
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+          </div>
+        </TransferSectionWrapper>
+      </SectionContainer>
+      <SectionContainer>
+        <TransferSectionWrapper title={'November'}>
+          <div className="my-2 space-y-2 border-solid">
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+          </div>
+        </TransferSectionWrapper>
+      </SectionContainer>
+      <SectionContainer>
+        <TransferSectionWrapper title={'2021'}>
+          <div className="my-2 space-y-2 border-solid">
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+          </div>
+        </TransferSectionWrapper>
+      </SectionContainer>
+    </div>
+  );
 };
 
 export default Transfers;
+
+const ColoredDiv = styled.div.attrs({className: 'h-6 w-full bg-blue-100'})``;
+
+const SectionContainer = styled.div.attrs({className: 'my-5'})``;
--- a/packages/web-app/src/components/sectionWrapper/index.tsx
+++ b/packages/web-app/src/components/sectionWrapper/index.tsx
@@ -9,6 +9,7 @@ import {AllTokens, AllTransfers} from 'utils/paths';
 export type SectionWrapperProps = {
   title: string;
   children: React.ReactNode;
+  showButton?: boolean;
 };
 
 // NOTE: It's possible to merge these two components. But I'm not sure it makes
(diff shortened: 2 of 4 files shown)


packages/web-app/src/pages/transfers.tsx
```diff
@@ -1,7 +1,50 @@
...
+        <TransferSectionWrapper title={t('allTransfer.thisWeek')}>
+          <div className="my-2 space-y-2 border-solid">
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+          </div>
+        </TransferSectionWrapper>
+      </SectionContainer>
+      <SectionContainer>
+        <TransferSectionWrapper title={'November'}>
```
WissenIstNacht, line 22: I think this should be computed dynamically.

Also, this must be translatable to different language.

packages/web-app/src/pages/transfers.tsx
```diff
@@ -1,7 +1,50 @@
...
+          <div className="my-2 space-y-2 border-solid">
+            <ColoredDiv />
+            <ColoredDiv />
+            <ColoredDiv />
+          </div>
+        </TransferSectionWrapper>
+      </SectionContainer>
+      <SectionContainer>
+        <TransferSectionWrapper title={'2021'}>
```
WissenIstNacht, line 31: Same here

packages/web-app/src/pages/transfers.tsx
WissenIstNacht, line 22: I think this should be computed dynamically.

Also, this must be translatable to different language.

packages/web-app/src/pages/transfers.tsx
WissenIstNacht, line 31: Same here

WissenIstNacht (changes requested):
Thanks mate. 🙂 

I think it should already be possible to set the sections titles dynamically (even if the data is not present. Maybe a hook that, given a series of dates, divides them into last week, last month, last year, older?

packages/web-app/src/pages/transfers.tsx
```diff
@@ -1,7 +1,50 @@
 import React from 'react';
+import styled from 'styled-components';
+import {useTranslation} from 'react-i18next';
+
+import {TransferSectionWrapper} from 'components/sectionWrapper';
```
Fabricevladimir, line 5: If we are using these guys outside of the finance overview page, we could combine them into one SectionWrapper or ListGroup component that shows IconButton and SeeAllButton conditionally. That said, this probably shouldn't be done right now.

packages/web-app/src/pages/transfers.tsx
Fabricevladimir: Agreed. There was a thread about a formatting and date library. Lets see if we can agree on one quickly so we can use it there.

packages/web-app/src/pages/transfers.tsx
```diff
@@ -1,7 +1,50 @@
...
+          </div>
+        </TransferSectionWrapper>
+      </SectionContainer>
+    </div>
+  );
 };
 
 export default Transfers;
+
+const ColoredDiv = styled.div.attrs({className: 'h-6 w-full bg-blue-100'})``;
```
Fabricevladimir, line 48: `blue-100` isn't part of the design system, is it?

Fabricevladimir (commented):
Looks good. We'll be able to flesh out the page easily now.

packages/web-app/src/pages/transfers.tsx
WissenIstNacht: I think he copy pasted from my PR a while back, sorry for that. 😅

That being, said their just placeholders 👍🏼

sepehr2github: @Fabricevladimir, I copied from @WissenIstNacht as a placeholder. All the ColoredDiv will remove soon

Fabricevladimir: Pffft! I didn't even realize that it was for the placeholder. This one is on me

packages/web-app/src/pages/transfers.tsx
sepehr2github: I had the same idea but due to lack of transfers data. It's not clear to me what do you expect?
@Fabricevladimir 
@WissenIstNacht

packages/web-app/src/pages/transfers.tsx
sepehr2github: Good idea. We will definitely do that

packages/web-app/src/pages/transfers.tsx
WissenIstNacht: What would you do if you had data? :)

Essentially, there's two things that need to be done here: 

1. Determine what is the last week, month, year (and/or maybe older). Note, this has nothing to do with transfer data being available.
2. Categorize data according to the answers above. Note, this does, but you should be able to come up with some data that allows you to at least test this.

The first step could be done with a simple `utils` type of function that returns the date that delimit the categories (`getDateCategories: () => dates[]`). The second could be done with a hook that takes transfer data and splits them up into these categories (`useCategorizedTransfers: (TransferData[]) => TransferData[][]`). For this step, I think we'll have to assume that there will be some type like `TransderData` that will contains some form of date.

WissenIstNacht:
Hey @sepehr2github why did you close this PR? 😅

sepehr2github:
> Hey @sepehr2github why did you close this PR? 😅

Because I want to open another PR. :)))

WissenIstNacht:
But you can have several open PRs ...? :)

Are you going to continue working on this later or your not going to be working on this anymore?

PR #74, opened 2021-12-06
Feature/Radio group component DAO-463

RakeshUP:


sepehr2github (approved):
Nice job! thank you.

WissenIstNacht (approved):
Thanks! :)

I went ahead and formatted your code, hope you don't mind :p

One minor detail (change it if you have time/ agree with it) would be to have the variable on the button renamed from `active` to something else. For example, `isSelected` or even better `isChecked` (which would more accurately reflect the radio group component).

PR #79, opened 2021-12-09
Feature/Transfers page DAO-478

RakeshUP:


packages/web-app/src/pages/transfers.tsx
```diff
@@ -1,8 +1,73 @@
+import {AddButton, SearchInput, TransferListItem} from '@aragon/ui-components';
 import React from 'react';
 import {withTransaction} from '@elastic/apm-rum-react';
+import {useTranslation} from 'react-i18next';
+import styled from 'styled-components';
+
+const transfers = [
```
WissenIstNacht: Could you add a type to this? (there should be a types file in `utils` for global-ish types)

packages/web-app/src/components/transferList/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+const TransferList: React.FC<TransferListProps> = ({transfers}) => {
+  const {t} = useTranslation();
+
+  if (transfers.length === 0)
+    return <p data-testid="transferList">{t('finance.noTransfers')}</p>;
+
+  return (
+    <div className="space-y-2" data-testid="transferList">
+      {transfers.map((transfer, index) => {
+        // key should be changed later
```
WissenIstNacht, line 20: Do you think it's worth adding a hashing library that computes `uid`s for keys?

packages/web-app/src/pages/transfers.tsx
```diff
@@ -1,8 +1,73 @@
...
+    <Layout>
+      <CenteredFlex>
+        <div className="space-y-1">
+          <Title>{t('finance.allTransfers')}</Title>
+          <SubTitle>$1,002,200.00 {t('finance.totalVolume')}</SubTitle>
+        </div>
+        <AddButton label={t('finance.newTransfer')} />
+      </CenteredFlex>
+      <SearchInput placeholder="Type to filter" />
+      <div className="space-y-1.5">
```
WissenIstNacht: Button group here? (I think you can ommit the Button labeled "External COntracts") for now.

WissenIstNacht (changes requested):
Thanks a lot! :)

Aside from the comments, don't forget to take care of the linting issues.

packages/web-app/src/pages/transfers.tsx
WissenIstNacht: Ignore the above comment, I forgot that there is a separate ticket for this. Sorry!

packages/web-app/src/components/transferList/index.tsx
Fabricevladimir: Each transfer will have an id coming from the subraph, we should be able to use that.

Fabricevladimir (approved):
Good work! This should be ready to merge once the changes requested by @WissenIstNacht are made. Also, do you mind copying over the `warning` colors from the ui-components package so that the withdraw can show with proper colors? In the future, we should probably delete the web-app tailwind config and reference the one from the ui-components if possible since there doesn't seem to be any difference between the two.

packages/web-app/src/components/transferList/index.tsx
RakeshUP: Yeah as Fabrice said, there would be an id present which should be used as the key

Most of the hashing libraries use the current time for calculating hash, so it would re-render the list item every time the parent renders. It would behave the same way as not passing the key prop only.

Using index causes bugs though!

packages/web-app/src/pages/transfers.tsx
RakeshUP: Types should come from the subgraph by using codegen. I saw a type definition for types in the utils file, which should probably be removed once we fetch types from the subgraph

WissenIstNacht: This is partly true:

- It is true for data that actually comes from subgraph. This will only be a subset of data we use, though (think token information).
- The development of subgraph is not done (and should not be done) in isolation. Sarkawt would be very happy to have some insight into what data we need. Types would be a great way to start.
- Nothing stops you from changing any types you declare now later :D

But it's OK for now. :)

packages/web-app/src/components/transferList/index.tsx
WissenIstNacht: > Yeah as Fabrice said, there would be an id present which should be used as the key

Right 👍🏼

> Most of the hashing libraries use the current time for calculating hash [...]

Well... yes... that would not be the smartest thing to do 😄. I was thinking along the lines of an object hash or something 👍🏼

WissenIstNacht:
Hey, thanks for getting back to me on the comments! :)

Please address the linting issues and we should be good to go 👍🏼 
(One is related to the above type issue. If you really can't bring yourself to quickly  copy it from your IDE then at least add an eslint ignore statement :D)

RakeshUP:
Fixed the linting issues and added the type.

WissenIstNacht (approved):
Sorry, forgot to approve.

WissenIstNacht:
Hey, there were some merge conflicts with my PR, so I resolved them. Also took care of the linting issue 👍🏼

I'll let you merge in case you want to have a look at the changes 🙂

PR #81, opened 2021-12-09
Fix: Token card size DAO-529

WissenIstNacht:
This ticket:

- adapts the token card's text sizes to the ones in the design. 
- changes the badge from being a span to being a div, since this leads to it not being displayed correctly
-  removes all font weight except for normal and bold, which are the only font weights defined for this project.

General note: Font-weight of normal is currently wort 500 (i.e., is effectively medium), instead of the usual 400. This should probably be changed in the design. Otherwise we have to specify `font-normal` in every component.


Fabricevladimir (approved):
Thanks for the fix.

RakeshUP (approved):

PR #82, opened 2021-12-10
Feature/transfer modal DAO-445 

sepehr2github:
- Modal added to ui components
- Radix Dialog package added
- Transfer modal added
- add background selection and hide border option to many & actionItemList
- change menu context to wallet context 
- transferModal context added (we can use this context for other transfer modal pages)


packages/ui-components/package.json
```diff
@@ -71,6 +72,7 @@
     "typescript": "^4.4.4"
   },
   "dependencies": {
+    "@radix-ui/react-dialog": "^0.1.1",
```
Fabricevladimir: Is there a reason why the package is installed as both dev and regular dependency?

packages/ui-components/src/components/modal/modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  */
+  background?: string;
+  /**
+  * Styles
+  */
+  style?: CSSProperties | undefined;
+  /**
+  * The `onClose` prop allows passing a function that will be called once the modal has been dismissed.
+  */
+  onClose: () => void;
```
Fabricevladimir: Can this property be optional?

packages/ui-components/stories/modal.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {Modal, ModalProps} from '../src';
+
+export default {
+  title: 'Components/Modal',
+  component: Modal,
+} as Meta;
+
+const Template: Story<ModalProps> = args => <Modal {...args} />;
+
+const popoverContent: ReactNode = (
```
Fabricevladimir: Copy paste? Should be `ModalContent`?

packages/web-app/src/pages/finance.tsx
```diff
@@ -64,26 +66,33 @@ const POLL_TOKEN_LIST = TEMP_TOKENS.map(({address, decimals}) => ({
...
-        <div className="my-2 space-y-2 border-solid">
-          <ColoredDiv />
-          <ColoredDiv />
-          <ColoredDiv />
-          <ColoredDiv />
-          <ColoredDiv />
-        </div>
-      </TransferSectionWrapper>
+      <PageWrapper
+        title={t('finance.allTransfers')}
```
Fabricevladimir: The title is on the wrong page. It should be on the `AllTransfers` page not Finance

packages/web-app/src/containers/transferMenu/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+import {useTranslation} from 'react-i18next';
+
+import {useTransferModalContext} from 'context/transfersModal';
+
+const TransferMenu: React.FC = () => {
+  const {isOpen, close} = useTransferModalContext();
+  const {t} = useTranslation();
+
+  return (
+    <Modal
```
Fabricevladimir, line 13: The position for the modal is different from that in the designs. In the designs, it's at the top of the page. Not a major concern for now probably. Just something to keep in mind.

packages/web-app/src/locales/en/common.json
```diff
@@ -29,6 +29,15 @@
     "tokenSection": "Token Section",
     "transferSection": "Transfer Section",
     "unknownUSDValue": "Value unknown",
-    "tokens": "Tokens"
+    "tokens": "Tokens",
+    "allTransfers":"All Transfers",
+    "addTransfers":"Add Transfer"
```
Fabricevladimir: Should be "New Transfer" for the button.

Fabricevladimir (changes requested):
Nice work setting things up!

Fabricevladimir (approved):

RakeshUP (approved):
The space between the list items in the modal is 24px vs 12px in the design
Modal close on pressing escape key doesn't seem to be working. Do you know why?

packages/ui-components/src/components/modal/modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+   */
+  title?: string;
+  /**
+   * Content
+   */
+  children: ReactNode;
+  /**
+  * Background color
+  */
+  background?: string;
```
WissenIstNacht: Hey, did you discuss this with Selim? It doesn't seem to me like we'll have a lot of different background colors on Modals.

Side note: I'm not a fan of letting the user pass Tailwind color names here. If anything, I think it would be better to give the user some colors they can chose from. Then, we translate that to tailwind classes inside the Modal component. Also, this would be something worth documenting in the comment above. ;)

packages/ui-components/src/components/modal/modal.tsx
```diff
@@ -0,0 +1,106 @@
...
+  open = true,
+  onOpenChange,
+  onClose,
+  ...props
+}) => {
+  return (
+    <>
+      <Root {...{ open, onOpenChange }}>
+        <Backdrop visible={open}/>
+        <ModalContent
```
WissenIstNacht: Small detail: I would call this `ModalContainer`, as the children are the content. (See your comment on line 21)

packages/ui-components/src/components/actionListItem/actionListItem.tsx
```diff
@@ -33,11 +38,12 @@ export const ActionListItem: React.FC<ActionListItemProps> = ({
   subtitle,
   title,
   onClick,
+  background,
+  bordered,
```
WissenIstNacht: Defaulting to true would avoid the need to change all existing items that have a border. On the other hand I realize we'll have a lot of borderless action items. Might be worth asking Selim on the topic?

packages/web-app/src/pages/finance.tsx
```diff
@@ -64,26 +66,34 @@ const POLL_TOKEN_LIST = TEMP_TOKENS.map(({address, decimals}) => ({
...
-          <ColoredDiv />
-          <ColoredDiv />
-        </div>
-      </TransferSectionWrapper>
+      <PageWrapper
+        title={'$469,657.98'}
+        buttonLabel={t('TransferModal.newTransfer') as string}
+        subtitle={'+ $120,200'}
+        onClick={open}
+        subtitleColor="primary-500"
```
WissenIstNacht: Again, I wouldn't pass tailwind colors into here.

packages/web-app/src/pages/finance.tsx
```diff
@@ -64,26 +66,34 @@ const POLL_TOKEN_LIST = TEMP_TOKENS.map(({address, decimals}) => ({
...
-        <div className="my-2 space-y-2 border-solid">
-          <ColoredDiv />
-          <ColoredDiv />
-          <ColoredDiv />
-          <ColoredDiv />
-          <ColoredDiv />
-        </div>
-      </TransferSectionWrapper>
+      <PageWrapper
+        title={'$469,657.98'}
```
WissenIstNacht, line 25: This should be dynamic. (I know, real data is not available. But there is some mock data. You could set up a hook that computes the necessary information given that mock data.)

packages/web-app/src/context/walletMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+type Props = Record<'children', ReactNode>;
+
+const WalletMenuProvider: React.FC<Props> = ({children}) => {
+  const [isOpen, setIsOpen] = useState<WalletMenuContextType['isOpen']>(false);
+
+  const open = () => setIsOpen(true);
+  const close = () => setIsOpen(false);
+
+  const value = useMemo(
```
WissenIstNacht, line 25: I don't think this is necessary. None of the computations here are expensive.

packages/web-app/src/context/transfersModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+type TransferModalContextType = {
+  isOpen: boolean;
+  open: () => void;
+  close: () => void;
+};
+
+type Props = Record<'children', ReactNode>;
+
+const TransferModalsProvider: React.FC<Props> = ({children}) => {
```
WissenIstNacht, line 21: Why do we need two such contexts? Wouldn't one be enough? You can't have both opened at the same time anyway, right?

packages/web-app/src/components/sectionWrapper/index.tsx
```diff
@@ -11,10 +16,52 @@ export type SectionWrapperProps = {
...
+
 // NOTE: It's possible to merge these two components. But I'm not sure it makes
 // things any simpler right now. However, if other sections wrappers like these
 // are added in the future and all have similar style, feel free to merge them.
 
+/**
+ * finance Page wrapper. Consists of a header with a title and a
+ * icon button.
+ */
+export const PageWrapper = ({
```
WissenIstNacht: Could you move this to a separate file? (Page != Section :D)

WissenIstNacht (commented):
Hey there, thanks a lot for all the work!

There's nothing objectively wrong with your PR, but there's a lot of things I find a bit strange. In particular, I feel like there's a lot of things in your PR that should have been discussed with Selim to avoid misunderstandings and problems later on.

WissenIstNacht:
Also a LOT of tests are currently not passing. Could you try to fix them?

// update: I think a lot of those are related to SVG issues, which is solved in Fabrices PR. Please ignore those. However, please adapt the `actionListItem` tests to reflect the changes on the wide prop. 👍🏼 

//update2: Just realized Fabrice is solving all the testing issues on his PR.

packages/web-app/src/context/transfersModal.tsx
sepehr2github: Context structure is one of the challenges that we need to discuss. I consider each context based on its usage. We Will use TransfetModal context for its related states also the same for the wallet modal. (In other words, It's a placeholder for more states). Let me know if you have different idea for defining contexts

packages/web-app/src/context/walletMenu.tsx
sepehr2github: For now yes, but in the future, We will add more wallet states. (When we want to implement the full wallet flow)

PR #83, opened 2021-12-10
Chore/update buttons - DAO-500

Fabricevladimir:
-Refactored button to match new designs
-Removed requested button components
-Updated web app to use new button
-Fixed ui-components tests


packages/ui-components/src/components/button/buttonBase.tsx
```diff
@@ -0,0 +1,97 @@
...
+  iconRight,
+  iconLeft,
+  iconOnly = false,
+  size = 'medium',
+  label,
+  onClick,
+  ...props
+}) => {
+  return (
+    <BaseStyledButton {...props} size={size} onClick={onClick}>
```
sepehr2github: In button base text should have gray text in disabled mode

packages/ui-components/src/components/button/buttonIcon.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+import {ButtonBase} from './buttonBase';
+import type {IconType} from '../icons';
+import type {ButtonBaseProps} from './buttonBase';
+
+export type ButtonIconProps = Omit<
+  ButtonBaseProps,
+  'label' | 'iconRight' | 'iconLeft' | 'iconOnly'
+> & {
+  bgWhite?: boolean;
```
sepehr2github, line 12: Can you explain about this property? Does it control the "on gray/on white" types?

sepehr2github (commented):
Good work! I know how much you hate refactoring hahaha :). Thank you!
a minor change is needed.

packages/ui-components/src/components/button/buttonIcon.tsx
Fabricevladimir: Yes lol. Suggestions concerning the name are very much welcomed!

packages/ui-components/src/components/button/buttonBase.tsx
Fabricevladimir: All coloring and state styles except for `focus` are left up to the specific buttons to implement themselves. `focus` is implemented in the `buttonBase` because every component has the same focus ring, but not every button is going to have gray text when disabled. Also, `buttonBase` is only meant to be used as a base for other buttons and is not being exported.

RakeshUP (approved):

packages/ui-components/src/components/button/buttonBase.tsx
```diff
@@ -0,0 +1,97 @@
...
+  visible: boolean;
+};
+
+const Label = styled.span.attrs(({visible}: LabelProps) => {
+  let className: string | undefined;
+  if (!visible) className = 'hidden';
+  return {className};
+})<LabelProps>``;
+
+const IconContainer = styled.span.attrs(({size = 'medium'}: SizeProps) => {
```
WissenIstNacht, line 90: is the `span` intentional?

packages/ui-components/src/components/button/buttonBase.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <IconContainer size={size}>{iconRight}</IconContainer>
+      )}
+    </BaseStyledButton>
+  );
+};
+
+/**********************************
+ *             STYLES             *
+ **********************************/
+const sizeStyles = {
```
WissenIstNacht, line 49: Cool idea :)

packages/ui-components/src/components/button/buttonBase.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export type ButtonBaseProps = ButtonHTMLAttributes<HTMLButtonElement> & {
+  disabled?: boolean;
+  iconOnly?: boolean; // Guard against passing label to ButtonIcon
+  iconLeft?: React.FunctionComponentElement<IconType>;
+  iconRight?: React.FunctionComponentElement<IconType>;
+  label?: string;
+  mode?: 'primary' | 'secondary' | 'ghost';
+  size?: 'small' | 'medium' | 'large';
+  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
```
WissenIstNacht: Isn't this part of the extended type already?

packages/ui-components/src/components/button/buttonText.tsx
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import styled from 'styled-components';
+
+import {ButtonBase} from './buttonBase';
+import type {ButtonBaseProps} from './buttonBase';
+
+// Omit label to make it required
+export type ButtonTextProps = Omit<ButtonBaseProps, 'label' | 'iconOnly'> & {
```
WissenIstNacht, line 8: why omit label when you add it back below?

//update: Here it's no longer optional, right?

packages/web-app/src/components/wallet/index.tsx
```diff
@@ -1,30 +0,0 @@
-import React from 'react';
```
WissenIstNacht, line 1: Slightly confused here ^^'

What replaces this?

packages/web-app/src/pages/tokens.tsx
```diff
@@ -41,7 +46,11 @@ const Tokens: React.FC = () => {
           <Title>5 {t('finance.tokens')}</Title>
           <SubTitle>$469,657.98 Holdings</SubTitle>
         </div>
-        <AddButton label="New Transfer" />
+        <ButtonText
+          size="large"
+          label={t('finance.newTransfer')}
```
WissenIstNacht, line 51: Token Page :D (should be `newToken` or something)

WissenIstNacht (commented):
Thanks a lot! :)

I really hope we don't have to repeat this exercise anytime soon.

packages/ui-components/src/components/button/buttonBase.tsx
Fabricevladimir: Yeah. It's being used to force the icon to take the sizing

packages/ui-components/src/components/button/buttonBase.tsx
Fabricevladimir: I was definitely inspired by the original code lol

packages/ui-components/src/components/button/buttonBase.tsx
Fabricevladimir: Good catch, `disabled` as well.

packages/ui-components/src/components/button/buttonText.tsx
Fabricevladimir: Exactly. There must be a better way around it though.

packages/web-app/src/components/wallet/index.tsx
Fabricevladimir: I'm not sure what you mean... The `Wallet` component was the first placeholder we had at the very beginning. I just removed it entirely from the project.

packages/web-app/src/pages/tokens.tsx
Fabricevladimir: This is consistent with the designs. I think it makes sense too because even adding a token itself is considered a transfer.

WissenIstNacht: You're right, my bad!

WissenIstNacht (approved):
Thanks for addressing the comments. Green light from my side 👍🏼

Fabricevladimir:
Not all commits were signed. Sending new PR

PR #85, opened 2021-12-10
Fix/value unknown DAO-525

Fabricevladimir:
- Modified static data to not show data unavailable when corresponding USD value is unknown


RakeshUP (approved):

sepehr2github (approved):

packages/ui-components/src/components/tokenCard/token.tsx
```diff
@@ -47,13 +53,25 @@ export const TokenCard: React.FC<TokenCardProps> = props => {
...
-          <Badge
-            label={props.percentageChangeDuringInterval}
-            colorScheme="success"
-          />
+          {props.changeDuringInterval && (
+            <ToggleMobileVisibility visible={false}>
+              <span
+                className={
+                  changeType === 'Positive'
+                    ? 'text-success-800'
```
sepehr2github, line 61: Do we have a warning status either or It's only successful or critical?

sepehr2github (approved):

packages/ui-components/src/components/tokenCard/token.tsx
Fabricevladimir: For the moment only successful or critical.

PR #86, opened 2021-12-10
Feature: typeahead search DAO-475

WissenIstNacht:
This PR:

- Exposes the `value` and `onChange` attributes of `<input />` (and propagates it up throughout the parent react components). Thereby allowing for to be used as a controlled component.
- Introduces a series of hooks (`useDaoTokens`, `useTokenInformation`) that should be used as a blue print for fetching token-related later on.
- It introduces the `HookData<T>` type that should be used as output type of asynchronous data-fetching hooks.
- It refactors the existing `usePollTokens` hook to conform to the above
- It introduces `sort` and `filter` methods for token information. 
- It refactors the existing search Input on `finance/tokens` page into a type-ahead search that allows to filter through tokens.


packages/ui-components/src/components/input/textInput.tsx
```diff
@@ -31,16 +33,17 @@ export const TextInput: React.FC<TextInputProps> = ({
   ...props
 }) => {
   return (
-    <Container data-testid="input" {...{mode,disabled,side}} >
-      <StyledInput {...props} disabled={disabled}/>
+    <Container data-testid="input" {...{mode, disabled, side}}>
+      <StyledInput {...props} disabled={disabled} {...props} />
```
Fabricevladimir: Spreading props twice.

packages/ui-components/src/components/input/textInput.tsx
```diff
@@ -1,25 +1,27 @@
-import React, {ReactNode} from 'react';
+import React, {BaseSyntheticEvent, ReactNode} from 'react';
 import styled from 'styled-components';
 
 export type TextInputProps = {
```
Fabricevladimir: This component doesn't seem to be using the base input type which means we lose all kinds of native input attributes. Can you modify the props to extend  `InputHTMLAttributes<HTMLInputElement>`? That way other than our additional props, everything is set automatically? Even the `onChange` event can be omitted from the props then.

packages/ui-components/src/components/input/textInput.tsx
```diff
@@ -1,25 +1,27 @@
...
+  /**
    * Wheter the icon is left or right of the input
    */
   side: 'left' | 'right';
   /**
    * Placeholder
    */
   placeholder?: string;
+  value: string;
+  onChange: (event: BaseSyntheticEvent) => void;
```
Fabricevladimir: I think this might be the following might be the proper type: `React.ChangeEventHandler<HTMLInputElement>`. I don't think it should matter much since `BaseSyntheticEvent` is (?) the parent of all the other events, but just in case so we don't lose some specific attributes...

packages/web-app/src/hooks/useDaoTokens.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * @returns List of token addresses that the DAO as well as the hook state.
+ */
+export const useDaoTokens = (daoAddress: Address) => {
+  const [tokens, setTokens] = useState<Address[]>([]); // eslint-disable-line
+  const [loading, setLoading] = useState(false); // eslint-disable-line
+  const [error, setError] = useState<Error>(); // eslint-disable-line
+
+  useEffect(() => {
+    // TODO Fetch data from subgraph here
+    if (daoAddress) 42;
```
Fabricevladimir, line 22: I see you 42 lol

packages/web-app/src/utils/tokens.ts
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @example sortTokens(baseTokenInfos[], 'name');
+ * @example sortTokens(baseTokenInfos[], 'count');
+ */
+export function sortTokens<K extends keyof BaseTokenInfo>(
+  tokens: BaseTokenInfo[],
+  criteria: K
+) {
+  function sorter(a: BaseTokenInfo, b: BaseTokenInfo) {
+    if (a[criteria] < b[criteria]) {
```
Fabricevladimir, line 19: Good call since we might have to sort them based on different criteria

packages/web-app/src/utils/tokens.ts
```diff
@@ -0,0 +1,56 @@
...
+ * term.
+ */
+export function filterTokens(tokens: BaseTokenInfo[], searchTerm: string) {
+  function tokenInfoMatches(token: BaseTokenInfo, term: string) {
+    const lowercaseTerm = term.toLocaleLowerCase();
+    const lowercaseSymbol = token.symbol.toLocaleLowerCase();
+    const lowercaseAddress = token.address.toLocaleLowerCase();
+    const lowercaseName = token.name.toLocaleLowerCase();
+    return (
+      lowercaseSymbol.indexOf(lowercaseTerm) >= 0 ||
```
Fabricevladimir, line 47: This man is a fan of IE 😄

Fabricevladimir (changes requested):
Great work! Ready for whatever data coming in

packages/web-app/src/utils/tokens.ts
WissenIstNacht: Wait, what? Internet explorer...? 😅

packages/ui-components/src/components/input/textInput.tsx
WissenIstNacht: Legacy... Will take a moment to add the proper types!

packages/web-app/src/utils/tokens.ts
```diff
@@ -0,0 +1,56 @@
...
+ * of the information object that can be compared using '<', '>'.
+ *
+ * @param tokens List of token (basic) token information
+ * @param criteria Field of the information object that determines the sort
+ * order. Must be a key of BaseTokenInfo
+ *
+ * @example sortTokens(baseTokenInfos[], 'name');
+ * @example sortTokens(baseTokenInfos[], 'count');
+ */
+export function sortTokens<K extends keyof BaseTokenInfo>(
```
sepehr2github, line 14: I like How you played with Typing stuff:)))). Well Done!

sepehr2github (approved):
Thank you, Vali. Nice PR!

packages/web-app/src/utils/tokens.ts
sepehr2github: Also Please Take a care of Fabrice Comments

packages/web-app/src/utils/tokens.ts
Fabricevladimir: Oh that was about preferring `indexOf` over `includes`. `includes` is ES6 so odds are IE does not support it!

WissenIstNacht: Ehrm, yes, exactly, that was my intention! haha

packages/ui-components/src/components/input/textInput.tsx
```diff
@@ -63,6 +53,8 @@ export const Container = styled.div.attrs(
...
 )<StyledCotainerProps>``;
 
-export const StyledInput = styled.input.attrs({
-  className: 'w-full bg-transparent focus:outline-none',
-})``;
+export const StyledInput = styled.input.attrs(() => {
+  let myClassName: string | undefined =
+    'w-full bg-transparent focus:outline-none';
+  return {className: myClassName};
+})<React.InputHTMLAttributes<HTMLInputElement>>``;
```
Fabricevladimir, line 60: Isn't `StyledInput` all by itself enough? Assuming that styled-components has their return type correctly I mean...

Fabricevladimir (approved):
Ready to merge!

packages/ui-components/src/components/input/textInput.tsx
WissenIstNacht: Without the <type>, you mean? Possibly :)

PR #88, opened 2021-12-10
init subgraph and unit testing

Rekard0:
- Setup Subgraph
- Add temporary ABI folder
- Unit test
- Add script for creating manifest
- Template draft
- Entity draft


packages/subgraph/abis/Registry.json
```diff
@@ -0,0 +1,73 @@
+{
+  "_format": "hh-sol-artifact-1",
+  "contractName": "Registry",
+  "sourceName": "contracts/src/registry/Registry.sol",
+  "abi": [
```
yuetloo, line 2: since the subgraph only needs the abi, this may simplify things (i.e. bytecode will change if contract implementation changes) if only abi is included in this file, like https://github.com/aragon/court-subgraph/blob/master/abi/CourtSubscriptions.json

packages/subgraph/manifest/subgraph.placeholder.yaml
```diff
@@ -0,0 +1,147 @@
...
+        - event: NewDAO(indexed string,indexed address,indexed address)
+          handler: handleNewDAO
+      file: ./src/dao.ts
+# templates
+templates:
+  # DAO
+  - name: DAO
+    kind: ethereum/contract
+    network: rinkeby
+    source:
```
yuetloo, line 34: network should be parameterized to be replaced by mustashe substitution similar to how dataSources are defined above.

packages/subgraph/subgraph.yaml
```diff
@@ -0,0 +1,148 @@
+# Do not edit subgraph.yaml,this is a generated file. 
+# Instead, edit subgraph.placeholder.yaml and run: yarn manifest-[network-name]
+specVersion: 0.0.2
+description: A template for Aragon subgraphs
```
yuetloo: I think this generated file shouldn't be added to repository.

packages/subgraph/src/dao.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  DaoEntity.name = event.params.name.toString();
+
+  DaoEntity.daoAddress = event.params.dao;
+  DaoEntity.creator = event.params.creator;
+
+  let daoContract = DAOContract.bind(event.params.dao);
+  let processesAdress = daoContract.processes();
+  let permissionsAdress = daoContract.permissions();
+
```
yuetloo, line 18: typo for `processesAdress` and `permissionsAdress`, missing `d`

packages/subgraph/src/dao.ts
```diff
@@ -0,0 +1,30 @@
...
+export function handleNewDAO(event: NewDAO): void {
+  let DaoEntity = new Dao(event.params.dao.toHexString()); // use dao address as id, because it should not repeat
+
+  DaoEntity.name = event.params.name.toString();
+
+  DaoEntity.daoAddress = event.params.dao;
+  DaoEntity.creator = event.params.creator;
+
+  let daoContract = DAOContract.bind(event.params.dao);
+  let processesAdress = daoContract.processes();
```
yuetloo: Here, I would suggest test registering a dao with invalid contract address (i.e. valid address but not a contract address) and see if thegraph will fail to index/sync when it fails to load `daoContract.processes()` because event.params.dao is not a contract address.  This is one of the scenarios we have to fix for the govern subgraph because of `bad data`.

you may have to use `try_` function to get the data as the `try catch` option.

packages/subgraph/src/dao.ts
```diff
@@ -0,0 +1,30 @@
...
+  let permissionsAdress = daoContract.permissions();
+
+  DaoEntity.metadata = daoContract.metadata();
+  DaoEntity.processes = processesAdress;
+  DaoEntity.permissions = permissionsAdress;
+  DaoEntity.executor = daoContract.executor();
+
```
yuetloo, line 23: looks like there need to be a way to track vault/voting contract added to the dao somewhere.

packages/subgraph/src/dao.ts
Rekard0: Thanks, i will address that once we have a deploy-able branch

packages/subgraph/src/dao.ts
Rekard0: aware of it, I am awaiting some changes from @novaknole

yuetloo (approved):
approving this as other issues will be addressed as the underlying smart contracts evolve.

PR #93, opened 2021-12-14
Feature/Label component DAO-530

RakeshUP:


packages/ui-components/src/components/label/label.tsx
```diff
@@ -0,0 +1,41 @@
...
+  helpText,
+  isOptional=false
+}) => {
+  return (
+    <VStack data-testid="label">
+      <LabelLine>
+        <Heading>{label}</Heading>
+        {isOptional && <Badge label='Optional' />}
+      </LabelLine>
+      {helpText && <HelpText>{helpText}</HelpText>}
```
Fabricevladimir, line 24: If `helpText` is being shown only when it's not null or undefined, should it not be optional as well?

packages/ui-components/src/components/label/label.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const Label: React.FC<LabelProps> = ({
+  label,
+  helpText,
+  isOptional=false
+}) => {
+  return (
+    <VStack data-testid="label">
+      <LabelLine>
+        <Heading>{label}</Heading>
+        {isOptional && <Badge label='Optional' />}
```
Fabricevladimir: For the sake of translations, the `label` property of the `Badge` should also be received as a prop. Very tedious to work with. Perhaps we can have something like the following for these things: `<Badge label={badgeLabel || 'Optional'} />`.  
But then again, that pollutes all the code. Any ideas for this kind of stuff in general? It's almost like someone not using internationalization is being punished...

Fabricevladimir (commented):
Good work!

packages/ui-components/src/components/label/label.tsx
RakeshUP: I've made changes as you've suggested. I too don't have any idea right now :)

packages/ui-components/src/components/label/label.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import styled from 'styled-components';
+import {Badge} from '../badge';
+
+export type LabelProps = {
+  label: string;
+  helpText?: string;
+  isOptional?: boolean;
```
WissenIstNacht, line 8: Very minor detail here:

To stay consistent with the designs (and the vocabulary in it), I would treat the badge the same way you did `helptext`. I.e., just have one optional field called `badge` of type `string`. If empty, no badge is shown, otherwise, the badge is displayed with the text passed in `badge`.

WissenIstNacht (dismissed):
Looking good, thanks! 👌🏻

packages/ui-components/src/components/label/label.tsx
Fabricevladimir: Agreed. In fact, the whole badge might have been meant to be passed in. Like what if you need a success badge next to your label? This would also resolve the need to pass in a badge label as props.

Fabricevladimir (approved):
Ready to merged once the requested changes are made!

packages/ui-components/src/components/label/label.tsx
```diff
@@ -0,0 +1,43 @@
...
+  label,
+  helpText,
+  isOptional = false,
+  badgeLabel
+}) => {
+  return (
+    <VStack data-testid="label">
+      <LabelLine>
+        <Heading>{label}</Heading>
+        {isOptional && <Badge label={badgeLabel || 'Optional'} />}
```
sepehr2github, line 22: Can you make the 'Optional' lowercase?

sepehr2github (approved):
Thank you Rakash. Nice Pr!

packages/ui-components/src/components/label/label.tsx
WissenIstNacht: This crossed my mind too. I decided against it since I would expect him to design the interface (on figma) differently. Currently, there's only a boolean variable for `badge`.

But now that I think about it again, I might be wrong. This might be worth clarifying with Selim. @RakeshUP Could you please ask him how he envisioned it and try to align both his design and your implementation?

packages/ui-components/src/components/label/label.tsx
RakeshUP: It's capital Optional in the design

packages/ui-components/src/components/label/label.tsx
RakeshUP: According to the design, the badge was a boolean. 
The other text prop lets us pass i18n text label for the badge. We don't have i18n in ui-component package.

RakeshUP:
Things are a little ambiguous around the badge used in this component. I'm still going ahead and merging it, lets change it later when we have more clarity

PR #95, opened 2021-12-14
Fix/button wallet DAO-501

Fabricevladimir:
- Partial updates to `buttonWallet` previously `walletButton`
- Fix navbar height
- Added `avatarWallet` and `avatarDao`
- Renamed stories for better grouping


packages/ui-components/src/components/avatar/avatarWallet.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import styled from 'styled-components';
+import Blockies from 'react-blockies';
+
+import {IsAddress} from '../../utils/addresses';
+
+const SQUARES = 8;
+
+const styles: {
+  [key: string]: {style: string; scale: number};
```
WissenIstNacht: I liked the type that was on here before better:

`type SizesType = Record<AvatarProps['size'], {sizes: string, scale: number}>;`

You can choose a different type name for sure. But the `Record` with `AvatarProps['size']` as keys is more fitting here. The reason being that it constricts you to the actual values you want to use (i.e., the various sizes), instead of any string.

//Obviously, you can also just inline the Record type, no need to name it if you don't want to

//If you find yourself having issues because of `undefined` that you get from making the thingy optional, slap a good old `NonNullable` around it :D

packages/ui-components/src/components/daoSelector/daoSelector.tsx
```diff
@@ -47,11 +42,11 @@ type StyledButtonProps = {
 };
 
 // TODO: Switch font from bold to the appropriate weight (700)
```
WissenIstNacht: Still necessary?

packages/ui-components/src/components/avatar/avatarDao.tsx
```diff
@@ -0,0 +1,86 @@
+import React, {useCallback, useMemo, useState} from 'react';
```
WissenIstNacht, line 1: Thanks for refactoring this. Might have been a bit premature, though. 😅

Let's leave it like this for now. We can adapt to the refactored design later 👍🏻

WissenIstNacht (approved):
Thanks, things seem to be working :)

packages/ui-components/src/components/avatar/avatarWallet.tsx
Fabricevladimir: Hmm, these types are quite interesting.

packages/ui-components/src/components/button/buttonWallet.tsx
```diff
@@ -0,0 +1,64 @@
...
+    </StyledButton>
+  );
+};
+
+type StyledButtonProp = Pick<ButtonWalletProps, 'isLoading'>;
+
+const StyledButton = styled.button.attrs(({isLoading}: StyledButtonProp) => {
+  const className: string = `${
+    isLoading ? 'text-primary-500' : 'text-ui-600'
+  } flex items-center tablet:space-x-1.5 font-bold px-2 h-6 hover:text-ui-800 active:text-ui-800 disabled:text-ui-300 bg-ui-0 hover:bg-ui-100 active:bg-ui-200 disabled:bg-ui-100 rounded-xl focus:outline-none focus:ring-2 focus:ring-primary-500`;
```
sepehr2github: Can you break this class name line (too long lines) into 2 or 3 lines?

sepehr2github (approved):
Thanks, Fabrice! Just a minor suggestion.

PR #96, opened 2021-12-15
Feature/New components - wizard & linear progress DAO-531

RakeshUP:


packages/ui-components/src/components/progress/linearProgress.tsx
```diff
@@ -0,0 +1,16 @@
+
+import React from 'react';
+import styled from 'styled-components';
+
+export const LinearProgress = styled.progress.attrs({
+  className: 'h-1 w-full',
+})<React.ProgressHTMLAttributes<HTMLProgressElement>>`
+  ::-webkit-progress-bar {
```
WissenIstNacht, line 8: Soooooo apparently this is not yet officially supported for firefox :joy:

If you have a fix, go for it. Otherwise don't worry. Just drop a "FIXME" or"TODO" and we'll find a way to deal with it later. 👍🏻

packages/ui-components/tailwind.config.js
```diff
@@ -88,6 +88,7 @@ module.exports = {
     // overwirtes screen breakpoints according to design system
     screens: {
       tablet: '768px',
+      lg: '976px',
```
WissenIstNacht: _technically_ `desktop` is supposed to be `lg` haha.

The designs were lagging behind here when we settled on names and value, hence why it's wrong. Sorry for that. Could you change `976px` to `1280px`? No need to change the rest rn. We'll try to have a dedicated task for this at some point.

packages/ui-components/src/components/wizard/wizard.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const Wizard: React.FC<WizardProps> = ({processName, currentStep, totalSteps, progressPercentage, title, description}) => {
+  return (
+    <StepCard data-testid="wizard">
+      <CenteredFlex>
+        <p className="font-bold text-primary-500">
+          {processName}
+        </p>
+        <p className="text-ui-400">Step {currentStep} of {totalSteps}</p>
```
WissenIstNacht, line 21: Parts of this is going to be untranslated 😕

I'm honestly not sure how to deal with this... It seems stupid to pass in "Step" and "of" as a label. In particular since I'm not sure it even works in every language. Even with the locales and i10n I'm not sure this is simple easy to achieve. Maybe we should sync with Selim on that and ask him to stick with an "x/y" format for now...

No need for action, that's ok for now. But could you add a TODO here, too?

packages/ui-components/src/components/wizard/wizard.tsx
```diff
@@ -0,0 +1,46 @@
...
+import styled from 'styled-components';
+import { LinearProgress } from '../progress';
+
+export type WizardProps = {
+  processName: string;
+  currentStep: number;
+  totalSteps: number;
+  progressPercentage: number;
+  title: string;
+  description?: string;
```
WissenIstNacht: Don't think this is ever optional 🤔

WissenIstNacht (changes requested):
Thanks for exporting the components :)

packages/ui-components/src/components/progress/linearProgress.tsx
Fabricevladimir: ~For the moment it might be best to bite the bullet and just have it as a nested div with the size of the inner one being the percentage. Like this the style is stuck to these specific colors too...~ But I really like the progress element...

packages/ui-components/src/components/wizard/wizard.tsx
Fabricevladimir: Either that or just have the user pass in the whole string (not a fan of). But if we are passing in `currentStep` as well as `totalSteps`, the percentage should be calculated by the component itself. ` <LinearProgress max={totalSteps} value={currentStep} />`

Fabricevladimir (changes requested):
Good work!

packages/ui-components/src/components/progress/linearProgress.tsx
sepehr2github: I think it's possible to make it compatible with firefox by adding a -moz- prefix? I'm not sure

sepehr2github (commented):
As always you did a great job! Thank you.

packages/ui-components/tailwind.config.js
RakeshUP: Changed lg to 1280px. I found out this figma grid component just now, we'll change the viewports like in this file later
https://www.figma.com/file/ZHEzSel0rN5aFIpXjn3341/Aragon-Design-Tokens?node-id=213%3A5

packages/ui-components/src/components/progress/linearProgress.tsx
RakeshUP: Fixed that. It was possible when using -moz prefix

packages/ui-components/src/components/wizard/wizard.tsx
RakeshUP: removed progressValue prop. The component now calculates the progress using the current and total steps

RakeshUP: @WissenIstNacht Good catch. I've added a TODO. Maybe we could switch to x/y format. I'll inform Selim and see what he comes up with

WissenIstNacht (approved):

Fabricevladimir (approved):

PR #97, opened 2021-12-16
Feature/token selection modal DAO-536 

sepehr2github:
- Token Modal added
- WalletAugmentedContext updated / provider added
- erc20Token api added
- FetchBalance service added
- formatUnits function added for hexadecimal
- Network token config added
- getTokenInfo & isTokenERC20 function added utils/token
- react-hook-form package added

Note1: because there is no direct way to fetch a list of wallet tokens from ethers.js I had to check the balance for each token one by one

Note2: I'm still working on the integration of react-hook-form for search


packages/web-app/src/utils/tokens.ts
```diff
@@ -54,3 +57,47 @@ export function filterTokens(tokens: BaseTokenInfo[], searchTerm: string) {
 
   return tokens.filter(t => tokenInfoMatches(t, searchTerm));
 }
+
+export async function isTokenERC20(
```
WissenIstNacht, line 85: This method is unused.

Also, I'm confused what you wanted to do with it? When would you use it? I mean if you create a contract from an erc20 abi although the address is not an erc20 contract, you will get an error anyway?

Maybe I'm missing something, let me know! :)

packages/web-app/src/utils/tokens.ts
```diff
@@ -54,3 +57,47 @@ export function filterTokens(tokens: BaseTokenInfo[], searchTerm: string) {
...
+  address: string,
+  provider: EthersProviders.Provider
+) {
+  const contract = new ethers.Contract(address, erc20TokenABI, provider);
+  let decimals = null,
+    name = null,
+    symbol = null;
+
+  try {
+    decimals = await contract.decimals();
```
WissenIstNacht: Please put these calls into an array and pass it to `Promise.all()`. Then await the promise.

packages/web-app/src/utils/tokens.ts
```diff
@@ -54,3 +57,47 @@ export function filterTokens(tokens: BaseTokenInfo[], searchTerm: string) {
...
+    // eslint-disable-next-line no-empty
+  } catch (err) {}
```
WissenIstNacht: Please don't swallow errors.

If another dev uses this and something goes wrong in this method, it's going to be absolutely impossible to find the problem. At the _very least_ `console.error` the error or don't catch it at all. Ideally, extend `Error` properly and throw it. In _any case_ please document that the function can go wrong and what happens when it does.

packages/web-app/src/utils/network.ts
```diff
@@ -0,0 +1,23 @@
+import {networksType} from './types';
+
+/**
+ * A token list added to network config (needed for fetching token balances)
+ */
+export const networks: networksType = {
```
WissenIstNacht: None of these are `networks` 😅

See my comments in `type.ts` for more detail.

packages/web-app/src/utils/types.ts
```diff
@@ -35,3 +35,23 @@ export type HookData<T> = {
...
+export type ChainId = 1 | 4;
+export type TokenSymbol = 'DAI' | 'USDT' | 'USDC';
+
+export interface EnvironmentConfig {
+  chainId: ChainId;
+  networkName: string;
+  curatedTokens: Record<TokenSymbol, string>;
+}
+
+export type networksType = {
```
WissenIstNacht: 1) Types are capitalized
2) The information in this type does not describe a network
3) I don't think this is necessary. You can put this directly into environmentConfig. By this I mean: You can directly put these tokens into the `curatedTokens` `Record` in `EnvironmentConfig`.

packages/web-app/src/utils/types.ts
```diff
@@ -35,3 +35,23 @@ export type HookData<T> = {
   isLoading: boolean;
   error?: Error;
 };
+
+export type ChainId = 1 | 4;
+export type TokenSymbol = 'DAI' | 'USDT' | 'USDC';
+
+export interface EnvironmentConfig {
+  chainId: ChainId;
+  networkName: string;
```
WissenIstNacht: `networkName` is not necessary, as it can always be accessed form `useWallet` (there is information in there for every chain).

packages/web-app/src/utils/types.ts
```diff
@@ -35,3 +35,23 @@ export type HookData<T> = {
   isLoading: boolean;
   error?: Error;
 };
+
+export type ChainId = 1 | 4;
```
WissenIstNacht:
```suggestion
export type SupportedChainId = 1 | 4;
```

packages/web-app/src/services/amount.ts
```diff
@@ -0,0 +1,21 @@
+import {Contract, providers as EthersProviders} from 'ethers';
+import {erc20TokenABI} from 'utils/abis';
+import {formatUnits} from 'utils/library';
+import {getTokenInfo} from 'utils/tokens';
+/**
```
WissenIstNacht: space (it's free! :D)

packages/web-app/src/context/augmentedWallet.tsx
```diff
@@ -1,23 +1,36 @@
...
+import React, {useContext, useEffect, useMemo, useState} from 'react';
 import {identifyUser} from 'services/analytics';
 import {UseWalletProvider, useWallet} from 'use-wallet';
 import {Wallet} from 'use-wallet/dist/cjs/types';
 import {providers as EthersProviders} from 'ethers';
 import {updateAPMContext, useAPM} from './elasticAPM';
+import {INFURA_PROJECT_ID} from 'utils/constants';
+
+type WalletAugmented = Wallet & {
+  isConnected: boolean;
```
WissenIstNacht: `useWallet` already provides `isConnected` and `account`, no need to retype them here.

packages/web-app/src/context/augmentedWallet.tsx
```diff
@@ -1,23 +1,36 @@
...
+}
 
 const WalletAugmented: React.FC<unknown> = ({children}) => {
   const wallet = useWallet();
   const ethereum: any = wallet.ethereum;
+  const fallbackProvider = new EthersProviders.InfuraProvider(
+    wallet.chainId, // set provider based on wallet chain id
+    INFURA_PROJECT_ID
+  );
+  const [provider, updateProvider] =
```
WissenIstNacht: please stick to the typical react style and call this `setProvider`, so everyone knows this function belongs to `useState`.

packages/web-app/src/context/transfersModal.tsx
```diff
@@ -11,27 +11,56 @@ const TransferModalContext = createContext<TransferModalContextType | null>(
...
+      case 'token':
+        setIsTokenOpen(false);
+        break;
+      default:
+        setIsTransferOpen(false);
+        break;
+    }
+  };
+  /**
+   * TODO: ==============================================
```
WissenIstNacht, line 50: Yes, so I guess this is an open discussion at this point. I personally would manage all the modals in one context, since they are all related to one another (in the sense that you can never two modals open at the same time). In other words, I agree with the changes you made here. I would even go on step further and expand this for every type of modal we have. 

But this is totally ok for now 👍🏻

packages/ui-components/src/components/input/SearchInput.tsx
```diff
@@ -1,17 +1,18 @@
 import React from 'react';
 import {TextInput, TextInputProps} from './textInput';
-import {IconSearch} from '../icons';
+import {IconSearch, IconClose} from '../icons';
 
 /** Simple input with variable styling (depending on mode) */
 
 export type SearchInputProps = Omit<TextInputProps, 'adornment' | 'side'>;
```
WissenIstNacht: `side` and `adornment` both no longer exist on the parent type.

packages/ui-components/src/components/input/textInput.tsx
```diff
@@ -5,39 +5,51 @@ export type TextInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
...
-  adornment?: ReactNode;
+  rightAdornment?: ReactNode;
+  leftAdornment?: ReactNode;
   /**
-   * Wheter the icon is left or right of the input
-   */
-  side: 'left' | 'right';
+  * adornments click events
+  */
+  rightAdornmentOnClick?: () => void;
```
WissenIstNacht: This is a tricky one. I understand why you did this, but I'm not sure it was a good idea. The reason I'm saying this is because the interface of this component is now completely different from the designs. For example, the `SearchInput` with the cross on the right hand side is just the active state of the component. It's not a text input with two adornments.

In other words, we will have to work this completely in the future. I don't want to throw your hard work away now, so we'll keep it as is now. But in the future, I would please ask you not to change components if they are different than on the design (If any one gives you trouble for this send them ot me. I'll explain to them why it's not your fault). In general, just work with what you have (even if it's not completely right), and then do a complete refactor later. If what you have is really unsuable (i.e., completely different from the designs), then please create a separate ticket and make the changes before you need them.

packages/web-app/src/services/amount.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param tokenAddress address of token contract
+ * @param ownerAddress owner address / wallet address
+ * @param provider interface to node
+ * @returns a promise that will return a balance amount
+ */
+export const FetchBalance = async (
```
WissenIstNacht: function name should be lower case.

packages/web-app/src/services/amount.ts
```diff
@@ -0,0 +1,21 @@
+import {Contract, providers as EthersProviders} from 'ethers';
+import {erc20TokenABI} from 'utils/abis';
+import {formatUnits} from 'utils/library';
+import {getTokenInfo} from 'utils/tokens';
+/**
+ * @param tokenAddress address of token contract
+ * @param ownerAddress owner address / wallet address
+ * @param provider interface to node
+ * @returns a promise that will return a balance amount
```
WissenIstNacht: fetchBalance should return a single numer value that is the balance of the `walletAddress` in the `contractAddress`.

Other information should be separated.

packages/web-app/src/utils/tokens.ts
```diff
@@ -54,3 +57,47 @@ export function filterTokens(tokens: BaseTokenInfo[], searchTerm: string) {
...
+  const contract = new ethers.Contract(address, erc20TokenABI, provider);
+  try {
+    await Promise.all([contract.balanceOf(address), contract.totalSupply()]);
+    return true;
+  } catch (err) {
+    return false;
+  }
+}
+
+export async function getTokenInfo(
```
WissenIstNacht, line 105: Why is this method necessary? Can't we get this method through the APIs we use?

packages/web-app/src/containers/tokenMenu/tokenBox.tsx
```diff
@@ -0,0 +1,58 @@
...
+  useEffect(() => {
+    // Fetch balance amount for each token
+    FetchBalance(tokenAddress, account, provider).then(balanceRes => {
+      setBalance(balanceRes.amount);
+      setSymbol(balanceRes.symbol);
+    });
+  }, [account, provider, tokenAddress]);
+
+  // This condition will change later with skeleton loading integration
+  return balance && balance !== '0.0' ? (
```
WissenIstNacht: I think ethers provides an isZero check, no?

packages/web-app/src/containers/tokenMenu/tokenBox.tsx
```diff
@@ -0,0 +1,58 @@
...
+import {useWallet} from 'context/augmentedWallet';
+import {FetchBalance} from 'services/amount';
+
+export type TokenProps = {
+  tokenAddress: string;
+  tokenName: string;
+  tokenLogo: string;
+};
+
+export default function TokenBox({
```
WissenIstNacht, line 14: According to the design, this should be a `ListItemToken`. It should be part of the ui library and not in the web-app.

Do not change this as part of this PR, but make create a separate subtask ticket. Ideally, this should have been done before this PR.

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+import {useWallet} from 'context/augmentedWallet';
+import {useTransferModalContext} from 'context/transfersModal';
+import {networks} from 'utils/network';
+import TokenBox from './tokenBox';
+import {useForm, Controller} from 'react-hook-form';
+
+const TokenMenu: React.FC = () => {
+  const {isTokenOpen, close} = useTransferModalContext();
+  const {chainId}: Wallet = useWallet();
+  const {control, watch, resetField} = useForm({
```
WissenIstNacht: This looks like a VERY complicated way of doing something that can be done in 2 lines of code :D.

Is this really necessary for a simple typeahead search?

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+              rightAdornmentOnClick={ResetSearchBox}
+            />
+          )}
+          name="searchToken"
+          control={control}
+          defaultValue=""
+        />
+        <TokensWrapper>
+          <TokensTitle>Your Tokens</TokensTitle>
+          {Object.entries(networks[chainId || 4].curatedTokens)
```
WissenIstNacht: I'm not sure about the data / types in this component. I've mentioned this somewhere else already, but hy fetch this data explicitly? Don't we fetch this information already? (and by "already" I mean, as soon as we are connected to our APIs. For now you would have to mock like we did for other tokens.)

WissenIstNacht (changes requested):
Hey, thanks for that huge piece of work! Lots of new and tricky things in there!

However, there's quite a few things I don't understand why you did them the way you did. Also, dividing your work into smaller, indenpendent tasks would really have helped here.

If there's anything unclear in my comments, anything you don't understand or you think I didn't understand, please let me know.

packages/web-app/src/containers/tokenMenu/tokenBox.tsx
sepehr2github: yes but It works with a BigNumber  format

packages/web-app/src/containers/tokenMenu/index.tsx
sepehr2github: You were right and I changed the name of the list. but let's keep it in that way because once the spike task is done we can do a refactor if needed! let me know if you have further ideas. Thank you!

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+import {useWallet} from 'context/augmentedWallet';
+import {useTransferModalContext} from 'context/transfersModal';
+import {networks} from 'utils/network';
+import TokenBox from './tokenBox';
+
+const TokenMenu: React.FC = () => {
+  const {isTokenOpen, close} = useTransferModalContext();
+  const {chainId}: Wallet = useWallet();
+  const {control, watch} = useForm();
```
Fabricevladimir, line 16: There is no real need to have this as a separate form. Leave it in for now, but add a todo so that when we wire the form together we remember to integrate it into the main form.

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const {t} = useTranslation();
+
+  const filterValidator = useCallback(
+    (token: string[]) => {
+      const searchValue = watch('searchToken');
```
Fabricevladimir, line 21: Careful with using the `watch` function because it causes an entire rerender. Maybe use the `useWatch` hook instead or remove the search logic from the form and have a piece of state for it.

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  );
+
+  const renderResult = () => {
+    const tokenList = Object.entries(
+      networks[chainId || 4].curatedTokens
+    ).filter(filterValidator);
+
+    return tokenList.length !== 0 ? (
+      <>
+        <TokenTitle>{t('TokenModal.yourToken')}</TokenTitle>
```
Fabricevladimir: Small typo, should be `yourTokens`

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+        {tokenList.map(([name, address]) => (
+          <TokenBox
+            tokenName={name}
+            tokenAddress={address}
+            tokenLogo={
```
Fabricevladimir, line 43: Are we taking care of tokens without a logo? If not drop a todo. Once the avatars have been finalized we'll have one that automatically does this for us.

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  //TODO: tokenLogo should be automate using coinkego api
+  //TODO: Cross Icon should added in the next released
+  return (
+    <Modal
+      open={isTokenOpen}
+      onClose={() => close('token')}
+      data-testid="TokenMenu"
+    >
+      <Container>
+        <Controller
```
Fabricevladimir, line 67: I don't think the search term should be stored in the form considering we never actually use it outside of this menu. If a token custom token is being added, we can use `setvalue('tokenName', searchTerm)` to store the token name in the form data when the add custom token button is clicked.

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+const TokenTitle = styled.h2.attrs({
+  className: 'text-base font-bold',
+})``;
+
+const TokenSubtitle = styled.h2.attrs({
+  className: 'text-sm',
+})``;
+
+const WideButton = styled(ButtonText).attrs({
+  className: 'w-full justify-center',
```
Fabricevladimir, line 109: When we make an iteration through the ui-components, I'll update the styling for the button to center the contents by default. Thanks!

packages/web-app/src/services/amount.ts
```diff
@@ -0,0 +1,22 @@
+import {Contract, providers as EthersProviders} from 'ethers';
```
Fabricevladimir: Curious, why have this as a service as apposed to a function in the token utils file?

packages/web-app/src/utils/library.ts
```diff
@@ -0,0 +1,9 @@
+// Library utils / Ethers for now
```
Fabricevladimir, line 1: What is the purpose of this separate utils file? Like what do we put in it?

packages/web-app/src/utils/tokens.ts
```diff
@@ -54,3 +57,40 @@ export function filterTokens(tokens: BaseTokenInfo[], searchTerm: string) {
...
+}
+/**
+ * This Function is necessary because
+ * you can't fetch decimals from the api
+ *
+ * @param address token contract address
+ * @param provider Eth provider
+ * @returns number for decimals for each token
+ */
+export async function getTokenDecimals(
```
Fabricevladimir: For the custom tokens, we don't have a token name... It would be nice if this could return the name and symbol as well. I realized this just now 😓

Fabricevladimir (commented):
Massive task! Nice work

packages/web-app/src/containers/tokenMenu/index.tsx
sepehr2github: In this component because with each input onchange event list should update using useWatch doesn't make so much difference

packages/web-app/src/containers/tokenMenu/index.tsx
sepehr2github: Thank you!

packages/web-app/src/services/amount.ts
sepehr2github: I thought because It's fetching data it can categorize as a service but you're right! utils/token definitely is a better place.

packages/web-app/src/utils/library.ts
sepehr2github: Some ethers numeric utils functions like formatUnits, parseUnits, toBigNum, etc...
reference: https://github.com/aragon/govern/blob/develop/packages/govern-console/src/utils/lib.tsx

packages/web-app/src/containers/tokenMenu/index.tsx
sepehr2github: Yes, maybe🤔. when the token selection logic was complete (Deposit flow) we can do a double-check and change it if needed

packages/web-app/src/containers/tokenMenu/tokenBox.tsx
```diff
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import {Avatar} from '@aragon/ui-components';
```
Fabricevladimir, line 3: Aye my friend, I am sorry to say that this will be refactored later because it's an official ui-component now (listItemToken)! Not now of course

Fabricevladimir (approved):
Looks good!

WissenIstNacht (approved):
Many thanks for going through all the comments!

PS: I have a two comments regarding the search. Things to keep in mind for later. 
- As I mentioned earlier: I agree with Fabrice that this could be simplified to just reading the value from the input. I don't think the extra form is necessary. It should be enough to read the value from the input can be used it directly as a search term. 
- More importantly, there's another related issue here. Due to the way the token data is structured, you can only search for the token address or token symbol, which is not ideal. In other words, "typing Ara" would not find "ANT". This should be corrected at some point. For inspiration, you cna check how this is done on the "All token" page.

packages/web-app/src/containers/tokenMenu/tokenBox.tsx
sepehr2github: Yes my friend. I know the tokenListItem is my Pr

PR #98, opened 2021-12-16
Feature: Transfer filter - DAO 481

WissenIstNacht:
This PR

- Adds additional mock transfer data
- Categorizes it by date using the corresponding hook
- Adds a button group to `/finance/transfers`
- Adds logic to filter transfer by type using the button group
- Fixes the current deployment issue

### DIFF
--- a/packages/web-app/src/components/wrappers/sectionWrappers.tsx
+++ b/packages/web-app/src/components/wrappers/sectionWrappers.tsx
@@ -4,6 +4,7 @@ import {Link} from 'react-router-dom';
 import {ButtonText, IconChevronRight} from '@aragon/ui-components';
 
 import {AllTokens, AllTransfers} from 'utils/paths';
+import {useTranslation} from 'react-i18next';
 
 export type SectionWrapperProps = {
   title: string;
@@ -20,6 +21,8 @@ export type SectionWrapperProps = {
  * define this.
  */
 export const TokenSectionWrapper = ({title, children}: SectionWrapperProps) => {
+  const {t} = useTranslation();
+
   return (
     <>
       <HeaderContainer>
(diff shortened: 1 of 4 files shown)


sepehr2github (approved):
Looks nice! thank you.

packages/web-app/src/components/wrappers/sectionWrappers.tsx
```diff
@@ -4,6 +4,7 @@ import {Link} from 'react-router-dom';
 import {ButtonText, IconChevronRight} from '@aragon/ui-components';
 
 import {AllTokens, AllTransfers} from 'utils/paths';
+import {useTranslation} from 'react-i18next';
```
Fabricevladimir, line 7: Just how did we go along without this? Lost in a merge perhaps?

packages/web-app/src/hooks/useCategorizedTransfers.tsx
```diff
@@ -1,27 +1,144 @@
...
+          break;
+        default:
+          week.push(t);
+          break;
+      }
+    });
+    setCategorizedTransfers({
+      week: week,
+      month: month,
+      year: year,
```
Fabricevladimir: If the name of the value and the key are the same, you can use the shorthand syntax like so `{week, month, year}`

Fabricevladimir (approved):
Looks great! I noticed that there was an issue with the button group (yay changes), should we be adding refactoring tasks in the backlog so we don't forget?

packages/web-app/src/components/wrappers/sectionWrappers.tsx
WissenIstNacht: Asked myself the same question...

packages/web-app/src/hooks/useCategorizedTransfers.tsx
WissenIstNacht: Thanks, must have missed that!

WissenIstNacht:
I'm not sure yet how we'll deal with improvements / refactoring... If we all just dump it into backlog I think we'll have a bad time haha.

PR #99, opened 2021-12-17
Feature/Historical data  DAO-472 

Fabricevladimir:
- fetch historical data from external api
- updated token types
- populated predefined data flow

When reviewing please check especially if promises are being resolved efficiently.
Also, suggestions concerning the `useDaoTreasury` hook are very welcomed. I vacillated between using the hook to manage data and having it in a separate context that would then be consumed by the relevant finance pages.


RakeshUP (approved):
Looks good!
BTW what's the purpose of 'reset' param in useInterval hook?

And also take a look at the apollo-client-poc. The kinds of nested API queries you've made using multiple promise.all becomes easy to read and understand when using Apollo REST API Link

Fabricevladimir:
> Looks good! BTW what's the purpose of 'reset' param in useInterval hook?
> 
> And also take a look at the apollo-client-poc. The kinds of nested API queries you've made using multiple promise.all becomes easy to read and understand when using Apollo REST API Link

The `reset` is an on the fly way to well, re-set the interval. Because the data passed into the `useInterval` callback isn't immediately available and is being fetched, the first execution of the callback is polling prices of an empty array. Reset was just a quick way to tell it when the data has changed, ie, the array now has tokens. 

The better approach though would be to check if the list of tokens is greater than zero, then call `useInterval` because currently, it sets the interval even when there is no data.

I have taken a quick look at it and am sort of trying to do a deep dive right now. I believe it should take care of all of what we're currently doing in this PR 😄

packages/web-app/src/pages/transfers.tsx
```diff
@@ -1,14 +1,16 @@
...
+
 import TransferList from 'components/transferList';
-import {Transfers} from 'utils/types';
+import {PageWrapper} from 'components/wrappers';
+import {useTransferModalContext} from 'context/transfersModal';
+
+import type {Transfer} from 'utils/types';
 
-const transfers: Array<Transfers> = [
+const allTransfers: Array<Transfer> = [
```
WissenIstNacht, line 13: FYI: I expanded this mock data for one of my PRs. I also moved it to the file with the hook that serves it to the application (`hooks/somethingTransfers`). Will probably come up as a conflict when merging. Would be nice if you could keep the incoming changes. But feel free to change any `Transfers` to `Transfer`, like you did in this PR.

WissenIstNacht: FYI: I expanded this mock data for one of my PRs. I also moved it to the file with the hook that serves it to the application (hooks/somethingTransfers). Will probably come up as a conflict when merging. Would be nice if you could keep the incoming changes. But feel free to change any Transfers to Transfer, like you did in this PR.

packages/web-app/src/pages/finance.tsx
```diff
@@ -3,19 +3,19 @@ import {useTranslation} from 'react-i18next';
...
-import {useTokenInfo} from 'hooks/useTokenInformation';
+import {TimeFilter} from 'utils/constants';
+import {useDaoTreasury} from 'hooks/useDaoTreasury';
 import {useTransferModalContext} from 'context/transfersModal';
-import {Transfers} from 'utils/types';
 
-const TEMP_TRANSFERS: Transfers[] = [
+import type {Transfer} from 'utils/types';
+
+const TEMP_TRANSFERS: Transfer[] = [
```
WissenIstNacht, line 18: Same here, reminder that this data has been moved in the meantime.

packages/web-app/src/hooks/useInterval.tsx
```diff
@@ -6,10 +6,12 @@ const DEFAULT_INTERVAL = 300000; //interval in ms
  * Hook running callback at specified interval
  * @param callback Function to run after each interval. Defaults to 300000ms
  * @param interval Delay in milliseconds
+ * @param reset Resets hook; used callback expects async data
  */
 const useInterval = (
   callback: () => void,
-  interval: number = DEFAULT_INTERVAL
+  interval: number = DEFAULT_INTERVAL,
+  reset?: boolean
```
WissenIstNacht, line 14: So I think I understand what `useInterval(bla, bla, true)` should do if you call it just once. But if you call it like this again, nothing's gonna happen, righ? On other hand, calling it a second time with `useInterval(bla, bla, false)` will do what you expect `useInterval(bla, bla, true)` to do...?

I am confusion 😅

WissenIstNacht (approved):
Thanks mate, well done. The information structure looks much better already and the live data seems to work well.

There's a couple of things that seem a bit convoluted in the hooks department, but it's ok. I haven't found anything that's flat out wrong. Except maybe for the `useInterval` hook. Not sure about the reset prop.

Also, left a comment where potential merge conflicts could arise.

packages/web-app/src/pages/tokens.tsx
```diff
@@ -1,45 +1,50 @@
...
+    searchTerm
+  );
 
   return (
     <Layout>
       <PageWrapper
         title={t('TransferModal.allTransfers') as string}
         buttonLabel={t('TransferModal.newTransfer') as string}
-        subtitle={'$1,002,200.00 Total Volume'}
+        subtitle={`${new Intl.NumberFormat('en-US', {
```
sepehr2github, line 36: It would be great if you add a utils function named "number format" and call the function

packages/web-app/src/services/prices.ts
```diff
@@ -1,47 +1,101 @@
+import {Address} from '@aragon/ui-components/dist/utils/addresses';
 import {constants} from 'ethers';
 
-const BASE_URL = 'https://api.0x.org';
-const BUY_TOKEN = 'USDC';
-const DEFAULT_SELL_AMOUNT = '1000000000000000000';
+import {TokenPricePercentages} from 'utils/types';
+
+const BASE_URL = 'https://api.coingecko.com/api/v3';
```
sepehr2github: Also here please add all URLs in the constants folder. So all external links would be in one place

sepehr2github (approved):
Thank you, Fabrice. and well done!
Just some minor suggestions

packages/web-app/src/pages/tokens.tsx
Fabricevladimir: Yeah, not a bad idea. It's the same thing for the currency. The only issue is that I don't want it to turn into a reinvention of the wheel considering we have 4 different formats for numbers. If we end up with a function api that's similar to Intl.NumberFormat, then there is essentially no real purpose. But yeah, all the separate formattings are a code smell

packages/web-app/src/services/prices.ts
Fabricevladimir: Sure thing

packages/web-app/src/hooks/useInterval.tsx
Fabricevladimir: Yeah, that one needs revisited. The intention is to indeed reset the interval when the data set has changed. While I could compare the previous and the new data, it gets very inefficient as the data grows...

PR #100, opened 2021-12-20
Feature/Deposit state flow - DAO-543 

Fabricevladimir:
- installed react-hook-forms
- setup deposit stepper
- stubs to show usage

Update: Now using separate deposit and withdraw pages. 
Note: Some thought should be given to what happens if the user disconnects their wallet mid-process


sepehr2github (changes requested):
Thank you! Unfortunately, App crashes on /finance/new-transfer 
image

Can you check it? Thanks

Fabricevladimir:
> Thank you! Unfortunately, App crashes on /finance/new-transfer image
> 
> Can you check it? Thanks

The transfer type is being read from the state, so navigating directly will crash (for now). I left a few comments about this in the code.

packages/web-app/src/utils/constants.ts
```diff
@@ -19,3 +19,8 @@ export const NAV_LINKS = [
   {label: 'Finance', path: Finance},
   {label: 'Community', path: Community},
 ];
+
+export enum TransferTypes {
```
WissenIstNacht: Ah yes, our dear ol' friends, enums. 😋

I don't this will ever become a bottleneck for performance, but: be aware that enums generate _really_ ugly JS code.

There's a cool sort of middle-ground that mitigates this but still allows to have enums: using the `const` enums. I.e.:

```
const enum TransferTypes {
  Deposit = 'Deposit',
  Withdraw = 'Withdraw',
}
```

Obviously this only works for examples like this one, where the enums are completely defined at compile time (and wont be changed later).

Don't worry about it though :)

packages/web-app/src/containers/transferMenu/index.tsx
```diff
@@ -1,15 +1,27 @@
...
-  const {isOpen, close} = useTransferModalContext();
   const {t} = useTranslation();
   const navigate = useNavigate();
+  const {isConnected} = useWallet();
+  const {isOpen, close} = useTransferModalContext();
+
+  const handleNewDepositClick = () => {
+    // TODO: change alert to proper error reporting mechanism,
+    // Move to proper placing
+    if (isConnected()) {
```
WissenIstNacht, line 20: Nice one, didn't think about this while reviewing Rakesh's PR.

packages/web-app/src/pages/newDeposit.tsx
```diff
@@ -1,31 +1,95 @@
...
+  amount: 0,
+  reference: '',
+  tokenAddress: '',
+  tokenSymbol: '',
+};
 
 const NewDeposit: React.FC = () => {
   const {t} = useTranslation();
+  const {open} = useWalletMenuContext();
+  const [currentStep, setStep] = useState<Steps>(Steps['Configure Deposit']);
```
WissenIstNacht: opportunity for a cool little hook: 

```
const [prev, curr, next] = useStepper(TOTAL_STEP_COUNT); 
```

where `prev` and `next` are essentially the callbacks from below, and curr is the `currentStep` (and all is managed by the same `useEffect` that you already defined).

Totally optional, though 👍🏻

packages/web-app/src/pages/newDeposit.tsx
```diff
@@ -57,60 +121,39 @@ const NewDeposit: React.FC = () => {
...
-          <FormItem>
-            <Label
-              label={t('labels.reference')}
-              helpText={t('newDeposit.referenceSubtitle')}
-              isOptional={true}
-            />
-          </FormItem>
-
-          <div className="flex justify-between mt-8">
+          {currentStep === Steps['Configure Deposit'] ? (
```
WissenIstNacht: Out of curiosity: Have you ever used `react-if`?

Cool little package that allows to write conditional statements in JSX. Some people don't like it, though...

WissenIstNacht (approved):
Thanks, looks good to me!

Left some comments about incosequential stuff. Feel free to make changes or merge the PR 👍🏻

sepehr2github (approved):
Thank you, Fabrice. awesome as always.

packages/web-app/src/pages/newDeposit.tsx
Fabricevladimir: I haven't, but I have always wanted to use if in JSX! I think that will be available with the new version of React.

packages/web-app/src/pages/newDeposit.tsx
Fabricevladimir: This is a good idea, especially since the Withdraw and Deposit are fully separated right now. We'll need a stepper for Withdraw as well.

packages/web-app/src/containers/transferMenu/index.tsx
Fabricevladimir: See this takes care of the modal, but not all the situations which begs the question, should we be gatekeeping here at all or move that logic to the Deposit page itself since a user can technically navigate to the new-deposit page without connecting their wallet...

packages/web-app/src/utils/constants.ts
Fabricevladimir: Thanks for pointing that out. I wasn't aware!

packages/web-app/src/containers/transferMenu/index.tsx
WissenIstNacht: Another good point... Another thing for Selim to figure out haha

PR #101, opened 2021-12-21
Feature/TransferCard - DAO-539 

Fabricevladimir:
- TransferCard (to/from) web-app component
- Fixed transfers test issue

Note: Component shown on web-app homepage


packages/web-app/src/locales/en/common.json
```diff
@@ -40,11 +39,15 @@
...
   "TransferModal": {
-    "allTransfers":"All Transfers",
+    "allTransfers": "All Transfers",
     "newTransfer": "New Transfer",
     "item1Title": "Deposit assets",
     "item1Subtitle": "Add assets to the DAO treasury",
     "item2Title": "Withdraw assets",
     "item2Subtitle": "Add assets to the DAO treasury"
+  },
+  "TransferCard": {
```
WissenIstNacht: Could go to the labels category :D

The only reaosn I'm pointing this out is because I just reviewed Rakesh's PR where he put a "to" into that the `labels` object. Both are fine and/or can be changed later though :)

packages/web-app/src/components/transferCard/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  className: 'flex-1 py-1.5 px-2 min-w-0 text-left bg-ui-0 rounded-xl',
+})``;
+
+const Label = styled.p.attrs({
+  className: 'text-sm text-ui-500',
+})``;
+
+const Address = styled.p.attrs({
+  className:
+    'overflow-hidden font-bold text-ui-800 overflow-ellipsis whitespace-nowrap',
```
WissenIstNacht: I think Sepehr introduced an address shortening function a while back that puts the ellipsis in the middle. I think this would be more aligned with other parts of the design (like the address on the wallet button).

WissenIstNacht (changes requested):
That was refreshingly easy to review :D

Thanks for the PR. I left a comment on a minor issue with the shortening of addresses. The rest look great!

packages/web-app/src/locales/en/common.json
RakeshUP: Fine for now. We could later move it to labels object(create a label inside finance object if it makes sense)

packages/web-app/src/utils/strings.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Capitalize the first letter of a string
+ * @param string - string to capitalize
+ * @param locale - desired language; defaults to browser language.
+ * @returns string with first letter capitalized
+ */
+export function capitalize(
```
RakeshUP: We could use css for this `text-transform: capitalize`

https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

packages/web-app/src/components/transferCard/index.tsx
RakeshUP: The function mentioned is being used in the ButtonWallet I guess

RakeshUP (approved):
Looks good! 
Can be merged after capitalize and wallet address ellipsis functions are taken care of

RakeshUP:
Woah too many merge conflicts!!

packages/web-app/src/utils/strings.ts
Fabricevladimir: I wasn't aware this would be language specific. Awesome suggestion!

packages/web-app/src/components/transferCard/index.tsx
Fabricevladimir: That function returns a shortened address. When the screen size gets smaller, it won't shrink accordingly. I'll add it here with a todo because the overflow-ellipsis is still needed for like a long ens or DAO name

WissenIstNacht (approved):

PR #102, opened 2021-12-22
Feature/Textarea Simple component DAO-547

RakeshUP:


packages/ui-components/__tests__/textarea.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+
+import {TextareaSimple} from '../src';
+
+describe('TextareaSimple', () => {
+  function setup(args: any) {
+    render(<TextareaSimple {...args} />);
+    return screen.getByRole('textbox');
```
Fabricevladimir, line 9: Whoa I wasn't aware of this selector! Nice

packages/ui-components/src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,23 @@
+import styled from 'styled-components';
+
+export type TextareaSimpleProps = {
+};
+
+export const TextareaSimple = styled.textarea.attrs({
+  className: `py-1.5 px-2 rounded-xl resize-none w-full border-2 border-ui-100 hover:border-ui-300 
+    disabled:bg-ui-100 disabled:border-ui-200 focus:border-primary-500 bg-white text-ui-600`,
```
Fabricevladimir, line 8: Missed the active state and the ever-present focus ring

Fabricevladimir (approved):
Nice work! Good to merge once the focus ring is added.

packages/ui-components/src/components/textarea/textarea.tsx
```diff
@@ -0,0 +1,23 @@
...
+  style: {minHeight: 144},
```
sepehr2github: You could set this style in the backtick section :D.

sepehr2github (approved):
Hooray! The new textarea components are ready! Thank you 👍

packages/ui-components/src/components/textarea/textarea.tsx
RakeshUP: So I saw the designs and there are two variants, focus and active for textarea. But if we specify focus styles, active is not applied at all. And active according to the designs looks like when the value is being entered in the input. I've put the active spec in design in the focus variant in code. 

I'll have to talk to Selim and see what to do about this

PR #103, opened 2021-12-22
Feature: 5 largest tokens - DAO-468

WissenIstNacht:
This PR:

- extends mock data for token
- shows the 5 largest tokens (by treasury share) in the token overview on `/finance`

### DIFF
--- a/packages/web-app/src/pages/finance.tsx
+++ b/packages/web-app/src/pages/finance.tsx
@@ -14,7 +14,10 @@ import {useDaoTreasury} from 'hooks/useDaoTreasury';
 import {useTransferModalContext} from 'context/transfersModal';
 
 import type {Transfer} from 'utils/types';
+import {sortTokens} from 'utils/tokens';
 
+// TODO remove this. Instead use first x transfers returned by categorized
+// transfers hook.
 const TEMP_TRANSFERS: Transfer[] = [
   {
     title: 'Deposit',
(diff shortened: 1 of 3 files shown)


RakeshUP (approved):

packages/web-app/src/pages/finance.tsx
```diff
@@ -49,6 +52,10 @@ const Finance: React.FC = () => {
   const {open} = useTransferModalContext();
   const {data: treasury} = useDaoTreasury('0xMyDaoAddress', TimeFilter.day);
 
+  sortTokens(treasury.tokens, 'treasurySharePercentage');
+  treasury.tokens.reverse();
```
Fabricevladimir: Why reversing after having sorted? Does that not indicate an issue in the sorting itself?

Fabricevladimir (commented):
Nice. I like the ability to search by different properties.

packages/web-app/src/pages/finance.tsx
WissenIstNacht: "issue" is kind of a harsh adjective 😅 

I don't think there's an issue with the sorting method, as it does what it says. But yeah, maybe I could expand its API to allow for a reverse ordering.

Fabricevladimir: Sorry, no harshness intended! 😅 

Ascending & descending would be great to have, but don't worry about that now. Another approach would maybe be to reverse the order in the search itself so instead of `a < b return -1` we could do `a > b return -1`, but I don't know what unintended consequences it might have on different values passed in.

Fabricevladimir (approved):

PR #104, opened 2021-12-22
Feature/Input Value component DAO-546

RakeshUP:


packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+)<StyledContainerProps>`
+  padding: 6px 6px 6px 16px;
+
+  :focus-within {
+    border-color: #003BF5;
+  }
+`;
+
+const StyledButton = styled.button.attrs({
```
Fabricevladimir: This should be ButtonText secondary, size small, bgWhite.

packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import styled from 'styled-components';
+import { StyledInput } from './textInput';
+
+export type ValueInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
+  /* Text that appears on the button present on the right side of the input  */
+  buttonText: string;
+
+  onButtonClick: () => void;
```
Fabricevladimir: I saw that the text input named the element adornment, can we agree on one name for the sake of consistency?

packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+type StyledContainerProps = Pick<ValueInputProps, 'mode' | 'disabled' >;
+
+export const Container = styled.div.attrs(
+  ({mode, disabled}: StyledContainerProps) => {
+    let className = `${disabled? 'bg-ui-100 border-ui-200' : 'bg-ui-0'} flex items-center space-x-1.5 
+      text-ui-600 rounded-xl border-2 hover:border-ui-300 `;
+
+    if (mode === 'default') {
```
Fabricevladimir, line 39: Not requesting a change for this. It seems we're going to use the same border styles for all the inputs, might be worth having that in one place.

packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+    } else if (mode === 'warning') {
+      className += 'border-warning-600';
+    } else if (mode === 'critical') {
+      className += 'border-critical-600';
+    }
+
+    return {className};
+  }
+)<StyledContainerProps>`
+  padding: 6px 6px 6px 16px;
```
Fabricevladimir: Can we switch this over to TW instead of straight CSS?  Documentation for the focus within can be found [here](https://v2.tailwindcss.com/docs/hover-focus-and-other-states#focus-within).

Fabricevladimir (changes requested):
Good work!

packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { StyledInput } from './textInput';
+
+export type ValueInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
+  /* Text that appears on the button present on the right side of the input  */
+  buttonText: string;
+
+  onButtonClick: () => void;
+
+  /** Changes a input's color schema */
+  mode?: 'default' | 'success' | 'warning' | 'critical';
```
sepehr2github, line 13: 'mode' prop is optional here but you never defined a default value!

packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  :focus-within {
+    border-color: #003BF5;
+  }
+`;
+
+const StyledButton = styled.button.attrs({
+  className: 'text-sm px-1.5 rounded-lg bg-ui-50 font-bold text-ui-700 disabled:bg-ui-100 disabled:text-ui-300',
+})`
+  padding-top: 6px;
```
sepehr2github: Also for padding please use tailwind classes. If needed you can define 0.75 in spacing.

sepehr2github (commented):
Thank you Rakash. Looks nice 👍

packages/ui-components/src/components/input/valueInput.tsx
RakeshUP: Apparently, TW's focus-within is not working when used along with the hover variant. When both are specified, the hover is taking precedence over focus-within. 

I used focus-within in plain CSS and it seems to be working as expected!

packages/ui-components/src/components/input/valueInput.tsx
RakeshUP: The design seems to use both focus and active variants for input, and I guess active should apply when value is being entered. I'll confirm with Selim and update the border styles for all inputs and also extract to a single place then

packages/ui-components/tailwind.config.js
```diff
@@ -67,6 +67,7 @@ module.exports = {
       // extends spacing accorging to design system
       spacing: {
         0.5: '4px',
+        0.75: '6px',
```
Fabricevladimir, line 70: If you get a chance, mirror that change over to the web-app as well!

Fabricevladimir (approved):
Ready to roll!

sepehr2github (approved):
Good to go 👍

PR #105, opened 2021-12-23
feature/tokenListItem DAO-538 

sepehr2github:
- Token List item added

### DIFF
--- a/packages/ui-components/src/components/tokenListItem/tokenListItem.tsx
+++ b/packages/ui-components/src/components/tokenListItem/tokenListItem.tsx
@@ -19,6 +19,10 @@ export type TokenListItemProps = {
    * src of token logo
    */
   tokenLogo: string;
+  /**
+   * Whether list item is disabled
+   */
+  disabled?: boolean;
   onClick?: () => void;
 };
 
@@ -27,11 +31,12 @@ export const TokenListItem: React.FC<TokenListItemProps> = ({
   tokenSymbol,
   tokenAmount,
   tokenLogo,
+  disabled,
   onClick,
 }) => {
   return (
     <Container
-      {...{onClick}}
+      {...{onClick,disabled}}
        data-testid="tokenListItem"
     >
       <TextWrapper>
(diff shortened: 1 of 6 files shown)


RakeshUP:
There's a component already tokenCard that looks similar. Do you think the tokenCard could be modified to make it compatible with the requirements of this PR DAO-538? @sepehr2github

sepehr2github:
> 

Thank you Rakash For this suggestion. but TokentemList is a button and has a different structure than tokenCard which is a div element but We can change it later if needed

packages/ui-components/src/components/tokenListItem/tokenListItem.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import styled from 'styled-components';
+import {Avatar} from '../avatar';
+
+export type TokenListItemProps = {
+  /**
```
Fabricevladimir, line 6: Don't forget the `bgWhite` prop!

packages/ui-components/src/components/tokenListItem/tokenListItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const Container = styled.button.attrs(({disabled}: StyledContentProps)=> ({
+  className:`w-full flex justify-between items-center py-1.5
+  px-2 hover:text-ui-800 hover:bg-ui-100 active:text-ui-800 
+  active:bg-ui-200 disabled:text-ui-300 rounded-xl
+  ${disabled ? 'bg-ui-100' : 'bg-white'}`,
+}))``;
+
+const AmountWrapper = styled.h3.attrs({
+  className:'font-semibold text-ui-600 text-base',
```
Fabricevladimir: I think the font weight has changed to 500 so normal.

Fabricevladimir (approved):
Nice, nice!

RakeshUP:
> > 
> 
> Thank you Rakash For this suggestion. but TokentemList is a button and has a different structure than tokenCard which is a div element but We can change it later if needed

Sorry I completely missed that fact that this component is a button!

RakeshUP (approved):

PR #107, opened 2021-12-23
Feature/Edit tokenCard - DAO-540

Fabricevladimir:
- Modified `tokenCard` to take `bgWhite` prop as well as type of `vault` or `transfer`
- Renamed card components to be more consistent with design system
- Restructured storybook to place all card components in one folder
- Updated web-app to match component name change
- Slight modification to token sort to keep undefined value at the end of list
- Added more MANA because I can


packages/ui-components/src/components/card/cardToken.tsx
```diff
@@ -34,54 +41,77 @@ export const TokenCard: React.FC<TokenCardProps> = ({
...
-            </ToggleMobileVisibility>
-          </SecondaryCoinDetails>
+          {isVault && (
+            <SecondaryCoinDetails>
+              <span>
+                {props.tokenCount} {props.tokenSymbol}
+              </span>
+              <ToggleMobileVisibility visible={false}>
+                <span>•</span>
+                <span> {props.tokenUSDValue}</span>
```
WissenIstNacht: If `tokenUsdValue` is not there for some reason (like, say, you forget to pass it to the corresponding story :D), this will only show the dot from line 55.

WissenIstNacht (changes requested):
Thanks Fabrice!

LIttle detail to handle on the token card :)

sepehr2github (approved):
Sounds Great! Thank you.

WissenIstNacht (approved):
Good to go!

PR #108, opened 2022-01-03
Feature/dropdown input feature/ DAO-545 

sepehr2github:
- dropdown input added to ui compoenets
- clickable prop added to base input (I used Text Input as a base component for all the other types to keep consistency)

Note: Also clickable added to the input component for changing cursor on disabled mode


packages/ui-components/src/components/input/index.ts
```diff
@@ -1,2 +1,3 @@
 export {TextInputProps, TextInput} from './textInput';
-export {SearchInput, SearchInputProps} from './SearchInput';
\ No newline at end of file
+export {SearchInput, SearchInputProps} from './searchInput';
```
RakeshUP: The file is not renamed in the PR. It's causing a build error

RakeshUP (approved):
Looks good. I've left a minor comment

packages/ui-components/src/components/input/index.ts
sepehr2github: Ah, That's wired! I renamed the file

Fabricevladimir (approved):

PR #114, opened 2022-01-05
Feature/Review page DAO-542 

Fabricevladimir:
- wired deposit form (minus validation)
- added review deposit
- implemented simple price fetch function
- modified getTokenInfo to return null when information cannot be fetched
- changed card ui-component stories to show up in the `cards` folder instead of the previous `card`

This PR also takes care of the open and close token modal logic (Desktop only)

UPDATE:
- wired withdraw form (minus validation)
- refactored token menu


packages/web-app/src/containers/reviewDeposit/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }, [
+    provider,
+    setValue,
+    values.amount,
+    values.isCustomToken,
+    values.tokenAddress,
+  ]);
+
+  return (
+    <div className="flex flex-col space-y-1.5 desktop:space-y-3">
```
WissenIstNacht: (technically, I think the flex is not necessary)

packages/web-app/src/containers/tokenMenu/index.tsx
```diff
@@ -1,50 +1,102 @@
...
+  const curatedTokenAddresses = useMemo(
+    () => Object.entries(curatedTokens[chainId || 4].curatedTokens),
+    [chainId]
+  );
+
+  const tokenBalances = useMemo(
+    () =>
+      curatedTokenAddresses.map(value => ({
+        address: value[1],
+        count: BigInt(0),
```
WissenIstNacht: So should that be the balance?

packages/web-app/src/containers/reviewDeposit/index.tsx
Fabricevladimir: True, true.

WissenIstNacht:
Hey Fabrice, thanks a lot for the work.

Unfortunately, I can't open the token menu. Therefore, there's a lot of features I can't test. Ping me when you can, I think it will be easier to discuss this on a call :)

packages/web-app/src/containers/tokenMenu/index.tsx
Fabricevladimir: Yes, the `WalletToken` type held a `balance` property that is the human readable balance. `count` here is the `BigInt` balance. 
That said, I have decided to do away with the `WalletToken` type and use the `count` property and format it whenever it is being displayed instead.

Fabricevladimir:
Converted PR back to draft due to issues with the dropdownInput not propagating click event properly. The status will be reverted once the latest dropdownInput changes have been merge (#115)

RakeshUP:
Hey @Fabricevladimir, there was a styling issue with the Progress bar(the bar should be rounded if val = max). 
I had missed this case when I first developed the component. I went ahead and put a fix for that. See if it looks good for you.

Link to commit https://github.com/aragon/zaragoza/pull/114/commits/02db1d053d64faa63d05d11e11745f4ebb99b5b1

sepehr2github:
Is there anything needed from my side to updraft this pr?

Fabricevladimir:
@RakeshUP  Thanks for the quick fix!

packages/ui-components/src/components/input/dropdownInput.tsx
```diff
@@ -8,6 +8,8 @@ export type DropDownInputProps = {
   mode?: 'default' | 'success' | 'warning' | 'critical';
   disabled?: boolean;
   value?: string;
+  name?: string;
```
WissenIstNacht, line 11: Was this strictly necessary to change for the review page? I'm asking because:

- That's not the right way to do it. placeholder and name should both be inherited by extending the HtmlInputElement type
- I've had to do this to get one of my PRs to work, which will likely mean merge issues.

packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -3,22 +3,22 @@ import styled from 'styled-components';
...
+export type ValueInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
   /* Text that appears on the button present on the right side of the input  */
   adornmentText: string;
 
   onAdornmentClick: () => void;
 
   /** Changes a input's color schema */
   mode?: 'default' | 'success' | 'warning' | 'critical';
-  disabled: boolean;
+  disabled?: boolean;
```
WissenIstNacht: this is no longer necesssary once you extend the type as you do now.

packages/web-app/src/containers/configureWithdraw/index.tsx
```diff
@@ -1,21 +1,35 @@
...
-import styled from 'styled-components';
-import {useTranslation} from 'react-i18next';
-import {Control, Controller} from 'react-hook-form';
 import {
   DropdownInput,
   Label,
   TextInput,
   ValueInput,
 } from '@aragon/ui-components';
+import styled from 'styled-components';
```
WissenIstNacht, line 7: Same here, seems like this has nothing to do with the deposit flow. Everyone working on the withdraw flow who actually has to change this will face merge issues.

packages/ui-components/src/components/input/dropdownInput.tsx
Fabricevladimir: The name? Sure, it's part of properly wiring up the form. The dropdownInput has to be changed at some point I agree, but I can't let this block the PR

packages/ui-components/src/components/input/valueInput.tsx
Fabricevladimir: This is explicitly being pulled out I assume because both the input and the container need that disabled value. That said, it could be gotten off of `...props`. This foray into the ui-components is because the input components weren't accepting the props needed to wire the form.

packages/web-app/src/containers/configureWithdraw/index.tsx
Fabricevladimir: Perhaps there will be issues, yes, but wiring the form is a prerequisite to validating it. This `configureWithdraw` already has all the necessary components.

packages/ui-components/src/components/input/dropdownInput.tsx
WissenIstNacht: Both `name` and `placeholder`, as well as `value` and `disabled` are part of `React.InputHTMLAttributes<HTMLInputElement>`.

packages/ui-components/src/components/input/valueInput.tsx
WissenIstNacht: But that's not bein explicit, it's doubling a type key :). What you have now is basically:

```
type A = {
  // all the stuff from React.InputHTMLAttributes<HTMLInputElement> and
  disabled?: boolean
} & {
  // the "explicit" stuff that you extend
  adornmentText: string;
  onAdornmentClick: () => void;
  mode?: 'default' | 'success' | 'warning' | 'critical';
  disabled?: boolean
}
```

If you explicitly need the type key you need to take it out of props when spreading the args in the component (Line 17 onwards)

packages/web-app/src/containers/configureWithdraw/index.tsx
WissenIstNacht: Sure, but isn't there a separate subtask for this?

//Edit: 585, you even assigned it to yourself 😅

packages/web-app/src/containers/depositForm/index.tsx
```diff
@@ -1,17 +1,43 @@
...
+  // TODO: This should probably come with the input ui-component
+  const handleClipboardActions = useCallback(
+    async (currentValue: string, onChange: (value: string) => void) => {
+      if (currentValue) {
+        await navigator.clipboard.writeText(currentValue);
+
+        // TODO: change to proper mechanism
+        alert('Copied');
+      } else {
+        const textFromClipboard = await navigator.clipboard.readText();
```
WissenIstNacht, line 35: This doesn't work on Firefox... Might be because it returns a promise.

packages/web-app/src/containers/depositForm/index.tsx
```diff
@@ -44,24 +111,47 @@ const DepositForm: React.FC<DepositFormProps> = ({control}) => {
...
+            fieldState: {error},
+          }) => (
+            <ValueInput
+              mode={error ? 'critical' : 'default'}
               name={name}
               value={value}
               onBlur={onBlur}
               onChange={onChange}
+              adornmentText="Max"
+              onAdornmentClick={() => onChange(tokenBalance)}
```
WissenIstNacht, line 125: Max button doesn't work. Tried with several tokens both on mainnet/rinkeby, but only on firefox.

packages/ui-components/src/components/input/dropdownInput.tsx
Fabricevladimir: The issue with extending `HtmlInputElement` at this point is that the `dropdownInput` itself is a `button`, not a usual `input`. There will be props (`min`, `maxLength`, `pattern`, `autocomplete` and so on) that are allowed onto it that shouldn't be at least at this moment.

packages/web-app/src/containers/configureWithdraw/index.tsx
Fabricevladimir: There isn't. It's the same thing that happened with the deposit. It needed to be shoehorned into one of the tasks. It was either this or validation (585). I only added it here because that's where I did the same for the deposit.

packages/web-app/src/containers/depositForm/index.tsx
Fabricevladimir: Good catch!

packages/web-app/src/containers/depositForm/index.tsx
Fabricevladimir: Huh, that's interesting. Works in Chrome. I'll investigate

WissenIstNacht:
Hey man, thanks for the PR.

I really appreciate your commitment to improving the code and by extension what we're building. But it's getting a bit out of hand. Recall, this ticket's goal was ot assmeble a few components into a page. This would mean changes to the new-deposit page, maybe a tweak or two to some utility methods and maybe a change to the deposit form. Yet, this PR has 31 files changed and a diff of a 1000 lines...

packages/ui-components/src/components/input/valueInput.tsx
Fabricevladimir: I've updated this. The explanation though is a bit misleading I believe. Because the same key exists with the **same type**, there is no actual doubling or type intersection for that key. There is only one `disabled` key in the resulting type. You are correct though that it is not necessary to have it in the "explicit" stuff that we extend.

packages/web-app/src/containers/depositForm/index.tsx
Fabricevladimir: Apparently this is a Firefox specific thing. It comes disabled by default. To enable it, toggle the `dom.events.testing.asyncClipboard` to true on Firefox's `about:config` page.

WissenIstNacht: Nah it's enabled. Wouldn't work for other sites either, otherwise :)

WissenIstNacht: So I fired up the dev environment again and now it works... 🤷🏼‍♂️

packages/web-app/src/containers/depositForm/index.tsx
WissenIstNacht: When I select a token, should the logged object at the bottom of the page have a `tokenBalance`? Cause currently it doesn't. Would explain why the max button doesn't change anything.

Fabricevladimir: It should for the curated tokens and it should be cleared if custom token is selected...

RakeshUP:
@Fabricevladimir @WissenIstNacht shall we just merge this? I checked this PR sometime back and also made some changes to the styling of LinearProgress at that time. I didn't review it after it was marked ready. 

Why don't we review and change things(if necessary) when we work on this page again(like subgraph integration, tx integration using our SDK)?

packages/web-app/src/containers/depositForm/index.tsx
WissenIstNacht: That might explain it, then. I haven't got any of the curated tokens :D

WissenIstNacht (approved):

WissenIstNacht:
From my side the PR is actually in a good state. There were a few things that needed checking, but I think it's ready to be merged now 👍🏼

RakeshUP (approved):

PR #115, opened 2022-01-05
Feature/numeric input DAO-556 

sepehr2github:
- Numeric input added
- Dropdown Input refactored and changed to a button because based on Figma files all dropdown inputs should open a modal so the functionality would be similar to a button, not an Input!


packages/ui-components/src/components/input/valueInput.tsx
```diff
@@ -3,14 +3,15 @@ import styled from 'styled-components';
 import {ButtonText} from '../button';
 import {StyledInput} from './textInput';
 
-export type ValueInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
+export type ValueInputProps = {
```
Fabricevladimir, line 6: Why remove the type?

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -0,0 +1,73 @@
+import React, {useRef} from 'react';
+import styled from 'styled-components';
+import {ButtonIcon} from '../button';
+import {IconAdd, IconRemove} from '../icons';
+
+export type NumberInputProps = {
```
Fabricevladimir: Is there a TS type that can be used to provide the usual intellisense for the normal input number? Like something including `max` and `min`

Fabricevladimir (commented):
Good work

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+  /** Changes a input's color schema */
+  mode?: 'default' | 'success' | 'warning' | 'critical';
+  disabled?: boolean;
+};
+
+export const NumberInput: React.FC<NumberInputProps> = ({
+  mode = 'default',
+  disabled,
+  ...props
```
WissenIstNacht, line 17: What other props are there besides mode and disabled?

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+}) => {
```
WissenIstNacht, line 18: How do you get the value out of here without value and onChange as props?

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  return (
+    <Container data-testid="number-input" {...{mode, disabled}}>
```
WissenIstNacht: The container does not scale in width. Entering a 6 digit number will not show it properly :p

RakeshUP (approved):
The Plus & Minus Icons are reversed
The input text color needs to be changed for disabled selector

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+        onClick={() => inputRef.current?.stepDown()}
+      />
+    </Container>
+  );
+};
+
+export type StyledContainerProps = Pick<NumberInputProps, 'mode' | 'disabled'>;
+
+const Container = styled.div.attrs(({mode, disabled}: StyledContainerProps) => {
+  let className = `${disabled ? 'bg-ui-100' : 'bg-ui-0'} flex p-1 bg-ui-0 w-18
```
WissenIstNacht: This relates to the above comment. There should not be a fixed width on here.

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const NumberInput: React.FC<NumberInputProps> = ({
+  mode = 'default',
+  disabled,
+  ...props
+}) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  return (
+    <Container data-testid="number-input" {...{mode, disabled}}>
+      <ButtonIcon
```
WissenIstNacht, line 23: Wrong way round :) should be minus on the left and plus on the right

WissenIstNacht (changes requested):
Hey Sepehr, thanks! :)

Thanks for trying to clean this up. However, I'm not entirely convinced that the dropdown should be a button only. Please consult with Selim before doing these changes. And do them in a separate PR. You're lucky this time because apparently the dropdown is not working properly and Fabrice needs these changes.

As for the number input, there's a couple of things that are wrong. 

Also, as Fabrice pointed out, please leave the types as they were.

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: Already Fixed

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: The value & onChange exist inside the (...props)

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: image
I think It shouldn't scale, even in normal numeric input there is a limit and if you passed the limit overflow will be hidden. So I think the question is how much the limit should be?

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: https://github.com/aragon/zaragoza/pull/115#discussion_r779577320

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: onChange,Value,min,max,...

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: I couldn't find any TS type only for number input. So I believe InputHTMLAttributes will support all the props

sepehr2github:
> Hey Sepehr, thanks! :)
> 
> Thanks for trying to clean this up. However, I'm not entirely convinced that the dropdown should be a button only. Please consult with Selim before doing these changes. And do them in a separate PR. You're lucky this time because apparently the dropdown is not working properly and Fabrice needs these changes.
> 
> As for the number input, there's a couple of things that are wrong.
> 
> Also, as Fabrice pointed out, please leave the types as they were.

Thank you, Vali for the feedbacks. We had a short discussion with Fabrice and we decided to change the dropdown and solve all the problems. But you're right Sorry I should've done all changes on a separate pull request but I thought this way is faster.

packages/ui-components/src/components/input/numberInput.tsx
WissenIstNacht: You're absolutely right that the input width does not scale with the content entered by the user. My bad on this, sorry :)

BUT: The input field and the container should still be of variable width, I think. The client of the component should say how much space the numeric input component should take, which in turn should dictate how much space the input can take. I.e., it should take up all available width. Does that make sense?

packages/ui-components/src/components/input/numberInput.tsx
WissenIstNacht: Ok, but then it doesn't coincide with the prop type. Why did you not add them? And why did you remove them in other places?

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: Yes and I added custom width

packages/ui-components/src/components/input/numberInput.tsx
sepehr2github: I had to because of a type error but I handle  it on a different way and All types added again

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const className: string | undefined = `${
+    disabled ? 'text-ui-300' : 'text-ui-600'
+  } bg-transparent focus:outline-none margin-0 w-full`;
+
+  return {
+    className,
+  };
+})`
+  text-align: center;
+  ::-webkit-inner-spin-button,
```
WissenIstNacht, line 91: nice :)

WissenIstNacht (approved):
Thanks a lot for the changes. Good to go! :)

PR #121, opened 2022-01-10
feature/textarea->WYSIWYG DAO-581

RakeshUP:
The set of icons from the markdown folder was added in this PR. That's why the big change!

The main essence is in the file packages/ui-components/src/components/textarea/wysiwyg.tsx


packages/ui-components/src/components/textarea/wysiwyg.tsx
```diff
@@ -0,0 +1,182 @@
...
+      Placeholder.configure({
+        placeholder,
+      }),
+    ],
+  });
+
+  return (
+    <Container>
+      <MenuBar editor={editor} />
+      <StyledEditorContent editor={editor} />
```
sepehr2github, line 121: Can you add a disabled prop?

sepehr2github (commented):
Thank you Rakash, You did a great job!!! Well Done.
also, can you add yarn.lock file

packages/ui-components/src/components/icons/markdown/icon_bold.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import {IconType} from '..';
+
+export const IconBold: IconType = ({height = 16, width = 16, ...props}) => {
+  return (
+    <svg
+      width={width}
+      height={height}
+      fill="none"
+      viewBox="0 0 12 16"
```
WissenIstNacht: This is debatable, but I think having square viewboxes for all icons is simpler to get them to align properly, etc.

You don't need to edit this manually, you can do this from figma directly. Instead of exporting the icon itself you can export the parent, which will always give you the svg wiht a viewbox of 16x16

WissenIstNacht (commented):
Awesome, thanks!

I left some comments on the icons, let me know what you think.

Also, almost all tests fail. Does that happen to you too?

As for Sepehr's comment, I agree with the disabled. But the lock file for yarn is there for me, I don't think there needs to be a second :yum:

packages/ui-components/src/components/textarea/wysiwyg.tsx
RakeshUP: Added

packages/ui-components/src/components/icons/markdown/icon_bold.tsx
RakeshUP: I've changed 2 icons that had a different viewBox. Now all the newly added ones are 16x16

sepehr2github:
> Awesome, thanks!
> 
> I left some comments on the icons, let me know what you think.
> 
> Also, almost all tests fail. Does that happen to you too?
> 
> As for Sepehr's comment, I agree with the disabled. But the lock file for yarn is there for me, I don't think there needs to be a second 😋

Yes, You're right. sorry @RakeshUP It wasn't load for me last time

sepehr2github (approved):
Thanks for Resolving all comments

WissenIstNacht (approved):
Thanks a lot for the changes to the view box, much appreciated!

PR #122, opened 2022-01-10
Feature: Dao-558 - time zone drop down menu

WissenIstNacht:
Contains:
- UTC modal
  - With all UTCs displayed
  - Search input to filter them 
- Slight modifications to the context that manages modals
- (temporary) add UTC dropdown that can open the utc modal
- Modification to some of the input types because they lead to faulty build


packages/web-app/src/app.tsx
```diff
@@ -59,6 +60,7 @@ function App() {
       <WalletMenu />
       <TransferMenu />
       <TokenMenu />
+      <UtcMenu />
```
Fabricevladimir, line 64: Can stay here for now, but this will need to move to the page it exists on in order to get the selection from it.

packages/web-app/src/containers/utcMenu/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const handleUtcClick = () => {
+    close('utc');
+  };
+
+  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setSearchTerm(event.target.value);
+  };
+
+  const filteredTimezones = timezones.filter(tz => tz.includes(searchTerm));
```
Fabricevladimir: Can you ignore casing for this? Currently searching for utc won't return anything.

Update - A nice to have would be autofocusing the search when the menu opens. If it takes more than a minute though, don't worry about it 😄

packages/web-app/src/containers/utcMenu/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    >
+      <ModalBody>
+        <SearchInput
+          placeholder="Type to filter"
+          value={searchTerm}
+          onChange={handleChange}
+        />
+        <Container>
+          <ScrollableDiv>
+            {filteredTimezones.map((tz: string) => {
```
Fabricevladimir, line 48: I do realize there are no designs for this, but can you add some quick spacing to this? The cards appear cut off when clicked on.

packages/web-app/src/context/transfersModal.tsx
```diff
@@ -13,34 +13,50 @@ const TransferModalContext = createContext<TransferModalContextType | null>(
...
+    useState<TransferModalContextType['isUtcOpen']>(false);
 
-  const open = (type?: string) => {
+  const open = (type?: MenuTypes) => {
     switch (type) {
       case 'token':
         setIsTokenOpen(true);
         break;
+      case 'utc':
+        console.log('Hihodfs');
```
Fabricevladimir: Forgot to remove this log!

Fabricevladimir (commented):
Good work!

packages/web-app/src/app.tsx
WissenIstNacht: Makes sense, hadn't thought about that. I'll add a TODO for whoever does the final assembly/wiring up/validation.

packages/web-app/src/containers/utcMenu/index.tsx
WissenIstNacht: Yes. The autofocus would be a nice touch. But given that this was freestyled and might be replaced by proper designs later I wont spend time improving it rn.

RakeshUP (approved):
Looks good. Could you just make the searching case ignorant? Other changes that Fabrice requested could be done later

Fabricevladimir (approved):
Merge merge merge!

PR #123, opened 2022-01-11
Feature/execution terminal DAO-589 

sepehr2github:
- execution terminal added to ui componenets 
- change "tokenUSDValue" prop type to optional in cardToken


packages/ui-components/src/components/card/cardExecution.tsx
```diff
@@ -0,0 +1,101 @@
...
+   */
+  title: string;
+  /**
+   * Description text
+   */
+  description: string;
+  /**
+   * Allows the Execution Card component grow horizontally
+   * */
+  wide?: boolean;
```
Fabricevladimir: Can this be switched to different styling on tablet instead since it seems that it's on tablet up that the terminal grows to fill the container?

packages/ui-components/src/components/card/cardExecution.tsx
```diff
@@ -0,0 +1,101 @@
...
+            tokenSymbol,
+            tokenCount,
+            treasuryShare,
+          }}
+          type={'transfer'}
+          bgWhite
+        />
+      </Content>
+      <Action>
+        <ButtonText label={'Execute Now'} />
```
Fabricevladimir: The button seems to take full width of its container on mobile and hug contents on tablet up. Also, you might have forgotten the `onClick` prop!

packages/ui-components/src/components/card/cardExecution.tsx
```diff
@@ -0,0 +1,101 @@
...
+    | 'type'
+    | 'bgWhite'
+    | 'changeType'
+    | 'tokenUSDValue'
+    | 'changeDuringInterval'
+    | 'treasurySharePercentage'
+    | 'percentageChangeDuringInterval'
+  >;
+
+export const CardExecution: React.FC<CardExecutionProps> = ({
```
Fabricevladimir: It seems we also need to add some sort of status prop that tells the terminal the status of the execution so the button can be enabled and disabled accordingly.

Fabricevladimir (commented):
Thanks! Good work

packages/ui-components/src/components/card/cardExecution.tsx
RakeshUP: Could you add text truncate ellipsis to longer texts(or maybe everything)? The layout is inconsistent in small mobile view

packages/ui-components/src/components/card/cardExecution.tsx
RakeshUP: loading state too I guess

RakeshUP (approved):
Looks good. Just a few changes needed before merging

packages/ui-components/src/components/card/cardExecution.tsx
sepehr2github: @RakeshUP I couldn't get what you exactly mean by that. can you give me an example?

packages/ui-components/src/components/card/cardExecution.tsx
sepehr2github: > loading state too I guess

Can you explain more?

RakeshUP: In this [figma](https://www.figma.com/file/t44Siq3leL4AnKxbc10FVp/Components?node-id=722%3A9591), the execute now button has 3 states(disabled, enabled, done) and I was guessing there might be a 4th state too(loading). But we could do that when we are doing the blockchain integration :)

packages/ui-components/src/components/card/cardExecution.tsx
RakeshUP: Screenshot 2022-01-12 at 10 48 04 AM

The component itself is wider than the viewport because the tokenCount string is very wide. We need to truncate it somehow without losing data(like showing 15,000,230.2323 DAI as 15M DAI). Not a pressing issue right now!

Fabricevladimir: There are a few changes for the mobile version (button size, card size and text truncating) as well as the different states. We can either do this now or revisit it when we do the actual integration. If we leave it for later, either add this in the notion document for issues & refactorings or add a TODO in the code.

Fabricevladimir (approved):

PR #124, opened 2022-01-11
chore/refactor button group

RakeshUP:
Most of the PR is file name changes!


packages/ui-components/src/components/button/buttonGroup/button.tsx
```diff
@@ -0,0 +1,47 @@
+import React, {ButtonHTMLAttributes} from 'react';
```
Fabricevladimir: What's the reason for creating a new button and not using the ghost variant of the existing `buttonText`?

packages/ui-components/src/components/button/buttonGroup/button.tsx
```diff
@@ -0,0 +1,47 @@
...
+      ${
+        isSelected
+          ? bgWhite
+            ? 'text-primary-500 bg-ui-0'
+            : 'text-primary-500 bg-ui-50'
+          : bgWhite
+          ? 'bg-ui-50 text-ui-600'
+          : 'bg-ui-0 text-ui-600'
+      }`,
+    style: {borderRadius: '0.625rem'},
```
Fabricevladimir: Yeah, I have a note for this, for whatever reason TW refuses to take a new border radius key. This is why all the default sized buttons still have the wrong border-radius

packages/ui-components/src/components/button/buttonGroup/buttonGroup.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onChange?: (value: string) => void;
+  fullWidth?: boolean;
+};
+
+export const ButtonGroup: React.FC<ButtonGroupProps> = ({
+  bgWhite,
+  defaultValue,
+  onChange,
+  fullWidth = false,
+  children,
```
Fabricevladimir, line 66: Is it worth limiting the children type to something like `React.FunctionComponentElement<Option>`?

packages/ui-components/__tests__/buttonGroup.test.tsx
```diff
@@ -1,18 +1,18 @@
 import React from 'react';
 import {render, screen} from '@testing-library/react';
 
-import {RadioGroup, Radio} from '../src';
+import {Option, ButtonGroup} from '../src';
 
 describe('Radio Group', () => {
```
Fabricevladimir: I think you missed the name change to Button Group instead of Radio Group :)

packages/ui-components/src/components/button/buttonGroup/buttonGroup.tsx
```diff
@@ -0,0 +1,88 @@
...
+      {children}
+    </ButtonGroupContext.Provider>
+  );
+};
+
+export type ButtonGroupProps = {
+  bgWhite: boolean;
+  defaultValue: string;
+  onChange?: (value: string) => void;
+  fullWidth?: boolean;
```
Fabricevladimir, line 57: On the components design, it seems as if this is supposed to only hug the contents.

Fabricevladimir (commented):
Thanks! Nice work. Left a few comments

packages/web-app/src/pages/transfers.tsx
```diff
@@ -51,12 +51,16 @@ const Transfers: React.FC = () => {
       >
         <div className="space-y-1.5">
           <SearchInput placeholder="Type to filter" />
-          <RadioGroup defaultValue="All" onChange={handleButtonGroupChange}>
-            <Radio value="All">All</Radio>
-            <Radio value="Deposit">Deposit</Radio>
-            <Radio value="Withdraw">Withdraw</Radio>
-            <Radio value="External Contract">External Contract</Radio>
-          </RadioGroup>
+          <ButtonGroup
```
sepehr2github, line 54: Can you add secondary color?
image

sepehr2github (commented):
Thank you Rakash. Now I can use this button group inside inputs

packages/ui-components/src/components/button/buttonGroup/button.tsx
RakeshUP: Hmmm, as long as you give the new keys inside the extend:{} object in TW config, it should work. I put this here because I thought 0.625rem is a one-off value and doesn't seem to be used anywhere else

packages/ui-components/src/components/button/buttonGroup/button.tsx
RakeshUP: I needed flex-1 in button. When the parent component is fullWidth, the button width should expand and all buttons should be of equal sizes. Also, there are small differences like the button in buttonGroup has primary-500 text color and the ghost of buttonText is some gray

packages/ui-components/src/components/button/buttonGroup/buttonGroup.tsx
RakeshUP: I too am not happy with the fullWidth prop. But the buttonGroup seems to be used in all possible ways. In the first image, it's in a new line, but not fullWidth. In the second image, the one in mobile view is fullWidth and the "md" variant is not max-content

Screenshot 2022-01-12 at 3 39 17 PM
Screenshot 2022-01-12 at 3 39 33 PM

packages/web-app/src/pages/transfers.tsx
RakeshUP: The buttonGroup varies a lot from the one needed in inputDate component. The buttonGroup in this PR has a space-between 12px between the buttons. I think a completely new one is needed for inputDate. Might not need to be a component though

packages/ui-components/src/components/button/buttonGroup/button.tsx
Fabricevladimir: I think the text color for ghost buttonText is primary-500 when `isActive` is true and ui-600 (the default gray for the unselected buttons in the group) when not. As for the flex-1, it is possible to pass it to the ghost buttonText as a className. I only stress this because in the component designs, I saw that the buttonText ghost is the one used in the buttonGroups.

RakeshUP: Screenshot 2022-01-17 at 5 30 15 PM

I tried using ghost buttonText and I got stuck here. For the text to be centered, flex-1 has to be added to both ButtonText and ButtonBase and I am not sure if those changes won't break anything!

RakeshUP: Adding justify-center worked!

packages/ui-components/src/components/button/buttonText.tsx
```diff
@@ -72,6 +82,6 @@ const StyledButton = styled(ButtonBase).attrs(
         } text-ui-0 hover:bg-primary-500 active:bg-primary-700 disabled:text-primary-300 disabled:bg-primary-100`;
     }
 
-    return {className};
+    return {className: `${className} ${inputClassName}`};
```
Fabricevladimir, line 85: Is this so certain things are overridden?

Fabricevladimir (approved):
Thanks. Left a question.

sepehr2github (approved):
Well Done! Thank you for resolving all comments. Good to go!

packages/ui-components/src/components/button/buttonText.tsx
RakeshUP: classNames are not overridden left-to-right i.e. the rightmost class is not given more preference than the one in the left. The classname appearing last in the stylesheet file has the most precedence.

And as for me, I just put it in that order randomly

PR #125, opened 2022-01-11
Feature/Link - DAO-608 

Fabricevladimir:
- added `Link` component 
- removed stray log in test


packages/ui-components/src/components/link/link.tsx
```diff
@@ -0,0 +1,62 @@
...
+  href,
+  ...props
+}) => {
+  return (
+    <StyledLink
+      href={disabled ? undefined : href}
+      rel="noopener noreferrer"
+      active={active}
+      disabled={disabled}
+      {...(external ? {target: '_blank'} : {})}
```
sepehr2github, line 35: Thank you for adding this option.

sepehr2github (approved):
Looks Great! Awsome. Fabrice

packages/ui-components/src/components/link/link.tsx
```diff
@@ -0,0 +1,62 @@
...
+        {!iconLeft && iconRight && iconRight}
+      </div>
+    </StyledLink>
+  );
+};
+
+type StyledLinkProps = {disabled: boolean; active: boolean};
+const StyledLink = styled.a.attrs(({active, disabled}: StyledLinkProps) => {
+  let className = `overflow-hidden text-primary-500 hover:text-primary-700
+      focus:bg-ui-0 rounded focus:ring-2 focus:ring-primary-500 focus:outline-none`;
```
WissenIstNacht: So apparently, the focus state is not meant to change the background. I t should always be transparent. The reason this was done on figma is because apparently you need to set a background to show the border or something. Not your fault, but it needs changing :)

packages/ui-components/stories/link.stories.tsx
```diff
@@ -0,0 +1,32 @@
...
+  iconRight: <IconChevronDown />,
+  label: 'Link text',
+  href: 'https://abc.com',
+};
+
+export const IconLeft = Template.bind({});
+IconLeft.args = {
+  iconLeft: <IconChevronDown />,
+  label: 'Link text',
+  href: 'https://abc.com',
```
WissenIstNacht, line 29: Exotic choice :D

packages/ui-components/src/components/input/dateInput.tsx
```diff
@@ -9,7 +9,7 @@ export type DateInputProps = React.InputHTMLAttributes<HTMLInputElement>;
 
 export const DateInput: React.FC<DateInputProps> = ({disabled, ...props}) => {
   const isFF = navigator.userAgent.indexOf('Firefox') !== -1;
-  console.log(isFF);
+
```
WissenIstNacht, line 12: Thanks!

WissenIstNacht (changes requested):
Thanks, looks great!

There's a tiny change needed, then we should be good to go.

packages/ui-components/stories/link.stories.tsx
Fabricevladimir: What can I say? I'm wild!

Fabricevladimir: What can I say? I'm wild!

WissenIstNacht (approved):

PR #127, opened 2022-01-12
Adds the adaptive ERC165 to the contracts

nivida:


nivida:
Got approved after a quick chat with @novaknole. Merging it now.

PR #128, opened 2022-01-13
Feature: DAO 613 - Progress status

WissenIstNacht:
This PR implements:
- Progress->Status component
- Corresponding story
- Corresponding test


packages/ui-components/src/components/icons/interface/icon_block.tsx
```diff
@@ -0,0 +1,23 @@
...
+      width={width}
+      height={height}
+      fill="none"
+      viewBox="0 0 16 16"
+      {...props}
+    >
+      <g clip-path="url(#clip0_433_16)">
+        <path
+          fill-rule="evenodd"
+          clip-rule="evenodd"
```
Fabricevladimir: Can you camel case `fill-rule` and `clip-rule`?

packages/ui-components/src/components/icons/interface/radio/cancle.tsx
```diff
@@ -0,0 +1,25 @@
...
+}) => {
+  return (
+    <svg
+      width={width}
+      height={height}
+      viewBox="0 0 16 16"
+      fill="none"
+      {...props}
+    >
+      <g clip-path="url(#clip0_430_5)">
```
Fabricevladimir: Camel casing for `clip-path`

packages/ui-components/src/components/icons/interface/radio/cancle.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import {IconType} from '../..';
+
+export const IconRadioCancle: IconType = ({
```
Fabricevladimir: Slight misspelling, should be 'cancel'.

packages/ui-components/src/components/progress/status.tsx
```diff
@@ -0,0 +1,151 @@
...
+    <TopContainer data-testid="progressStatus">
+      <LeftContainer mode={mode}>
+        <IconContainer>
+          <Icon mode={mode} />
+        </IconContainer>
+        <LabelContainer>
+          <CustomLabel label={label} helpText={date} />
+        </LabelContainer>
+      </LeftContainer>
+      <BlockContainer>
```
Fabricevladimir: If no block is given, the icon shouldn't be rendered.

packages/ui-components/src/components/progress/status.tsx
```diff
@@ -0,0 +1,151 @@
...
+    case 'failed':
+      return <IconRadioCancle className={iconColors[mode]} />;
+    default:
+      return <IconSuccess className={iconColors[mode]} />;
+  }
+};
+
+type CustomLabelProps = Pick<LabelProps, 'label' | 'helpText'>;
+
+/* Lord knows it hurts. */
```
Fabricevladimir, line 129: Yes, this is just sad, really.

packages/ui-components/src/components/progress/status.tsx
```diff
@@ -0,0 +1,151 @@
...
+   * it is in the future).
+   *
+   * If no date is passed when one is required, a fallback text will be displayed.
+   * */
+  date?: string;
+  /**
+   * If the progress status changed due to an event on a blockchain, the
+   * corresponding block MAY be passed. Note that upcoming, active and rejected
+   * mode can NEVER have a block associated.
+   */
```
Fabricevladimir, line 44: I appreciate this comment. Is this something that can or even should be implemented in the code itself? Like if the mode is one of that shouldn't show the block, hide the block even if the property is provided? While it does limit the user, it also doesn't ever make sense to have the block shown. Thoughts?

packages/ui-components/src/components/progress/status.tsx
```diff
@@ -0,0 +1,151 @@
...
+   * Describes the name of the progress step. Think of it as the name of the
+   * todo in the example above.
+   */
+  label: string;
+  /**
+   * Describes when the progress status was last changed. Every mode of progress
+   * status MUST have a date, EXCEPT for "upcoming", which NEVER has a date, (as
+   * it is in the future).
+   *
+   * If no date is passed when one is required, a fallback text will be displayed.
```
Fabricevladimir, line 37: I think you either forgot to add the fallback text or update the comment!

Also, would hiding the date even if it is given when the mode is `upcoming` a better approach? Related to the block prop comment.

packages/ui-components/src/components/progress/status.tsx
```diff
@@ -0,0 +1,151 @@
...
+  IconBlock,
+  IconRadioCancle,
+  IconRadioDefault,
+  IconSpinner,
+  IconSuccess,
+} from '../icons';
+import {LabelProps} from '../label';
+
+type ModeType = 'active' | 'failed' | 'done' | 'succeeded' | 'upcoming';
+export type ProgressStatusProps = {
```
Fabricevladimir, line 13: Very thorough documentation; I like! 👍🏾

packages/ui-components/src/components/progress/status.tsx
```diff
@@ -0,0 +1,151 @@
...
+      </BlockContainer>
+    </TopContainer>
+  );
+};
+
+type ModeProps = {
+  mode: ModeType;
+};
+
+const TopContainer = styled.div.attrs({className: 'flex justify-between'})``;
```
Fabricevladimir, line 83: This seems to be a fixed width on the designs. Currently, the content is squished together on mobile.

Fabricevladimir (changes requested):
Nice work, thanks! Left a few comments here and there. I'm pleasantly surprised at how complicated the component itself turned out to be with the different props affecting each other.

RakeshUP (approved):
Nice. The whole BlockContainer should be conditionally rendered. The other comments are very minor too.

packages/ui-components/src/components/progress/status.tsx
WissenIstNacht: Hmm. You're right. But does that make sense...?

FYI: The reason for the squishing on mobile is not in the component itself but in the way I set up the story. I wrapped it into a div that takes 1/3 of the screen. On mobile, that's obviously way too small and squishes the content. I've modified the story to only do the 1/3 width on tablet or larger.

Should I still fix the width or do you think it makes sense after the correction?

packages/ui-components/src/components/progress/status.tsx
WissenIstNacht: Turns out I pretty much forgot to do all of that stuff. Thanks for catching it!

packages/ui-components/src/components/progress/status.tsx
Fabricevladimir: Ah, that makes sense. I'm not sure why it's fixed width (maybe Figma thing?). To me it makes more sense having it take up the size of the parent.

Fabricevladimir (approved):

PR #130, opened 2022-01-13
feature/Table->Voters component DAO-610

RakeshUP:


packages/ui-components/src/components/table/tableCell.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export type TableCellProps = {
+  type: 'text' | 'foot' | 'head' | 'link' | 'tag';
+  text?: string;
+  subtext?: string;
+  rightAligned?: boolean;
+  bgWhite?: boolean;
+  children?:
+    | React.FunctionComponentElement<BadgeProps>
+    | React.FunctionComponentElement<ButtonTextProps>;
```
WissenIstNacht, line 13: Oops, missed that dependency. Could try to use the link component that Fabrice finished? Don't sweat it if it takes too much time. It looks just as good with the button.

WissenIstNacht: Oops, missed that dependency. Could try to use the link component that Fabrice finished? Don't sweat it if it takes too much time. It looks just as good with the button.

WissenIstNacht (changes requested):
Looks good, thanks! :)

Could you add the a test with the basic test scaffolding?

packages/ui-components/src/components/table/tableCell.tsx
```diff
@@ -0,0 +1,73 @@
...
+  type,
+  text,
+  subtext,
+  bgWhite = false,
+  rightAligned = false,
+  children,
+}) => {
+  if (type === 'text') {
+    return (
+      <td
```
Fabricevladimir: In the designs, the cell types, except for `head` and `foot`, have an inset boxshadow.

packages/ui-components/src/components/table/tableCell.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <p className="font-bold text-primary-500 cursor-pointer">{children}</p>
+      </td>
+    );
+  } else if (type === 'tag') {
+    return (
+      <td className={`px-2 py-1.5 ${!bgWhite && 'bg-ui-0'}`}>{children}</td>
+    );
+  }
+
+  return null;
```
Fabricevladimir: Would it be a better approach to return the most often used type (`text` maybe?) by default instead of null?

packages/ui-components/src/components/table/votersTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+  onLoadMore,
+}) => {
+  return (
+    <Table>
+      <thead>
+        <tr>
+          <TableCell type="head" text="Wallet" />
+          <TableCell type="head" text="Option" />
+          <TableCell type="head" text="Voting Power" rightAligned />
+          <TableCell type="head" text="Token Amount" rightAligned />
```
Fabricevladimir, line 29: I want to cry while writing this, but the table head text should be translatable strings.

packages/ui-components/src/components/table/votersTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+        </tr>
+      </thead>
+      <tbody>
+        {voters.map((voter, index) => (
+          <tr key={index}>
+            <TableCell type="text" text={voter.wallet} />
+            <TableCell type="tag">
+              <Badge
+                label={voter.option}
+                colorScheme={voter.option === 'Yes' ? 'success' : 'critical'}
```
Fabricevladimir, line 39: This is something to think about for future maybe, but are 'yes' and 'no' the only options we have? If so, perhaps the type of the `option` prop should reflect so.

packages/ui-components/src/components/table/votersTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+  );
+};
+
+export const Table = styled.table.attrs({
+  className: 'border-separate',
+})`
+  border-spacing: 0;
+
+  tr th,
+  tr td {
```
Fabricevladimir, line 71: If we could do any of this with TW it would be great for stuff like themes.

Fabricevladimir (changes requested):
Great work! Thanks

packages/ui-components/src/components/table/tableCell.tsx
RakeshUP: Screenshot 2022-01-17 at 6 41 54 PM

It's causing a double border effect in link cell in the footer. The required border style for the whole table is in the table component

packages/ui-components/src/components/table/votersTable.tsx
RakeshUP: Yeah, I thought of it too. Maybe we add i18n to the ui-components library too. @sepehr2github had faced a problem in this PR https://github.com/aragon/zaragoza/pull/139

packages/ui-components/src/components/table/votersTable.tsx
RakeshUP: It seems to be only 2 options for now. I think it is Yes/No option for the widthdrawal related proposals

packages/ui-components/src/components/table/votersTable.tsx
Fabricevladimir: Yes! I very much much like the idea of the translation in ui-components

packages/ui-components/src/components/table/votersTable.tsx
RakeshUP: Adding TW classes using className prop to TableCell component introduces a lot of classes in that component and the code might become difficult to understand. About [TW first and last modifiers](https://v2.tailwindcss.com/docs/hover-focus-and-other-states#:~:text=It%E2%80%99s%20important%20to%20note%20that%20you%20should%20add%20any%20first%3A%20utilities%20to%20the%20child%20element%2C%20not%20the%20parent%20element.)

Fabricevladimir (approved):
Looks good!

WissenIstNacht (approved):
Thanks for the changes! :)

PR #131, opened 2022-01-13
Chore/search input DAO-544 

sepehr2github:
- refactored text input for adding right/left adornment
- add a cross icon to search input
- implement loading state for search input


packages/ui-components/src/components/input/textInput.tsx
```diff
@@ -5,42 +5,43 @@ export type TextInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
   /** Changes a input's color schema */
   mode?: 'default' | 'success' | 'warning' | 'critical';
   /**
-   * adornment
+   * left adornment
    */
-  adornment?: ReactNode;
+  leftAdornment?: ReactNode;
```
WissenIstNacht, line 10: Technically the text inputs do not define any adonrments. But I guess it's ok since they are not mandatory.

WissenIstNacht (approved):
Thanks, looks good! :)

I left a comment regarding the input's props. But don't worry, we can refactor that later.

packages/ui-components/src/components/input/textInput.tsx
sepehr2github: Yes, Both of them are optional

Fabricevladimir (approved):
Good to go!

PR #136, opened 2022-01-13
Feature/Input validation (deposit) - DAO-614 

Fabricevladimir:
- Added validation functions for amount, address
- Form validation for amount input and custom token address
- Disabled next button while form is not valid

Note: Form values shown for easier review & will be removed before merging. 
~~Also, aware that the amount field doesn't validate for alphabetical characters. Will be fixed with valueInput component chore.~~


Fabricevladimir:
**UPDATE:** Last commit restricts user from entering non-numeric characters in the amount field by switching input type from `text` to `number`.

packages/web-app/src/utils/tokens.ts
```diff
@@ -117,8 +118,8 @@ export async function getTokenInfo(
     symbol = null,
     name = null;
 
-  // is Ether
-  if (address === constants.AddressZero) {
+  // if ether (address zero)
```
WissenIstNacht: Yup, I like that better, too. Comment still necessary?

packages/web-app/src/pages/newDeposit.tsx
```diff
@@ -69,9 +69,29 @@ const NewDeposit: React.FC = () => {
     isConnected,
     provider,
   }: useWalletProps = useWallet();
+
+  /**
+   * (Temporary)
+   * NOTE: the mode can also be onBlur (clicking out of the field).
```
WissenIstNacht: Interesting. Could you quickly elaborate on what makes still makes it the better choice than onBlur, given that it comes with performance issues?

packages/web-app/src/pages/newDeposit.tsx
```diff
@@ -36,7 +36,7 @@ const steps = {
 const TOTAL_STEPS = Object.keys(steps).length;
 
 export type FormData = {
-  amount: number;
+  amount: string;
```
WissenIstNacht, line 39: Quick question: FormData seems to hold more stuff than the default Value object. Do you know what useForm does to the rest? Or let me ask differently: Is defaultValues even needed?

packages/web-app/src/containers/depositForm/index.tsx
```diff
@@ -1,28 +1,157 @@
...
+
+      return validationResult;
+    },
+    [provider, resetField]
+  );
+
+  const amountValidator = useCallback(
+    async (amount: string) => {
+      // check if a token is selected using it's address
+      if (tokenAddress === '') return 'A token must be selected';
```
WissenIstNacht: Would you not need translation here, too?

packages/web-app/src/containers/depositForm/index.tsx
```diff
@@ -111,21 +260,32 @@ const DepositForm: React.FC = () => {
...
+                name={name}
+                type="number"
+                value={value}
+                onBlur={onBlur}
+                onChange={onChange}
+                adornmentText={t('labels.max')}
+                onAdornmentClick={() => handleMaxClicked(onChange)}
+              />
+              {error?.message && (
+                <AlertInline label={error.message} mode="critical" />
```
WissenIstNacht: Minor detail here: The Inline Alert moves the max balance label down. Do you think you can easily add a div around it? (Taking into account that you'd have to deal with the question of what happens when the balance is large enough to clash with the alert)

WissenIstNacht (approved):
Solid work man, thanks.

Some minor suggestions and a few questions. But the PR is good to go from my POV.

packages/web-app/src/pages/newDeposit.tsx
Fabricevladimir: Because the `onChange` validation runs, well, on change, the UI appears to be more responsive. Also, the performance lag can only be seen when interacting with an input that has "heavy" validation (currently, `tokenAddress`). The idea is since the token address will most probably be pasted in, the `onChange` isn't dispatched often, which means the lag won't be noticed. As for whether it's better than `onBlur`, it depends on what is wanted in terms of UI interactivity. `onBlur` of course will require the user to click/focus out of the input before the validation is ran. Essentially, I picked one and ran with it!

packages/web-app/src/pages/newDeposit.tsx
Fabricevladimir: The rest are considered undefined. What I found was if there wasn't a default value provided for controlled inputs, the form would throw an error because it considers such an input to be both controlled and uncontrolled. 

Alternatively, a `defaultValue` can be passed to the Controller of each component.

packages/web-app/src/containers/depositForm/index.tsx
Fabricevladimir: Hmm, I can add that. We need to check in with Selim for what to do about when the two clash.

packages/web-app/src/pages/newDeposit.tsx
WissenIstNacht: Right, got you. I think you make a good point with the pasting mitigating the performance issues.

packages/web-app/src/pages/newDeposit.tsx
WissenIstNacht: Interesting, thanks!

packages/web-app/src/containers/depositForm/index.tsx
```diff
@@ -1,28 +1,156 @@
...
+
+  /*************************************************
+   *                    Hooks                      *
+   *************************************************/
+  useEffect(() => {
+    if (!account) return;
+
+    const fetchTokenInfo = async () => {
+      if (errors.tokenAddress !== undefined) {
+        if (dirtyFields.amount) trigger(['amount', 'tokenSymbol']);
```
sepehr2github, line 47: What Do you mean by dirtyFields?

sepehr2github (approved):
Just a single question. Looks Great! Thank you Fabrice.

packages/web-app/src/containers/depositForm/index.tsx
Fabricevladimir: Those are fields that have been touched and changed at any time during the life of the form

PR #139, opened 2022-01-13
Feature/card proposal DAO-595 

sepehr2github:
- proposal card component added
- automated breakpoint based on screen size 
- add two more colors to the badge component (tags)

Note: Please, everyone, review this pr. I made some decisions and I need to know your ideas. Especially for button label prop and alert message prop

Thank you!


packages/ui-components/src/components/cards/cardProposal.tsx
```diff
@@ -0,0 +1,218 @@
...
+    </Card>
+  );
+};
+
+const Card = styled.div.attrs({
+  className: 'flex justify-between flex-col bg-white rounded-xl p-3 space-y-3',
+})``;
+
+const Header = styled.div.attrs({
+  className: 'flex space-y-1 justify-between',
```
RakeshUP: This space-y-1 is causing the badge height to increase. This can be seen when the state is 'Active' or 'Pending'

RakeshUP (approved):
Approved. Just a single comment from me

packages/ui-components/src/components/cards/cardProposal.tsx
sepehr2github: Ah thank you for mentioning

packages/ui-components/src/components/cards/cardProposal.tsx
```diff
@@ -0,0 +1,218 @@
...
+  publisherAddress?: Address;
+  /**
+   * Button label for different status
+   * ['pending / executed / defeated label', 'active label', 'succeeded label', 'draft label']
+   * TODO: I thought to add 4 button Label
+   * props for different states and implement
+   * condition here but i decided to use only one prop
+   * for reducing the complexity
+   */
+  buttonLabel: string[];
```
Fabricevladimir, line 46: Thanks for the documentation! I understand why this is being done, but I also think this shows a need to revisit our approach to translations in the ui-components.

packages/ui-components/src/components/cards/cardProposal.tsx
```diff
@@ -0,0 +1,218 @@
...
+  publishLabel,
+  publisherAddress,
+  buttonLabel,
+  AlertMessage,
+  onClick,
+}: CardProposalProps) => {
+  const headerOptions: {
+    [key in CardProposalProps['state']]: ReactNode;
+  } = {
+    draft: <Badge label="Draft" />,
```
Fabricevladimir: Hold off on this change for now, but the labels will need to be translated as well 😬

packages/ui-components/src/components/cards/cardProposal.tsx
```diff
@@ -0,0 +1,218 @@
...
+    pending: (
+      <>
+        <Badge label="Pending" />
+        <AlertInline label={AlertMessage[0] || ''} />
```
Fabricevladimir: Very minor. While not having an alert message shouldn't happen, it might be best to show the `AlertInline` conditionally. Currently if the message is passed in, the icon still gets rendered.

image

packages/ui-components/src/components/cards/cardProposal.tsx
```diff
@@ -0,0 +1,218 @@
...
+  };
+
+  return (
+    <Card data-testid="cardProposal">
+      <Header>{headerOptions[state]}</Header>
+      <TextContent>
+        <Title>{title}</Title>
+        <Description>{description}</Description>
+        <Publisher>
+          {publishLabel} <Link label={shortenAddress(publisherAddress || '')} />
```
Fabricevladimir: Are we certain this is a `Link` component? If so, where does it lead to?

Also, the link text should be "text-sm" ("text-base" should be removed for the Label in the `Link` component so that any text style can be passed into the component like so `<Link className="text-sm" href="abc" />`)

packages/ui-components/src/components/cards/cardProposal.tsx
```diff
@@ -0,0 +1,218 @@
...
+        </Publisher>
+      </TextContent>
+      {state === 'active' && (
+        <LoadingContent>
+          <ProgressInfoWrapper>
+            <ProgressTitle>{voteTitle}</ProgressTitle>
```
Fabricevladimir, line 152: If the voteTitle is long, we need to handle how it gets displayed on mobile.

packages/ui-components/src/components/cards/cardProposal.tsx
```diff
@@ -0,0 +1,218 @@
...
+            <TokenAmount>
+              {tokenAmount} {tokenSymbol}
+            </TokenAmount>
+          </ProgressInfoWrapper>
+          <LinearProgress max={100} value={voteProgress} />
```
Fabricevladimir, line 157: Nice shortcut! Saves a prop

Fabricevladimir (commented):
Thanks! Nice work; left a couple of comments

packages/ui-components/src/components/cards/cardProposal.tsx
sepehr2github: The vote title is a fixed text 'Winning Option' It's variable for the translation

Fabricevladimir (approved):
Ready to roll!

PR #141, opened 2022-01-14
Fixed the forward state check in DisputableProcess

nivida:
The require was missleading and there was also a missing one to check if the execution isn't already entirely stopped from his execution.


nivida:
Approved after a quick chat on Discord.

PR #145, opened 2022-01-17
Fix/Add custom token button - DAO-612 

Fabricevladimir:
- Add custom token is now only shown after the user has searched for a token that is not in the list


WissenIstNacht (approved):
Works, thanks :)

sepehr2github (approved):
Nice! good to go

PR #147, opened 2022-01-17
Chore: Default font-weight

WissenIstNacht:
This PR changes the default fontweight of the ui-components and the app from 400 to 500.

This is done because the default font-weights of the design is 500. HTML on the other hand as a default of 400. This means that every instance of non-bold text must be modified with the `font-medium` TW class. This is cumbersome and easy to forget.

I have not found a way to do this in TW, unfortunately.

Also, I have not refactored the web-app, since I didn't want to cause any conflicts.

### DIFF
--- a/packages/web-app/index.html
+++ b/packages/web-app/index.html
@@ -10,7 +10,8 @@
     href="'https://fonts.googleapis.com/css2?family=Manrope:wght@500;700&display=swap'"
     rel="stylesheet" />
   <style>
-    p {
+    p,
+    span {
       font-weight: 500;
     }
   </style>
--- a/packages/web-app/index.html
+++ b/packages/web-app/index.html
@@ -1,61 +1,66 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <link rel="icon" type="image/svg+xml" href="/src/public/favicon.svg" />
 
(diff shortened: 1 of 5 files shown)


packages/web-app/index.html
```diff
@@ -1,61 +1,66 @@
...
-          rudderAnalytics.page();
+  <!-- TODO: remove once fonts have been downloaded and set in stone -->
+  <link
+    href="'https://fonts.googleapis.com/css2?family=Manrope:wght@500;700&display=swap'"
+    rel="stylesheet" />
+  <style>
+    p {
+      font-weight: 500;
+    }
+  </style>
```
WissenIstNacht: These are the only actual changes in this file. The rest is just formatting.

sepehr2github (approved):
Thank you, Vali. Looks good to me. the only thing we should care about is to remember adding the API key later.

packages/web-app/index.html
sepehr2github: btw Vali, Don't you think we should add this font-weight for the 'span' element?

WissenIstNacht: Eeeeeeexcellent point good point, thanks! 👍🏼

RakeshUP (approved):

WissenIstNacht:
@sepehr2github I just checked regarding the key. Apparently, this was done moths ago 😅

I'll delete the TODO and merge the PR :)

PR #152, opened 2022-01-18
Feature/simplevoting tests

novaknole:


packages/contracts/contracts/core/processes/Process.sol
```diff
@@ -132,21 +132,19 @@ abstract contract Process is Component {
     /// @return executionId The id of the newly created execution.
     function start(Proposal calldata _proposal) 
         external
```
nivida: Probably worth to do a quick call related the permissions. But I see the point of making it optional instead of enforcing it. The only thing I would like to see is to have consistent role identifiers over different concrete process implementation cause it increases the composability of those.

packages/contracts/contracts/processes/votings/simple/SimpleVoting.sol
```diff
@@ -51,12 +50,16 @@ contract SimpleVoting is VotingProcess, TimeHelpers {
     event StartVote(uint256 indexed voteId, address indexed creator, string description);
     event CastVote(uint256 indexed voteId, address indexed voter, bool voterSupports, uint256 stake);
     event ExecuteVote(uint256 indexed voteId);
-    event ChangeSupportRequired(uint64 supportRequiredPct);
-    event ChangeMinQuorum(uint64 minAcceptQuorumPct);
+    event ChangeConfig(uint64 supportRequiredPct, uint64 minAcceptQuorumPct);
```
nivida, line 53: Do we want to implement the generic handling of configuration properties for all concrete processes on a later point?

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -116,6 +118,12 @@ contract DAOFactory {
         items[6] = ACLData.BulkItem(ACLData.BulkOp.Revoke, dao.ROOT_ROLE(), address(this));
 
         dao.bulk(address(dao), items);
+
+        address ANY_ADDR = address(type(uint160).max);
+
+        // give voting AND executing capabilities to anyone on the voting process
+        dao.grant(address(voting), ANY_ADDR, voting.PROCESS_VOTE_ROLE());
+        dao.grant(address(voting), ANY_ADDR, voting.PROCESS_EXECUTE_ROLE());
```
nivida: Here is at least one role missing (MODIFY_CONFIG).. but as said let us discuss about this at all..

packages/contracts/contracts/processes/votings/simple/SimpleVoting.sol
```diff
@@ -10,8 +10,7 @@ import "./../../../core/DAO.sol";
 import "../../../utils/TimeHelpers.sol";
 
 contract SimpleVoting is VotingProcess, TimeHelpers {
-    bytes32 public constant MODIFY_SUPPORT_ROLE = keccak256("MODIFY_SUPPORT_ROLE");
-    bytes32 public constant MODIFY_QUORUM_ROLE = keccak256("MODIFY_QUORUM_ROLE");
+    bytes32 public constant MODIFY_CONFIG = keccak256("MODIFY_VOTE_CONFIG");
```
nivida, line 13: The last time I looked at this implementation of you did we had duplicated events for when a new voting does get created and also when a vote does get casted. Probably worth to check/fix and also discuss about if required.

packages/contracts/hardhat.config.ts
```diff
@@ -39,6 +39,14 @@ const config: HardhatUserConfig = {
       accounts:
         process.env.PRIVATE_KEY !== undefined ? [process.env.PRIVATE_KEY] : [],
     },
+    hardhat: {
+      throwOnTransactionFailures: true,
+      throwOnCallFailures: true,
+      allowUnlimitedContractSize: true,
```
nivida: ```allowUnlimitedContractSize``` where did you reach this? I would say this is actually a problem..

packages/contracts/hardhat.config.ts
```diff
@@ -39,6 +39,14 @@ const config: HardhatUserConfig = {
...
+      blockGasLimit: 0x1fffffffffffff,
```
nivida: blockGasLimit: I think it is right now around 15mio on eth L1...

packages/contracts/test/registry.ts
```diff
@@ -28,9 +28,10 @@ describe('Registry', function () {
     const daoName = 'my-dao';
     const daoAddress = wallet.address;
 
-    await expect(await registry.register(daoName, daoAddress, ownerAddress))
-          .to.emit(registry, EVENTS.NewDAORegistered)
-          .withArgs(daoAddress, ownerAddress, daoName)
+    await expect(
```
nivida, line 31: Ignoring code style here cause we have to discuss about and configure prettier accordingly.

packages/contracts/test/test-utils/equal-overwrite.ts
```diff
@@ -0,0 +1,57 @@
+import { BigNumber } from 'ethers'
+
+// this function is adapted from Waffle
+const deepEqual = (actual: any, expected: any): boolean => {
```
nivida, line 4: lol this is still not fixed in Waffle?

packages/contracts/test/test-utils/index.ts
```diff
@@ -0,0 +1,5 @@
+import { equalOverwrite } from './equal-overwrite'
+
+export default function (chai: Chai.ChaiStatic, utils: Chai.ChaiUtils) {
```
nivida, line 3: probably better to have named exports already here.. also if not yet really required..

packages/contracts/test/voting/simple-voting.ts
```diff
@@ -0,0 +1,401 @@
+import chai, { expect } from 'chai'
```
nivida: Probably worth to use semicolons.. it also prevents us from some strange edge cases were ugly syntax is required.

packages/contracts/test/voting/simple-voting.ts
```diff
@@ -0,0 +1,401 @@
...
+            )   
+    }
+
+    describe("initialize: ", async () => {
+        it("reverts if quorom is less than support required", async () => {
+            await expect(
+                initializeVoting([2, 1,3],[])
+            ).to.be.revertedWith(ERRORS.ERROR_INIT_PCTS)
+        })
+        it("reverts if trying to re-initialize", async () => {
```
nivida, line 116: I will move this and also care about the todo below cause those two cases test the abstract contract. Just here as a hint.

packages/contracts/test/voting/simple-voting.ts
```diff
@@ -0,0 +1,401 @@
...
+                .withArgs(20, 10)
+        })
+    })
+
+    describe("StartVote", async () => {
+        beforeEach(async () => {
+            await initializeVoting([1, 2, 3], [])
+        })
+
+        it("reverts if allowed action is not passed through proposal", async () => {
```
nivida, line 158: I will also remove this cause of the Process test I will merge asap.

packages/contracts/contracts/processes/votings/simple/SimpleVoting.sol
novaknole: correct. For now, let's leave it like this..

packages/contracts/test/test-utils/equal-overwrite.ts
novaknole: Same stuff. Waffle really seems to be kind of abandoned.

nivida: Let us take a note to probably move to a better tooling eg Foundry later on..

PR #153, opened 2022-01-18
Feature/Checkbox ListItem DAO-615

RakeshUP:


packages/ui-components/src/components/checkbox/checkboxListItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+  isMulti?: boolean;
+};
+
+export const CheckboxListItem: React.FC<CheckboxListItemProps> = ({
+  label,
+  helptext,
+  multiSelect,
+  isActive = false,
+  disabled = false,
+  isMulti = false,
```
Fabricevladimir: Can you explain a bit what this prop is for? It's a bit ambiguous especially with the other prop `multiselect`

packages/ui-components/src/components/checkbox/checkboxListItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+    default: <IconCheckboxDefault />,
+  },
+  radio: {
+    active: <IconRadioSelected />,
+    multi: <IconRadioDefault />,
+    default: <IconRadioDefault />,
+  },
+};
+
+export type CheckboxListItemProps = {
```
Fabricevladimir, line 24: How do we toggle the state of the checkbox?

Fabricevladimir (commented):
Good work, and thanks for fixing the circular dependency issue. Left a couple of questions.

packages/ui-components/src/components/checkbox/checkboxListItem.tsx
sepehr2github: There is no onClick or onChange implemented so when I click on the list item nothing happened!. And I think we should have a better way of handling the props. So based on the design we have two types (multi-selection: yes or no) and we have three states for multi-selection (default/active/multi) and two for single selection (default/active) all status would change by user clicks

packages/ui-components/src/components/checkbox/checkboxListItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const iconType = isActive ? 'active' : isMulti ? 'multi' : 'default';
+
+  return (
+    <Container
+      data-testid="checkboxListItem"
+      isActive={isActive}
+      disabled={disabled}
+      isMulti={isMulti}
+    >
+      <HStack isActive={isActive} disabled={disabled} isMulti={isMulti}>
```
sepehr2github: and also can you add cursor: pointer for hover?

sepehr2github (commented):
Thank you! Rakash. I left a few comments for you. 👍

RakeshUP:
@Fabricevladimir @sepehr2github I've added onclick prop and a controlled way to toggle the state inside the component itself. I've changed the other ambiguous props too

Fabricevladimir (approved):
Looks great! Thanks

sepehr2github (approved):

PR #155, opened 2022-01-19
Feature: DAO 597 - widget status

WissenIstNacht:
This PR:

- implements the status widget
- implements the corresponding story
- implements the corresponding tests
- resolves the circular dependency issues (cherry picked from Rakesh)

### DIFF
--- a/packages/ui-components/src/components/widgets/status.tsx
+++ b/packages/ui-components/src/components/widgets/status.tsx
@@ -9,11 +9,15 @@ export type WidgetStatusProps = {
 
 export const WidgetStatus: React.FC<WidgetStatusProps> = ({data}) => {
   return (
-    <Card>
+    <Card data-testid="widgetStatus">
       <Header>Status</Header>
-      {data.map(d => {
-        return <ProgressStatus {...d} />;
-      })}
+      {data ? (
+        data.map(d => {
+          return <ProgressStatus {...d} />;
+        })
+      ) : (
+        <p className="text-ui-400">Progress unavailable</p>
+      )}
     </Card>
   );
 };
--- a/packages/ui-components/src/components/widgets/status.tsx
+++ b/packages/ui-components/src/components/widgets/status.tsx
@@ -4,16 +4,22 @@ import styled from 'styled-components';
 import {ProgressStatus, ProgressStatusProps} from '../progress/status';
 
 export type WidgetStatusProps = {
-  data: ProgressStatusProps[];
+  /**
+   * The widget status displays the status of a process. Such an process
+   * typically consists of a number of steps. Each of these steps has a set of
+   * attributes (see `ProgressStatusProps`). These attributes must be passed as
+   * an array of objects.
+   */
+  steps: ProgressStatusProps[];
 };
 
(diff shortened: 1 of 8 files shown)


packages/ui-components/stories/widgetStatus.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+    mode: 'succeeded',
+    block: '132,123,231',
+  },
+];
+
+/* STORIES ================================================================== */
+
+export const Published = Template.bind({});
+Published.args = {
+  data: stepDataPublished,
```
RakeshUP: Is this prop supposed to be "steps"? It's the same for all stories

RakeshUP (approved):
I changed the props in stories and tested the component. Looks good!

packages/ui-components/stories/widgetStatus.stories.tsx
WissenIstNacht: Yes, thanks!

packages/ui-components/stories/widgetStatus.stories.tsx
```diff
@@ -0,0 +1,117 @@
...
+  steps: stepDataReady,
+};
+
+export const Executed = Template.bind({});
+Executed.args = {
+  steps: stepDataExecuted,
+};
+
+export const Empty = Template.bind({});
+Empty.args = {};
```
Fabricevladimir: Because `steps` is required, if we actually use the component this way, TS will go crazy

packages/ui-components/src/components/widgets/status.tsx
```diff
@@ -0,0 +1,37 @@
...
+   * an array of objects.
+   */
+  steps: ProgressStatusProps[];
+};
+
+export const WidgetStatus: React.FC<WidgetStatusProps> = ({steps}) => {
+  return (
+    <Card data-testid="widgetStatus">
+      <Header>Status</Header>
+      {steps ? (
```
Fabricevladimir: Because `steps` is required, when the component is used properly, the falsy path shouldn't ever trigger... Should we compare it to an empty array instead? Or if this is to prevent the component from displaying incorrectly because the user didn't pass in `steps`, it might be better to make `steps` optional and default it to an empty array. Thoughts?

Fabricevladimir (approved):
Good work, thanks! Approved, and left a couple of comments

packages/ui-components/stories/widgetStatus.stories.tsx
WissenIstNacht: The problem is while steps is required, it can be empty, or have undefined members. Unfortunately, there's no simple way to force steps to be non-empty from a type point of view.

packages/ui-components/src/components/widgets/status.tsx
WissenIstNacht: Ahm, yes. So as mentioned in the other comments the issue is that steps can be empty. Which this should have mitigated. BUt admittedly, I thought `[] === false` while I was doing this, which turns out not to be the case haha

I'll change it to guard against empty arrays instead.

PR #158, opened 2022-01-20
Feature/Add balance - DAO-643 

Fabricevladimir:
- Shows DAO token balance on withdraw
- Adds max button for token amount
- Shows warning if user puts in an amount greater than the DAO token balance


WissenIstNacht (approved):
Thanks,  works fine for me! :)

RakeshUP (approved):
Nice!

PR #159, opened 2022-01-20
Feature/pagination

sepehr2github:
- pagination added
- fix the storybook background settings

### DIFF
--- a/packages/ui-components/src/components/pagination/pagination.tsx
+++ b/packages/ui-components/src/components/pagination/pagination.tsx
@@ -1,4 +1,4 @@
-import React, {useState} from 'react';
+import React, {useState, useEffect} from 'react';
 import styled from 'styled-components';
 import {IconChevronRight, IconChevronLeft} from '../icons/interface';
 
@@ -15,7 +15,7 @@ export interface PaginationProps {
    * active steps
    */
   defaultStep?: number;
-  onChange?: () => void;
+  onChange?: (step: number) => void;
 }
 
 /**
@@ -25,10 +25,16 @@ export const Pagination: React.FC<PaginationProps> = ({
   totalSteps = 10,
   defaultStep = 1,
   bgWhite = false,
-  // onChange,
+  onChange,
 }) => {
   const [step, setStep] = useState<number>(defaultStep);
   const MovetoNextStep = (step: number) => setStep(step);
+
+  useEffect(() => {
+    if (onChange) {
+      onChange(step);
+    }
+  }, [onChange, step]);
 
   function ButtonList(
     activeStep: number,
@@ -36,8 +42,8 @@ export const Pagination: React.FC<PaginationProps> = ({
     MovetoNextStep: (step: number) => void
   ) {
     const List = [];
-    if (activeStep < 5) {
-      for (let i = 1; i < 6; i++) {
+    if (activeStep <= 5) {
+      for (let i = 1; i <= Math.min(5, totalSteps); i++) {
         List.push(
           <Page
             isActive={activeStep === i}
@@ -51,7 +57,7 @@ export const Pagination: React.FC<PaginationProps> = ({
       return (
         <>
           {List}
-          {totalSteps > 6 && (
+          {totalSteps > 5 && totalSteps !== 6 && (
             <>
               <Separator>...</Separator>
               <Page
@@ -64,7 +70,7 @@ export const Pagination: React.FC<PaginationProps> = ({
           )}
         </>
       );
-    } else if (activeStep >= 5 && activeStep <= totalSteps - 5) {
+    } else if (activeStep >= 5 && activeStep <= totalSteps - 4) {
       for (let i = activeStep - 1; i <= activeStep + 1; i++) {
         List.push(
           <Page
@@ -92,8 +98,8 @@ export const Pagination: React.FC<PaginationProps> = ({
           </Page>
         </>
       );
-    } else if (activeStep > totalSteps - 5) {
-      for (let i = totalSteps - 5; i <= totalSteps; i++) {
+    } else if (totalSteps - 5 && activeStep > totalSteps - 4) {
+      for (let i = totalSteps - 4; i <= totalSteps; i++) {
         List.push(
           <Page
             isActive={activeStep === i}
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,176 @@
...
+   */
+  bgWhite?: boolean;
+  /**
+   * Number of total steps
+   */
+  totalSteps?: number;
+  /**
+   * active steps
+   */
+  defaultStep?: number;
```
WissenIstNacht: I'm not sure I understand this. When I change it to something else on the story, nothing seems to change.

//Edit: Nvm, I got it :D But maybe calling it \`activePage\` would be better. In the beginning I thought
 maybe this is how far you \_step\_ when you click on the chevron buttons. But it's realy just a way for the user to set the current active page, right?

//Edit 2: In general, maybe it would be better to change step to pages in the most cases? For example, \`totalPages\` instead of \`totalStep\`. Or \`setPage\` instead of \`setSteps\`.

packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,176 @@
...
+  onChange,
+}) => {
+  const [step, setStep] = useState<number>(defaultStep);
+
+  useEffect(() => {
+    onChange && onChange(step);
+  }, [onChange, step]);
+
+  function ButtonList() {
+    const List = [];
```
WissenIstNacht: lowercase :p

packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,176 @@
...
+    <HStack data-testid="pagination">
+      <ButtonIcon
+        mode="secondary"
+        size="large"
+        onClick={() => setStep(step - 1)}
+        disabled={step === 1}
+        icon={<IconChevronLeft />}
+        {...(bgWhite && {bgWhite})}
+      />
+      {ButtonList()}
```
WissenIstNacht: ButtonList is pretty much a React component, no? You coul actually use this in JSX, then. Like `<ButtonList />`

packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,176 @@
...
+            mode="secondary"
+            size="large"
+            isActive={step === i}
+            onClick={() => setStep(i)}
+            {...(bgWhite && {bgWhite})}
+            label={`${i}`}
+          />
+        );
+      }
+      return <>{List}</>;
```
WissenIstNacht:
```suggestion
      return List;
```

Maybe a bit less complicatedd :p

packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    if (step - 1 <= Distance) {
+      for (let i = 1; i <= 5; i++) {
```
WissenIstNacht, line 63: careful here: the maximum of the loop is actually related to the `distance` variable. Maybe you should express 5 as in terms of distance

// Edit: nevermid, I think I misunderstood your code. 5 is the number of pages shown before the pagination window starts moving, correct? so 1,2,3,4,5 on the beginning and 96,97,98,99,100 in the end (assuming a total of 100 pages). This should definitely be a variable too (since you use it in two places), and it might even be worth exposign this as a prop as well.

packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,176 @@
...
+}) => {
+  const [step, setStep] = useState<number>(defaultStep);
+
+  useEffect(() => {
+    onChange && onChange(step);
+  }, [onChange, step]);
+
+  function ButtonList() {
+    const List = [];
+    const Distance = 3;
```
WissenIstNacht: Not sure, but it might be worth exposing this as a prop.

Also, lowercase :D

packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,176 @@
...
+            onClick={() => setStep(totalSteps)}
+            {...(bgWhite && {bgWhite})}
+            label={`${totalSteps}`}
+          />
+        </>
+      );
+    }
+
+    if (totalSteps - step <= Distance) {
+      for (let i = totalSteps - 5; i <= totalSteps; i++) {
```
WissenIstNacht: You have a off by one error here. Should be 4 instead of 5.

```suggestion
      for (let i = totalSteps - 4; i <= totalSteps; i++) {
```

But ideally it should be `totalSteps - (tailSize - 1)`, where `tailSize` is the numer of elements you want to show at the beginning/end.

WissenIstNacht (changes requested):
Thanks a lot, cool component! :)

I found a small error in the logic. Also, I left some comments on the code itself (variable naming, props and such). Let me know if you dis- / agree!

packages/ui-components/src/components/pagination/pagination.tsx
sepehr2github: the return of the arrey is a react element and can not return an array. So I tried to return an array as react element

packages/ui-components/src/components/pagination/pagination.tsx
sepehr2github: That's exactly what you said!

packages/ui-components/src/components/pagination/pagination.tsx
sepehr2github: Thanks, you're right!

RakeshUP (approved):
Looks good.

packages/ui-components/src/components/pagination/pagination.tsx
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * active page
+   */
+  activePage?: number;
+  /**
+   * With this parameter we can define break points for
+   * change pagination from 1...567...9 to 1...789
+   * views
+   */
+  distance?: number;
```
WissenIstNacht, line 23: Thanks a lot the changes and exposing this.

There is still an edge case that is slightly wrong. If you set the distance to one and click the next page button twice you get `1 ... 2 3 4 ... 10`. But it's not a huge deal.

I'll approve the PR.

WissenIstNacht (approved):

packages/ui-components/src/components/pagination/pagination.tsx
sepehr2github: Thanks. Selim said the distance will be fixed

PR #164, opened 2022-01-23
Feature/factory tests

novaknole:


packages/contracts/contracts/core/processes/Process.sol
```diff
@@ -140,7 +140,10 @@ abstract contract Process is Component {
             
             for (uint256 i = 0; i < actionsLength; i++) {
                 IDAO.Action calldata action = _proposal.actions[i];
-                if (allowedActions[action.to][bytes4(action.data[:4])] == false) {
+                if (
+                    allowedActions[action.to][bytes4(action.data[:4])] == false && 
+                    action.to != address(this)
```
nivida: why this check?

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -15,78 +15,88 @@ import "./../registry/Registry.sol";
...
+    Registry public registry;
+    TokenFactory public tokenFactory;
+    
+    struct DAOConfig {
         string name;
-        string symbol;
+        bytes metadata;
+    }
+
+    struct AllowedActions {
```
nivida: Can you elaborate closer on this struct? from my perspective is this not required :))

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -15,78 +15,88 @@ import "./../registry/Registry.sol";
...
-            GovernanceWrappedERC20(token).initialize(
-                IERC20Upgradeable(_tokenConfig.addr),
-                _tokenConfig.name,
-                _tokenConfig.symbol
-            );
-        }
-
+        DAOConfig calldata _daoConfig,
+        TokenFactory.TokenConfig calldata _tokenConfig,
+        TokenFactory.MintConfig calldata _mintConfig,
```
nivida, line 67: Can't we have token and mint config in the same struct?

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -15,78 +15,88 @@ import "./../registry/Registry.sol";
...
-            token = Clones.clone(governanceWrappedERC20Base);
-            // user already has a token. we need to wrap it in our new token to make it governance token.
-            GovernanceWrappedERC20(token).initialize(
-                IERC20Upgradeable(_tokenConfig.addr),
-                _tokenConfig.name,
-                _tokenConfig.symbol
-            );
-        }
-
+        DAOConfig calldata _daoConfig,
```
nivida, line 65: We have to think here about how we can pass the minimum amount of bytes for the creation of a DAO.. It will have quite some impact on how much a DAO creation costs..🤔

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -15,78 +15,88 @@ import "./../registry/Registry.sol";
...
     }
 
     // @notice Creates a new DAO based with his name, token, metadata, and the voting settings.
-    // @param name The DAO name as string
-    // @param _metadata The IPFS hash pointing to the metadata JSON object of the DAO
-    // @param _tokenConfig The address of the token, name, and symbol. If no addr is passed will a new token get created.
-    // @return dao The DAO contract created
-    // @return voting The voting process for this DAO - Currently a hard-coded process. With the planned marketplace will this be more dynamic.
-    // @return token The token passed or created that belongs to this DAO. - Probably not a requirement in the future.
+    // @param _daoConfig The DAO name and metadata
```
nivida: I think the natspec is missing a few definitions.

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -15,78 +15,88 @@ import "./../registry/Registry.sol";
...
+    ) {
+        require(_mintConfig.tos.length == _mintConfig.amounts.length, ERROR_MISMATCH);
+        
+        // create dao
         dao = DAO(createProxy(daoBase, bytes("")));
+        // initialize dao
+        dao.initialize(_daoConfig.metadata, address(this));  
+
+        // Create token and merkle minter
+        dao.grant(address(dao), address(tokenFactory), dao.ROOT_ROLE());
```
nivida, line 83: Probably worth to comment here why we grant the TokenFactory permissions on the DAO. Noticed this cause I had to read the TokenFactory contract first to understand why exactly.

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -106,8 +116,21 @@ contract DAOFactory {
         // Add voting process
         dao.addProcess(voting);
 
-        ACLData.BulkItem[] memory items = new ACLData.BulkItem[](7);
+        // set roles on the Voting Process
+        ACLData.BulkItem[] memory items = new ACLData.BulkItem[](4);
```
nivida: Can't we do all those permissions within one ACL.Bulk?

packages/contracts/contracts/factories/DAOFactory.sol
```diff
@@ -120,25 +143,13 @@ contract DAOFactory {
...
-        address ANY_ADDR = address(type(uint160).max);
-
-        items[0] = ACLData.BulkItem(ACLData.BulkOp.Grant, voting.PROCESS_VOTE_ROLE(), ANY_ADDR);
-        items[1] = ACLData.BulkItem(ACLData.BulkOp.Grant, voting.PROCESS_EXECUTE_ROLE(), ANY_ADDR);
-        items[2] = ACLData.BulkItem(ACLData.BulkOp.Grant, voting.PROCESS_START_ROLE(), ANY_ADDR);
-        items[3] = ACLData.BulkItem(ACLData.BulkOp.Grant, voting.MODIFY_CONFIG(), address(dao));
-
-        dao.bulk(address(voting), items);
+    
+        emit DAOCreated(_daoConfig.name, address(token), address(voting));
```
nivida, line 131: Probably worth to sync about the registry and daofactory events on a quick call.. we could also have just here in the factory one event and no event in the registry..

packages/contracts/contracts/factories/TokenFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+    MerkleDistributor private distributorBase;
+
+    struct TokenConfig {
+        address addr;
+        string name;
+        string symbol;
+    }
+
+    struct MintConfig {
+        address[] tos;
```
nivida: I would use a better name as ``tos``

packages/contracts/contracts/factories/TokenFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+        uint256[] amounts;
+    }
+
+    event DAOCreated(string name, address indexed token, address indexed voting);
+
+    constructor() {
+        setupBases();
+    }
+
+    // TODO: Worth considering the decimals ?
```
nivida: yup definitely worth :-)

packages/contracts/contracts/factories/TokenFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // TODO: Worth considering the decimals ? 
+    // @notice Creates or Wraps Token based with its name and symbol.
+    // @param _dao The dao address
+    // @param _metadata The IPFS hash pointing to the metadata JSON object of the DAO
+    // @param _tokenConfig The address of the token, name, and symbol. If no addr is passed will a new token get created.
+    // @return _mintConfig Where to mint initial tokens
+    // @return ERC20VotesUpgradeable Created or Wrapped Token
+    // @return MerkleMinter Merkle Minter
```
nivida: natspec is misleading

packages/contracts/contracts/factories/TokenFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+        TokenConfig calldata _tokenConfig,
+        MintConfig calldata _mintConfig
+    ) external returns (ERC20VotesUpgradeable, MerkleMinter) {
+        address token = _tokenConfig.addr;
+
+        // deploy token
+        if(token != address(0)) {
+            token = governanceWrappedERC20Base.clone();
+            // user already has a token. we need to wrap it in our new 
+            // token to make it governance token.
```
nivida: probably worth in general to check the grammar of the added comments.. sorrey😇

packages/contracts/contracts/factories/TokenFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+    // @param _tokenConfig The address of the token, name, and symbol. If no addr is passed will a new token get created.
+    // @return _mintConfig Where to mint initial tokens
+    // @return ERC20VotesUpgradeable Created or Wrapped Token
+    // @return MerkleMinter Merkle Minter 
+    function newToken(
+        DAO _dao,
+        TokenConfig calldata _tokenConfig,
+        MintConfig calldata _mintConfig
+    ) external returns (ERC20VotesUpgradeable, MerkleMinter) {
+        address token = _tokenConfig.addr;
```
nivida, line 61: I think this isn't required cause ``token`` as defined here will only be used on line 65 and otherwise overwritten.

packages/contracts/contracts/factories/TokenFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+                _tokenConfig.symbol
+            );
+
+            return (
+                ERC20VotesUpgradeable(token), 
+                MerkleMinter(address(0))
+            );
+        }
+
+        // IMPORTANT: If creator already has a token, It's that token's
```
nivida: Probably worth to move this comment to the function natspec.

packages/contracts/contracts/factories/TokenFactory.sol
```diff
@@ -0,0 +1,132 @@
...
+    // @dev private helper method to set up the required base contracts on TokenFactory deployment.
+    function setupBases() private {
+        distributorBase = new MerkleDistributor();
+        governanceERC20Base = address(new GovernanceERC20());
+        governanceWrappedERC20Base = address(new GovernanceWrappedERC20());
+        merkleMinterBase = address(new MerkleMinter());
+    }
+}
+
+
```
nivida: one break too much🤓

packages/contracts/contracts/permissions/DAOPermissionHandler.sol
```diff
@@ -0,0 +1,23 @@
...
+abstract contract DAOPermissionHandler {
+    
+    IDAO internal dao;
+
+    modifier auth(bytes32 _role)  {
+        require(dao.hasPermission(address(this), msg.sender, _role, msg.data), "component: auth");
+        _;
+    }
+
+    function __Initialize_DAO_Permission(IDAO _dao) internal {
```
nivida: Nice you used here the recommended naming of OZ... should we after this PR apply it to all contracts? WDYT?

packages/contracts/contracts/permissions/DAOPermissionHandler.sol
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity 0.8.10;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "./../core/component/Component.sol";
+import "../core/acl/IACLOracle.sol";
+
+abstract contract DAOPermissionHandler {
```
nivida: Let us ask some other folks as well to come up with a better name.. my eyes somehow burn..

packages/contracts/contracts/permissions/DAOPermissionHandler.sol
```diff
@@ -0,0 +1,23 @@
+/*
+ * SPDX-License-Identifier:    MIT
+ */
+
+pragma solidity 0.8.10;
```
nivida, line 5: Let us add here the natpec as on all other contracts as well :))

packages/contracts/contracts/tokens/GovernanceERC20.sol
```diff
@@ -4,15 +4,43 @@
...
+    function mint(address to, uint256 amount) external auth(TOKEN_MINTER_ROLE) {
+        _mint(to, amount);
+    }
 
-contract GovernanceERC20 is ERC20VotesUpgradeable {
-    function initialize(string calldata name, string calldata symbol) external initializer {
-        __ERC20_init(name, symbol);
-        __ERC20Permit_init(name);
+    // TODO: https://forum.openzeppelin.com/t/self-delegation-in-erc20votes/17501/12?u=novaknole
+    function delegates(address account) public view virtual override returns (address) {
```
nivida, line 42: noice! yup this requirement will definitely come up..

packages/contracts/contracts/tokens/GovernanceToken.sol
```diff
@@ -0,0 +1,26 @@
...
+
+pragma solidity 0.8.10;
+
+import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20VotesUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+import "../core/IDAO.sol";
+import "../permissions/DAOPermissionHandler.sol";
+
+abstract contract GovernanceToken is Initializable, DAOPermissionHandler, ERC20VotesUpgradeable {
```
nivida: what is the reason for this abstract contract? In this PR it isn't used anywhere.. 

Natspec is missing.

packages/contracts/contracts/tokens/MerkleDistributor.sol
```diff
@@ -0,0 +1,65 @@
...
+        uint256 mask = (1 << claimedBit_index);
+        return claimedWord & mask == mask;
+    }
+
+    function _setClaimed(uint256 _index) private {
+        uint256 claimedWord_index = _index / 256;
+        uint256 claimedBit_index = _index % 256;
+        claimedBitMap[claimedWord_index] = claimedBitMap[claimedWord_index] | (1 << claimedBit_index);
+    }
+}
```
nivida: EOF

packages/contracts/contracts/tokens/MerkleDistributor.sol
```diff
@@ -0,0 +1,65 @@
+/*
+ * SPDX-License-Identifier:    GPL-3.0
```
nivida, line 2: Just to double check.. you overtook the license from Uniswap right? not that we get into troubles :))

packages/contracts/contracts/tokens/MerkleMinter.sol
```diff
@@ -0,0 +1,51 @@
+/*
+ * SPDX-License-Identifier:    GPL-3.0
```
nivida: Can't we have this licensed under MIT as everything else is?

packages/contracts/test/utils/voting.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function createVoteData(supports: boolean, executeIfDecided: boolean) {
+    return defaultAbiCoder.encode([
+        'bool',
+        'bool'
+    ],[
+        supports,
+        executeIfDecided
+    ]);
+}
```
nivida: EOF

nivida (changes requested):

packages/contracts/contracts/factories/DAOFactory.sol
novaknole: Our GovernanceERC20 token has a `mint` function that is auth restricted. TokenFactory needs to have permissions so that it can allow itself to mint tokens right away on GovernanceERC20.

packages/contracts/contracts/factories/DAOFactory.sol
novaknole: no, one of them is bulking on `voting` , and the other one is on the `dao`.

packages/contracts/contracts/factories/TokenFactory.sol
novaknole: Not sure exactly what you mean.. elaborate ?

packages/contracts/contracts/permissions/DAOPermissionHandler.sol
novaknole: Yes... Much safer.

packages/contracts/contracts/tokens/GovernanceToken.sol
novaknole: No reason. Forgot to remove this.

packages/contracts/contracts/core/processes/Process.sol
novaknole: After talking and discussing, This check shouldn't be here.

novaknole: After talking and discussing, This check shouldn't be here.

packages/contracts/contracts/factories/DAOFactory.sol
novaknole: `allowedActions` struct removed. It makes sense though, but for MVP, no need.

packages/contracts/contracts/factories/DAOFactory.sol
novaknole: In the future,  We might decide to right away pass the merkle mint settings as well(like root hash and tree). In this way, I don't want to have everything in single struct.

packages/contracts/contracts/factories/DAOFactory.sol
novaknole: you mean, `bytes metadata` in the `DAOConfig` ? It can be doable with bytes32(ipfs hash) and we will definatelly look into this and change it later.

packages/contracts/contracts/factories/DAOFactory.sol
nivida: Yeah I just thought it would be worth to add such a comment to the code. But we can also keep it as it is.

packages/contracts/contracts/factories/DAOFactory.sol
nivida: ohh true.. thanks:)

packages/contracts/contracts/factories/TokenFactory.sol
nivida: You can use tokenConfig.addr directly instead of writing it first into a mem var.

PR #160, opened 2022-01-20
Feature: DAO 599 - Proposal skeleton page

WissenIstNacht:
This PR:
- Implements a page skeleton for the proposal page
- A hook to get mock data for that proposal page (to be used in ui-components until we get actual data)

PR #165, opened 2022-01-23
Integration/all proposals DAO-596 

sepehr2github:
- Implement the governance page skeleton
- change alert message prop on proposal card
- add getDateProposals to convert the proposal's timestamp to a proper format


sepehr2github:
@WissenIstNacht I think we should discuss about the combination of this one and #160

packages/web-app/src/utils/date.ts
```diff
@@ -24,3 +26,25 @@ export function getDateSections(): {
...
 }
+
+/**
+ * Note: this function will convert the proposal's timestamp to proper string to show
+ * as a alert message on proposals card
+ * @param type return the message if the type was pending or active
+ * @param timestamp proposal creat/end timestamp
+ * @returns a message with i18 translation as proposal ends alert
+ */
+export function getDateProposals(
```
WissenIstNacht: Architectural point: Ideally, functions should do as few things as possible (thereby increasing reusability and composability). A good rule of thumb to see whether that is the case is by comparing the function's name and what it returns. 

In your case, the function's name suggest that you want to get a Date's proposal (or a proposal's date). The output, however, is a formatted and translatated string containing the proposal's type AND date. In this case, a more accurate name would be `formatProposalDateAndType` (but even with this, it's not really clear how you format the date). This very long name indicates that the function probably does more than it should. Personally, I'd break it up into two functions: `getRemainingTime` and `translateProposalType`. Now you can get a remainig time even if it's not about a proposal and you can translate the proposal type even if there's not date involved.

packages/web-app/src/pages/governance.tsx
```diff
@@ -1,8 +1,96 @@
...
+  return (
+    <Container>
+      <PageWrapper
+        title={'Proposals'}
+        buttonLabel={'New Proposal'}
+        subtitle={'1 active Proposal'}
+        onClick={() => null}
+      >
+        <div className="mt-8 space-y-1.5">
+          <SearchInput placeholder="Type to filter" />
```
WissenIstNacht: I think we can remove the SearchInput until we actually have a search functionality in place.

packages/web-app/src/components/proposalList/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+              t('governance.proposals.states.defeated'),
+            ]}
+            {...(AlertMessage && {AlertMessage})}
+            buttonLabel={[
+              t('governance.proposals.buttons.read'),
+              t('governance.proposals.buttons.vote'),
+              t('governance.proposals.buttons.execute'),
+              t('governance.proposals.buttons.edit'),
+            ]}
+            key={index}
```
WissenIstNacht, line 56: Probably would be better to use a string containing the publisher and the date here (something like `{publisheraddress + startDate}`) instead of the `index`. This should be enough to uniquely identify each list element.

Edit: I guess at some point we'll have a proper proposal ID that will uniquely identify the proposal. You can therefore discard this comment if you want.

WissenIstNacht (changes requested):
Very nice PR, thanks a lot!

Left a few suggestions. Let me know what you think!

WissenIstNacht:
As for the combination with my PR. I suggest you merge yours first, then I will adapt mine.

packages/web-app/src/pages/governance.tsx
```diff
@@ -1,8 +1,96 @@
...
+    }
+    return 1;
+  }
+
+  // TODO: this filter / sort function should implement using graph queries
+  if (filterValue) {
+    displayedProposals = daoProposals.filter(
+      t => t.type === filterValue || filterValue === 'all'
+    );
+    displayedProposals.sort(sortProposals);
```
Fabricevladimir, line 34: Filtered proposals should go through whatever algorithm decides how many pages of proposals are to be displayed. Currently, the number of pages isn't being updated to correspond to the actual number of proposals to be displayed (`displayedProposals`). 

Also, we probably shouldn't show pagination if filtered proposals can fit on one page.

packages/web-app/src/utils/date.ts
Fabricevladimir: ^ Solid points

Fabricevladimir (changes requested):
Good work! Thanks

packages/web-app/src/pages/governance.tsx
```diff
@@ -1,8 +1,96 @@
...
+        </div>
+        <ListWrapper>
+          <ProposalList
+            proposals={displayedProposals.slice((page - 1) * 6, page * 6)}
+          />
+        </ListWrapper>
+        <PaginationWrapper>
+          {displayedProposals.length !== 0 && (
+            <Pagination
+              totalPages={Math.ceil(displayedProposals.length / 6) as number}
```
Fabricevladimir: The pagination buttons are being rendered as a list, but in the component we forgot to pass each button a key. 
Btw, might be a good idea to extract 6 as symbolic constant.

Fabricevladimir (changes requested):

packages/web-app/src/pages/governance.tsx
sepehr2github: It will update. May be I couldn't get the point

Fabricevladimir (approved):
Good to go for me

WissenIstNacht:
Very nice, thanks!

WissenIstNacht (approved):

PR #166, opened 2022-01-23
Chore/wallet tokens DAO-619 

sepehr2github:
- Fetch a list of wallet's tokens in the deposit flow
- Add no token available screen
- add storage icon

Note: I'm not sure about the performance. It looks good to me but let me know if loading tokens was slow for you! Thank you..


RakeshUP (approved):
Works perfectly. The infura request took about 250ms for me which I think is super fast! Surprising that it's so fast

Fabricevladimir (approved):
Good work! Once Ether is taken into account, this will be good to go

PR #168, opened 2022-01-24
Feature/Voting terminal DAO-588

RakeshUP:


WissenIstNacht:
Hey there, thanks for the PR.

I'm not sure how I feel about everything being hardcoded. I kind of see the point with the strings (especially if we're going to change the way translation works), but I'm not sure about the data. What's the issue from your POV in defining props at that point in time? Are you afraid they will change too much in the future (thus making your work now redundant)?

Aside from that, there's a bit of logic missing in the component:
- Both voting options are selectable (at the same time) (-> important)
- The typeahead search is not functional (-> less important)

Could you elaborate a bit on why you didn't do it here?

RakeshUP:
I am thinking this component will directly query the voting results from the subgraph. Data won't be passed in the props. 

As for the components, I was thinking of fixing the Checkbox(voting options) component as a separate chore. I haven't yet thought about how the component API will become. Again for the typeahead search, I thought it would be less work when we implement it during integration. @WissenIstNacht

Fabricevladimir:
What's the status of this? Is it to be reviewed or should we wait until the checkbox fix?

WissenIstNacht:
@Fabricevladimir I left a message on discord earlier. Essentially, I'm torn between approving the PR and requesting changes 😂

I get Rakesh's points, but I'm not sure I agree with (all of) them. I was waiting for another reviewer who can weigh in.

packages/ui-components/src/components/widgets/votingTerminal.tsx
```diff
@@ -0,0 +1,227 @@
...
+  },
+  {
+    wallet: '0xc54c...ee7a',
+    option: 'Yes',
+    votingPower: '13.333%',
+    tokenAmount: '250TN',
+  },
+];
+
+export const VotingTerminal: React.FC = () => {
```
Fabricevladimir, line 35: I agree with @WissenIstNacht on this one. I think this terminal will have to take in data as props. It will be too opinionated if it fetches its own data. This should be more of a presentational component, smart with what it can calculate using the props, but ultimately still presentational.

Fabricevladimir, line 35: I agree with @WissenIstNacht on this one. I think this terminal will have to take in data as props. It will be too opinionated if it fetches its own data. This should be more of a presentational component, smart with what it can calculate using the props, but ultimately still presentational.

Fabricevladimir: I agree with @WissenIstNacht on this one. I think this terminal will have to take in data as props. It will be too opinionated if it fetches its own data. This should be more of a presentational component, smart with what it can calculate using the props, but ultimately still presentational.

Fabricevladimir (changes requested):

RakeshUP:
Guys @Fabricevladimir @WissenIstNacht I've moved this component to web-app package and changed all strings to use translation. I think a useQuery hook can be directly used in this component to get data. 

This still needs some changes and refactoring, we could do it later while integration

packages/ui-components/src/components/checkbox/checkboxListItem.tsx
```diff
@@ -24,7 +24,7 @@ const Icons = {
 export type CheckboxListItemProps = {
   label: string;
   helptext: string;
-  multiSelect: boolean;
+  multiSelect?: boolean;
```
WissenIstNacht, line 27: Why do we keep this in here? Why not let a parent component (e.g., a checkbox group keep track of this).

To be more concrete: The issue here is that you can not have a single checkboxListItem with a constantly active state.

packages/web-app/src/locales/en/common.json
```diff
@@ -183,5 +183,33 @@
       "title": "Configure Governance",
       "description": "How are decisions made in the DAO? How long may a proposal be open for voting?"
     }
+  },
+  "votingTerminal": {
+    "title": "Voting",
+    "breakdown": "Breakdown",
+    "voters": "Voters",
+    "info": "Info",
```
WissenIstNacht, line 191: //Bonus task: could you try to organize some of those into existing categories? (maybe put the voting terminals object into governance, and add things single-word stuff like "yes", "no", "info", etc to the labels object?)

WissenIstNacht (commented):
Hey Rakesh, thakns a lot for the changes. I think this is ok.

Hey for some reason I got a type error on running the web-app. Something to do with indirect import of TextInputProps. Are you getting this as well...?

Let me know what you think on the Checkbox thingy and the I'll approve.

packages/ui-components/src/components/checkbox/checkboxListItem.tsx
RakeshUP: I guess @Fabricevladimir worked on a similar checkbox group component for one of the pages. 

I didn't do the checkbox group component in the first place because I didn't know what options might be needed. Will one option be checked by default/ multiple options be selected by default(for multiselect checkbox group), etc. Maybe we can factor out Fabrice's existing component

WissenIstNacht (approved):

packages/ui-components/src/components/checkbox/checkboxListItem.tsx
Fabricevladimir: These checkboxes are interesting. It would probably be best to have them wrapped in a parent, even if only in the web-app

Fabricevladimir (approved):

PR #170, opened 2022-01-24
Feature/Step component DAO-633

RakeshUP:
Most of the changes are from the yarn.lock file. 

The file fullScreenStepperDemo.tsx will be deleted before merging. It's just for the demo. You could visit http://localhost:3000/#/form to access the demo. Other related stuff like path & router changes in app.tsx will also be removed

The structure and props are still not satisfactory for me. Thought of changing things once it's actually used in other pages


packages/web-app/src/components/fullScreenStepper/fullScreenStepperDemo.tsx
```diff
@@ -0,0 +1,49 @@
...
+        >
+          <Step3 />
+        </Step>
+      </FullScreenStepper>
+      {/* Stuff like modal usage could go here */}
+    </FormProvider>
+  );
+};
+
+const Step3: React.FC = () => {
```
sepehr2github: It would be nice if you can choose a better name for this function. sth like moveToStep3 or moveToLastStep

packages/ui-components/src/components/icons/interface/index.ts
```diff
@@ -19,7 +19,6 @@ export {IconLinkExternal} from './icon_link_external';
 export {IconMenu} from './icon_menu';
 export {IconPerson} from './icon_person';
 export {IconRadioDefault} from './icon_radio_default';
-export {IconRadioSelected} from './icon_radio_selected';
```
sepehr2github, line 22: thanks for this fixing.

sepehr2github (approved):
Looks good to me! thanks alot. Just a minor suggestion

Fabricevladimir (approved):
Great work. I like this.

packages/web-app/src/components/fullScreenStepper/fullScreenStepperDemo.tsx
RakeshUP: This is just a demo, I'll remove this whole file before merging

PR #173, opened 2022-01-24
Fix/custom token button - DAO-665

Fabricevladimir:
- shows custom token without user having searched
- adds label when no token found in DAO treasury
- fixes warning not showing for withdraw amount greater than DAO token balance


RakeshUP (approved):

sepehr2github (approved):
Thanks, Fabrice. everything works totally fine.

PR #174, opened 2022-01-25
Feature/Card with Image+Label DAO-647 

RakeshUP:


packages/web-app/src/components/cardWithImage/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+};
+
+const CardWithImage: React.FC<CardWithImageProps> = ({
+  caption,
+  title,
+  subtitle,
+}) => {
+  return (
+    <Container>
+      {/* TODO: Change this url to an imported local image */}
```
WissenIstNacht: Can you expose the src url as well?

packages/web-app/src/components/cardWithImage/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+})``;
+
+const ImageContainer = styled.img.attrs({
+  className: 'object-cover mb-2 rounded-xl w-full',
+})<React.ImgHTMLAttributes<HTMLImageElement>>`
+  height: 160px;
+
+  @media (min-width: 768px) {
+    height: 240px;
+  }
```
WissenIstNacht: Was on the phone with Selim earlier and had a quick chat about this. The size of the image should be dictated by the parent and an image ratio, not by the image itself. For now I think most SVGs we'll pass will be more or less square, so I'd just go with full width and let it take the vertical space it needs.

WissenIstNacht (changes requested):
Thanks! :)

packages/web-app/src/components/cardWithImage/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export default CardWithImage;
+
+const Container = styled.div.attrs({
+  className: 'flex-1 p-3 rounded-xl bg-ui-0',
+})``;
+
+const ImageContainer = styled.img.attrs({
+  className: 'object-cover mb-2 rounded-xl w-full',
+})<React.ImgHTMLAttributes<HTMLImageElement>>`
```
Fabricevladimir: Should be able to remove the type there, I think.

Fabricevladimir: Should be able to remove the type there, I think.

Fabricevladimir: Should be able to remove the type there, I think.

Fabricevladimir (approved):
Thanks. Good work. Ready to merge once the other comments have been addressed.

WissenIstNacht (approved):

PR #175, opened 2022-01-26
Feature/Withdraw custom token - DAO-666

Fabricevladimir:
- Added add custom token button on withdraw token menu
- Added custom token flow in withdraw, complete with validation


Fabricevladimir:
One thing I've noticed is that with this PR, new deposit and new withdraw have become almost exactly the same. At some point it might be worth looking into combining the two into one process.

RakeshUP (approved):

packages/web-app/src/containers/depositForm/index.tsx
```diff
@@ -218,12 +218,13 @@ const DepositForm: React.FC = () => {
               validate: addressValidator,
             }}
             render={({
-              field: {name, onBlur, onChange, value},
+              field: {name, onBlur, onChange, value, ref},
```
WissenIstNacht, line 222: Could you briefly explain what passing the ref here does?

packages/web-app/src/containers/configureWithdraw/index.tsx
```diff
@@ -4,62 +4,168 @@ import {
...
+          setValue('tokenSymbol', data.symbol);
+          setValue('tokenImgUrl', data.imgUrl);
+        } else {
+          const {name, symbol} = await getTokenInfo(tokenAddress, provider);
+          setValue('tokenName', name);
+          setValue('tokenSymbol', symbol);
+        }
+        setValue('tokenBalance', balance);
+      } catch (error) {
+        /**
```
WissenIstNacht, line 77: Ahm, ok. But what happens if an API (call) fails?

WissenIstNacht (approved):
Hey thanks!

PR looks ok to me. (Please at least drop a console.error in the catch blocks, the empty catch blocks really keep me up at night haha)

packages/web-app/src/containers/depositForm/index.tsx
Fabricevladimir: I needed to be able to focus the custom token address field once the "Add custom token" button is clicked and the menu is closed.

packages/web-app/src/containers/configureWithdraw/index.tsx
Fabricevladimir: No really, this one is completely redundant to the point of needing to be removed even lol. The api functions catch their own errors and prints to the console. I don't even know if I like that.

RakeshUP: We could later on add some service like sentry here.

WissenIstNacht: Well then... 😅

PR #176, opened 2022-01-26
Removes processes

nivida:
After a longer discussion with the involved engineers did we decide to remove the process types at all. This mostly because of the planned marketplace.


packages/contracts/contracts/votings/simple/SimpleVoting.sol
```diff
@@ -47,29 +49,28 @@ contract SimpleVoting is VotingProcess, TimeHelpers {
     string private constant ERROR_CAN_NOT_FORWARD = "VOTING_CAN_NOT_FORWARD";
     string private constant ERROR_NO_VOTING_POWER = "VOTING_NO_VOTING_POWER";
 
-    event StartVote(uint256 indexed voteId, address indexed creator, string description);
+    event StartVote(uint256 indexed voteId, address indexed creator, bytes description);
     event CastVote(uint256 indexed voteId, address indexed voter, bool voterSupports, uint256 stake);
     event ExecuteVote(uint256 indexed voteId);
```
Rekard0, line 54: is ExecuteVote is used?

packages/contracts/contracts/votings/simple/SimpleVoting.sol
```diff
@@ -161,23 +153,25 @@ contract SimpleVoting is VotingProcess, TimeHelpers {
...
+    * @dev Method to execute a vote if allowed to
+    * @param _voteId The ID of the vote to execute
     */
-    function _execute(Execution memory execution) internal override {
-        require(_canExecute(execution.id), ERROR_CAN_NOT_EXECUTE);
+    function execute(uint256 _voteId) public {
+        require(_canExecute(_voteId), ERROR_CAN_NOT_EXECUTE);
 
-        dao.execute(execution.proposal.actions);
+        dao.execute(votes[_voteId].actions);
```
Rekard0, line 188: shouldn't we pass address(this) or something similar to the DAO, so the DAO knows these execution belong to this voting and not another?

nivida: in DAO we would have msg.sender we can use.. but yup we should emit it in DAO :)

PR #177, opened 2022-01-26
Fix/ListItem link - DAO-609

Fabricevladimir:
- makes icon clickable
- fix lint warning and error

### DIFF
--- a/packages/ui-components/src/components/listItem/link.tsx
+++ b/packages/ui-components/src/components/listItem/link.tsx
@@ -21,17 +21,16 @@ export const ListItemLink: React.FC<ListItemLinkProps> = ({
 }) => {
   return (
     <Container>
-      <TitleContainer>
-        <Link
-          rel="noopener noreferrer"
-          {...props}
-          {...(external ? {target: '_blank'} : {})}
-          data-testid="listItem-link"
-        >
-          {props.label ? props.label : props.href}
-        </Link>
-        <IconLinkExternal className="w-1.5 h-1.5" />
-      </TitleContainer>
+      <Link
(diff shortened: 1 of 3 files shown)


RakeshUP (approved):

packages/ui-components/src/components/listItem/link.tsx
```diff
@@ -21,17 +21,16 @@ export const ListItemLink: React.FC<ListItemLinkProps> = ({
...
-        <IconLinkExternal className="w-1.5 h-1.5" />
-      </TitleContainer>
+      <Link
+        rel="noopener noreferrer"
+        {...props}
+        {...(external ? {target: '_blank'} : {})}
+        data-testid="listItem-link"
+      >
+        <Title>{props.label ? props.label : props.href}</Title>
+        <Icon />
```
WissenIstNacht, line 31: Shouldn't this be dependenent on the `external` prop?

WissenIstNacht (approved):
Thanks :)

packages/ui-components/src/components/listItem/link.tsx
Fabricevladimir: Do you mean the Icon? I think that is built in.

WissenIstNacht: I see. But then the external prop seems redundant. Don't worry about it for now, just was just curious :)

Fabricevladimir: I'll revisit it during refactor. The external is for how the link is opened, whether in the same tab or in a new one. I wanted to set new tab by default, that's why I added it.

PR #179, opened 2022-01-26
scroll to top added & add address link to card proposal DAO-595 DAO-644 

sepehr2github:
- add scroll to top on change page event
- add a linked address to card proposals

### DIFF
--- a/packages/ui-components/src/components/link/link.tsx
+++ b/packages/ui-components/src/components/link/link.tsx
@@ -13,6 +13,7 @@ export type LinkProps = React.AnchorHTMLAttributes<HTMLAnchorElement> & {
   iconLeft?: React.FunctionComponentElement<IconType>;
   /** optional label for the link, defaults to the href if value not provided */
   label?: string;
+  size?: 'small' | 'default' | 'large';
 };
 
 /** Default link component */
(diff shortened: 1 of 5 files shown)


RakeshUP (approved):

packages/ui-components/src/components/link/link.tsx
```diff
@@ -13,6 +13,7 @@ export type LinkProps = React.AnchorHTMLAttributes<HTMLAnchorElement> & {
   iconLeft?: React.FunctionComponentElement<IconType>;
   /** optional label for the link, defaults to the href if value not provided */
   label?: string;
+  size?: 'small' | 'default' | 'large';
```
WissenIstNacht, line 16: Oh, I didn't know they had different sizes. Designs only have one, right?

WissenIstNacht (approved):
Thanks Sepehr!

Not sure about the link size, the rest looks good though!

packages/ui-components/src/components/link/link.tsx
sepehr2github: So In designs, there is no size variant but I had to add this for supporting different font sizes. The best practice is to change the font size by className. We will fix it later!

PR #183, opened 2022-01-28
Chore/router suspense fix

RakeshUP:
This fixes the intermediate empty content state when switching tabs in the app for the first time. This is more pronounced in the hosted environment. Check https://zaragoza-staging.aragon.org/#/. The navbar should not disappear when switching tabs and this PR fixes that

Also changing the way 404 works. Instead of navigating to the not found page, I am rendering that component keeping the URL unchanged, so the user knows the actual URL that resulted in the 404

### DIFF
--- a/packages/web-app/src/app.tsx
+++ b/packages/web-app/src/app.tsx
@@ -1,8 +1,8 @@
-import React, {useEffect, lazy, Suspense} from 'react';
+import React, {useEffect, lazy, Suspense, ReactElement} from 'react';
 
 // FIXME: Change route to ApmRoute once package has been updated to be
 // compatible with react-router-dom v6
-import {Navigate, Routes, Route, useLocation, Outlet} from 'react-router-dom';
+import {Routes, Route, useLocation, Outlet} from 'react-router-dom';
 
 import Footer from 'containers/footer';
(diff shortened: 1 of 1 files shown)


Fabricevladimir:
This issue was introduced by the process routes New Deposit and New Withdraw being on a different routing level as the other pages because we hadn't refactored the navbar to show differently on the process pages. I think the better solution is to bring these routes back into the same level and have a single navbar that is outside the nested routes. I've approved the PR (twice for whatever reason 😄), but I think this refactor should be done at some point.

**UPDATE**: Also, be aware that not redirecting to the notfound page while is better solution, will potentially cause breadcrumbs to nonexistent pages to be generated. For example, assuming `/governance/abc` is not a valid route, navigating to `/governance/abc/def` will render both `governance` and `abc` as breadcrumbs despite `abc` not being a valid route. It's not a big deal, but it can become annoying for the user.

Fabricevladimir (approved):
Thanks

Fabricevladimir (approved):
Thanks

RakeshUP:
Closing this PR. The method Fabrice mentioned would solve this same issue and some more!

PR #186, opened 2022-01-28
Chore - transfer refactor

WissenIstNacht:
This PR:

- refactors the new-deposit page using the full-screen stepper
- refactors the new-transfer page using the full-screen stepper
- fixes a bug that would not show tokens held by user when changing chain
- fixes the copy on the second step of new-deposit

### DIFF
--- a/packages/web-app/src/hooks/useWalletTokens.tsx
+++ b/packages/web-app/src/hooks/useWalletTokens.tsx
@@ -1,0 +1,49 @@
+import {useState, useEffect} from 'react';
+import {constants} from 'ethers';
+import {formatUnits} from 'ethers/lib/utils';
+
+import {isETH, fetchBalance} from 'utils/tokens';
+import {useWallet} from 'context/augmentedWallet';
+import {useWalletProps} from 'containers/walletMenu';
+import {TokenBalance} from 'utils/types';
+
+// TODO return as HookData
+export function useWalletTokens() {
+  const {account, chainId, provider, getTokenList, balance}: useWalletProps =
+    useWallet();
+  const [walletTokens, setWalletTokens] = useState<TokenBalance[]>([]);
+
+  // fetch tokens and corresponding balance on wallet
+  useEffect(() => {
+    async function fetchWalletTokens() {
+      if (account === null) {
+        setWalletTokens([]);
+        return;
+      }
+
+      const tokenList = await getTokenList();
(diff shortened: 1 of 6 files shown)


RakeshUP (approved):

RakeshUP:
The color of the 'Optional' badge for the reference label in new deposit is different. Could you check that?

packages/web-app/src/hooks/useWalletTokens.tsx
```diff
@@ -0,0 +1,49 @@
+import {useState, useEffect} from 'react';
```
Fabricevladimir, line 1: Thanks for moving this out.

packages/web-app/src/pages/newDeposit.tsx
```diff
@@ -162,107 +85,34 @@ const NewDeposit: React.FC = () => {
...
-          </FormLayout>
-        </FormProvider>
-        <TokenMenu
-          onTokenSelect={handleTokenSelect}
-          tokenBalances={walletTokens}
-        />
-      </Layout>
-    </>
+    <FormProvider {...formMethods}>
+      <FullScreenStepper
```
Fabricevladimir, line 89: This is simply beautiful. We need to take into account the button labels though at some point..

packages/web-app/src/pages/newWithdraw.tsx
```diff
@@ -1,55 +1,45 @@
...
-};
+export type TransferFormData = TransferData &
+  TokenFormData & {
+    isCustomToken: boolean;
+    type: TransferTypes;
+  };
 
-const TOTAL_STEPS = Object.keys(steps).length;
+export type WithdrawFormData = TransferFormData & {
+  // NOTE: Is this really just a withdrawl thing?
```
Fabricevladimir, line 40: Good catch. This one is on me, I don't believe we're using that form property anymore, so it can be removed.

Fabricevladimir (approved):
Thanks, quite satisfying!

packages/web-app/src/pages/newDeposit.tsx
WissenIstNacht: Do you mean the "New Transfer" one? I wanted to do that but forgot. I'll add a fix for it!

WissenIstNacht:
@RakeshUP The Badge _itself_ is defined to use `ui-50` for some reason. I'm changing it to `100` :)

PR #187, opened 2022-01-31
Feature/Assemble page (Select blockchain) - DAO-634

Fabricevladimir:
- added select blockchain view
- wired to form


RakeshUP (approved):

RakeshUP:
Could you make the FullScreenStepper take the total steps of the wizard as a prop and currentStep as the child option prop? The useStepper would still have children.length as the value so that the next, prev, etc work properly. Just the display values in wizard would be coming in as props.

packages/web-app/src/locales/en/common.json
```diff
@@ -38,6 +38,15 @@
     "submitDeposit": "Submit Deposit",
     "resources": "Resources",
     "submitWithdraw": "Publish Proposal",
+    "sortBy": "Sort by",
+    "mainNet": "Main net",
```
WissenIstNacht: I would either go with Mainnet \(Testnet, Whatevernet\) or Main Network \(Test Network, Whatever Network\). I'd suggest the first, to stay consistent with our other products and the definitions we have in useWallet.

Just saw that you copied that from the desings. My bad. But then "Main Net" :)

packages/web-app/src/utils/library.ts
```diff
@@ -28,3 +28,13 @@ export async function handleClipboardActions(
     onChange(textFromClipboard);
   }
 }
+
+/** (Temporary)
+ * Check if network is supported test net
+ * @param chainId network chain id
+ * @returns whether network is one of the supported testnets
+ */
+export function isTestNet(chainId: number) {
```
WissenIstNacht: https://github.com/aragon/use-wallet/blob/master/src/chains.ts

Every chain listed has a boolean field called `testnet` :)

packages/web-app/src/containers/selectChainForm/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    tag: i18n.t('labels.mostPopular'),
+    title: i18n.t('labels.popularity'),
+  },
+  security: {
+    tag: i18n.t('labels.safest'),
+    title: i18n.t('labels.security'),
+  },
+};
+
+// TODO: Find proper place. Configs maybe?
```
WissenIstNacht: yeah, or constants.

packages/web-app/src/containers/selectChainForm/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+          <Label label={t('labels.sortBy')} />
+          {/* TODO: replace with proper dropdown */}
+          <Popover
+            side="bottom"
+            open={isOpen}
+            align="start"
+            width={264}
+            onOpenChange={value => setIsOpen(value)}
+            content={
+              <DropdownContent>
```
WissenIstNacht, line 82: Opening the dropdown always focuses the first item (I would expect it to focus either no item, or the one currently selected)

If it's a hassle to change leave it as is and drop a TODO for later. I'm not even sure what Selim intended here in the first place.

packages/web-app/src/containers/selectChainForm/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+  polygon: {
+    domain: 'L2 Blockchain',
+    logo: 'https://assets.coingecko.com/coins/images/4713/large/matic-token-icon.png?1624446912',
+  },
+};
+
+const supportedNetworks = {
+  main: {
+    cost: [
+      {id: 137, name: 'Polygon', ...chainMetadata.polygon},
```
WissenIstNacht: You could have saved yourself some typing by adding the id and name to the metadata as well and then simply list the metadata objects :yum:

WissenIstNacht (changes requested):
Hey there, thanks for the PR, looks great.

I left a few minor comments thorughout the code. nothing importnat.

There is one thing however: The stepper is not adapted to mobile. It shouldn't take too much to change the layout for mobile.

packages/web-app/src/utils/library.ts
Fabricevladimir: Sweet

packages/web-app/src/containers/selectChainForm/index.tsx
Fabricevladimir: That's true yes. Wasn't sure where and how these guys were going to live

packages/web-app/src/containers/selectChainForm/index.tsx
Fabricevladimir: Isn't first item focus default for dropdowns? But, yeah, this is residue from using Popover instead of Dropdown. There is a TODO for it.

sepehr2github (approved):
Awsome Fabrice. One more step to finish creation flow 👍 . 
Just a minor thing:
Screen Shot 2022-02-07 at 1 01 31 AM

What happened to the internal border-radius?

And a simple question: Why do we have a network switch at all? We can direct creation flow based on the active wallet chain? because It's sort of weird that I can create dao in the test net while my active wallet active chain is mainnet!

Fabricevladimir:
@sepehr2github About the borderRadius, I finally added the proper 10px border radius for the default button in the ui-components, but the key hadn't been added to the web-app TW config. @RakeshUP has added it though in a different PR so I'll disregard it here just to minimize conflicts.

As for the network type, I assume the network type you choose here is takes precedence over the wallet chain. So at the end of the flow, when you want to deploy your DAO, at some point we'll have to point out to the user that the desired network and the one their wallet is currently connected to aren't the same...

WissenIstNacht (approved):

PR #189, opened 2022-02-01
Fix/Searching token menu

Fabricevladimir:
- sets token name and symbol to optional

### DIFF
--- a/packages/web-app/src/containers/tokenMenu/index.tsx
+++ b/packages/web-app/src/containers/tokenMenu/index.tsx
@@ -53,7 +53,7 @@ const TokenMenu: React.FC<TokenMenuProps> = ({
     (token: BaseTokenInfo) => {
       if (searchValue !== '') {
         const re = new RegExp(searchValue, 'i');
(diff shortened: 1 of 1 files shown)


githubdoramon (approved):

RakeshUP (approved):

WissenIstNacht:
Hey apparently I haven't done a review on this file for a while. Two things I'd like to point out for the future:
1. Generally, Regular expressions are (more) expensive (than simple string manipulations). In this case, I'm pretty sure the simple include function provided by the standard library would be faster. It might not be significant, but still.
2. Please don't declare (complex) components inside other components (e.g. `renderTokens`). This defeats the whole purpose of react(-components) as hinders it react to track it across updates. This can have serious performance impacts.  And if you do, please call them using JSX and not as actual functions. :D

Fabricevladimir:
@WissenIstNacht You are indeed correct on both counts. I'll add this to the notion or push up a chore refactor once I'm done with my current tasks.

PR #190, opened 2022-02-01
Feature/Single Proposal assemble page DAO-598

RakeshUP:


sepehr2github (commented):
Thank you Rakash. I want to mention sth but I don't know if it's on your task area or not!
So when I clicked on the "read proposal" button to see proposal details the app will redirect me to a route like "/governance/proposals/400" but It's wrong because there is no "/governance/proposals" page and if I click on "proposals" link in breadcrumb. I will face the 404 page, which is sort of wired :/

packages/web-app/src/pages/proposals.tsx
```diff
@@ -1,43 +1,197 @@
...
+})``;
+
+const SummaryText = styled.p.attrs({
+  className: 'text-lg text-ui-600 mt-3',
+})``;
+
+const ButtonContainer = styled.div.attrs({
+  className: 'mt-3',
+})`
+  button {
```
sepehr2github, line 172: It would be nice if you can implement this CSS style using tailwind classes

packages/web-app/src/pages/proposals.tsx
```diff
@@ -1,43 +1,197 @@
...
+    @media (max-width: 767px) {
```
sepehr2github: It would be nice if you can implement this CSS style using tailwind classes

sepehr2github (commented):
It seems to work well! 👍 a suggestion and a question from my side.

packages/web-app/src/pages/proposals.tsx
```diff
@@ -1,43 +1,186 @@
...
+const stepDataRunning: ProgressStatusProps[] = [
+  publishedDone,
+  {
+    label: 'Running',
+    date: '2021/11/16 4:30 PM UTC+2',
+    mode: 'active',
+  },
+];
 
 const Proposal: React.FC = () => {
```
Fabricevladimir: I think this might have been a typo, do you mind changing the name of the page to "Proposal" as well instead of "Proposals". Makes it sound like a list of proposals. The same thing happened with the path as well. It's just a minor suggestion.

Fabricevladimir (approved):
Good work! Thanks

RakeshUP:
@sepehr2github The breadcrumb issue is fixed and the other comment is addressed too!

sepehr2github (approved):
Thank you for addressing all comments! Good To go 👍

PR #193, opened 2022-02-01
Feature/Setup community with existing token - DAO-653

Fabricevladimir:
Adds existing token components (no contract address logic)


RakeshUP (approved):

sepehr2github (approved):
Look nice 👍 
I think Rakash added a description list Component to the UI library. U can use it here but because this one was before rakash pr I'm gonna approve it.

PR #194, opened 2022-02-02
Chore/Update CheckboxListItem & TextInput - DAO-673 

Fabricevladimir:
CheckboxListItem
- removes state to make it a fully controlled component (@RakeshUP you were correct in having it as so in the first place 😅)
- removes removes ability to click on disabled component
- changes cursor to normal cursor when component is disabled

Adds border to disabled TextInput component and key for medium sized button borderRadius


WissenIstNacht (approved):
Thanks for this, was facing issues with it on my PR as well.

RakeshUP (approved):

PR #195, opened 2022-02-02
Feature / DAO-632

Rekard0:
- add module for contracts in templates
- add readme
- setup unit testing env via docker
- refactor mapping and schema to adapt with new contracts


packages/subgraph/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM ubuntu:20.04
```
nivida: Could you use a docker compose for this. I don't think we have to create a docker image especially for this. Feel free to copy the one from Govern I created over as I mentioned once.

package.json
```diff
@@ -1,15 +1,21 @@
...
+  "private": true,
+  "workspaces": [
+    "packages/*"
+  ],
+  "devDependencies": {
+    "lerna": "^3.22.1"
+  },
+  "scripts": {
+    "build:contracts": "lerna run --scope=@aragon/zaragoza-contracts --stream compile",
+    "manifest:subgraph:arbitrum:rinkeby": "lerna run --scope=@aragon/zaragoza-subgraph --stream manifest-arbitrum-rinkeby",
```
nivida: Most of those commands should be in the subgraph package.

packages/subgraph/README.md
```diff
@@ -0,0 +1,103 @@
+# Zaragoza Subgraph
+
+## Quick Start
+
+Ensure the monorepo’s dependencies are installed:
+
+```console
+cd ../.. && yarn && cd packages/subgraph
```
nivida: it doesn't depend if I'm correct if you run yarn in packages/subgraph or on the root folder cause we have a yarn workspace.

packages/subgraph/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+## Build Contracts
+
+make sure the Contracts package at /root/packages/contracts are compiled:
+
+```console
+cd ../.. && yarn build:contracts && cd packages/subgraph
```
nivida: This could be included directly in the commands of this package.

packages/subgraph/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+## Build & deploy the subgraph
+
+You can build and deploy the subgraph using a single `yarn deploy-<network>` command:
+
+```console
+GRAPH_KEY=<GRAPH_KEY> yarn deploy-<DESIRED_NETWORK> <THEGRAPH_USERNAME> <SUBGRAPH_NAME> <DESIRED_NETWORK>
```
nivida: Probably worth to have all this in a .env file and using it in a bash script.

packages/subgraph/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Test the subgraph
+
+build the the contracts as explained above.
+build the subgraph as explained above
+
+run tests via docker:
+
+```console
+yarn test:docker
```
nivida: "yarn test"  should be enough :)

packages/subgraph/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+however due to matchstick not supporting monorepo as of now, using docker for testing might not be practical during developing process, a work arround could be copying the entire /node_modules from stand alone subgraph project and pasting it in packages/subgraph only for development purposes.
+
+### Deploy the subgraph locally
```
nivida, line 61: All this would be obsolet if you use the Govern scripts etc. as mentioned once.

packages/subgraph/manifest/subgraph.placeholder.yaml
```diff
@@ -58,118 +54,31 @@ templates:
...
-          handler: handleCastVote
-        - event: ChangeMinQuorum(uint64)
-          handler: handleChangeMinQuorum
-        - event: ChangeSupportRequired(uint64)
-          handler: handleChangeSupportRequired
-        - event: ExecuteVote(indexed uint256)
-          handler: handleExecuteVote
-        - event: StartVote(indexed uint256,indexed address,string)
-          handler: handleStartVote
+  # # Voting (package process)
```
nivida, line 57: Not required to commit such out commented sections.

package.json
```diff
@@ -1,15 +1,21 @@
...
+  "description": "Human-centered Dao infrastructure",
+  "repository": "https://github.com/aragon/Zaragoza",
+  "author": "Aragon Association",
+  "license": "MIT",
+  "private": true,
+  "workspaces": [
+    "packages/*"
+  ],
+  "devDependencies": {
+    "lerna": "^3.22.1"
```
nivida, line 23: I'm wondering how lerna found into this repo. If you have the could you investigate if we really need it? Currently, I think we don't.

packages/subgraph/package.json
```diff
@@ -5,21 +5,21 @@
...
     "deploy-local": "LOCAL=true scripts/deploy.sh aragon zaragoza arbitrum-rinkeby",
     "create-local": "graph create aragon/aragon-zaragoza-arbitrum-rinkeby --node http://localhost:8020",
-    "test": "rm -r tests/.bin; graph test",
-    "coverage": "graph test -c",
-    "manifest-arbitrum-rinkeby": "scripts/build-manifest.sh arbitrum-rinkeby"
+    "coverage": "graph test -c"
   },
   "dependencies": {
-    "@graphprotocol/graph-cli": "^0.24.1",
+    "@graphprotocol/graph-cli": "^0.25.1",
```
nivida, line 20: Is the CLI not a devDependency?

packages/subgraph/schema.graphql
```diff
@@ -80,17 +50,15 @@ type VaultDeposit @entity {
...
 type Dao @entity {
-  id: ID!
+  id: ID! # use address as id
   name: String!
   creator: Bytes!
   metadata: String!
   token: Bytes!
-  processes: [ProcessDao!]! @derivedFrom(field: "dao")
-  proposals: [Proposal!]! @derivedFrom(field: "dao")
+  actions: [Action!]! @derivedFrom(field: "dao")
```
nivida, line 59: shouldn't this be executions?

packages/subgraph/schema.graphql
```diff
@@ -80,17 +50,15 @@ type VaultDeposit @entity {
...
   ethDeposits: [VaultEthDeposit!]! @derivedFrom(field: "dao")
   tokenDeposits: [VaultDeposit!]! @derivedFrom(field: "dao")
```
nivida, line 61: can't we merge eth- and tokenDeposits just to deposits?

packages/subgraph/schema.graphql
```diff
@@ -80,17 +50,15 @@ type VaultDeposit @entity {
...
   withdraws: [VaultWithdraw!]! @derivedFrom(field: "dao")
-  executions: [Execution!]! @derivedFrom(field: "dao")
+  erc20Balances: [ERC20Balance!] @derivedFrom(field: "dao")
```
nivida, line 63: I think this should be balances and include eth and tokens.

packages/subgraph/src/dao.ts
```diff
@@ -1,61 +1,21 @@
...
   SetMetadata,
-  ProcessAdded,
-  ProcessRemoved,
   Executed,
   Deposited,
   ETHDeposited,
   Withdrawn
 } from '../generated/templates/DAO/DAO';
-import {SimpleVoting} from '../generated/templates';
+// import {SimpleVoting} from '../generated/templates';
```
nivida: Can get removed at all if not used.

packages/subgraph/src/dao.ts
```diff
@@ -1,61 +1,21 @@
...
   Dao,
-  Process,
   VaultEthDeposit,
   VaultDeposit,
-  VaultWithdraw,
-  ProcessDao
+  VaultWithdraw
 } from '../generated/schema';
 import {DataSourceContext, store} from '@graphprotocol/graph-ts';
 import {log} from 'matchstick-as/assembly/index';
```
nivida, line 17: Shouldn't we replace process added with the ACL events as discussed yesterday?

packages/subgraph/src/dao.ts
```diff
@@ -74,55 +34,52 @@ export function handleDeposited(event: Deposited): void {
...
-    entity.reason = event.params._reference;
-    entity.save();
-  }
+  let entity = new VaultWithdraw(id);
+
+  entity.dao = daoId;
+  entity.token = event.params.token;
+  entity.to = event.params.to;
+  entity.amount = event.params.amount;
+  entity.reason = event.params._reference;
```
nivida: I think we can call it "reference" cause it is a known wording from finance (payment reference).

packages/subgraph/src/dao.ts
```diff
@@ -74,55 +34,52 @@ export function handleDeposited(event: Deposited): void {
...
-    entity.dao = daoId;
-    entity.sender = event.params.sender;
-    entity.amount = event.params.amount;
-    entity.save();
-  }
+  let entity = new VaultEthDeposit(id);
+
+  entity.dao = daoId;
+  entity.sender = event.params.sender;
+  entity.amount = event.params.amount;
```
nivida, line 63: Wdyt? should we add in the contracts the possibility to add a payment reference as well?

nivida (changes requested):
I've added a few comments :)

btw. deploy.sh got checked in and is an empty file.

packages/subgraph/schema.graphql
Rekard0: yes, it will be in the next PR feature/DAO-677

packages/subgraph/src/dao.ts
Rekard0: amended

packages/subgraph/src/dao.ts
Rekard0: yes, it will be in PR feature/DAO-675

packages/subgraph/src/dao.ts
Rekard0: yes, agreed

packages/subgraph/src/dao.ts
Rekard0: do u mean creating function receiveEth(reference) in contract so we can pass reference to event, and also in the receive() passing 'no reference' as reference? 
can we override receive() to have a param?

packages/subgraph/README.md
Rekard0: yes, amended

packages/subgraph/README.md
Rekard0: yes, thanks, amended.

package.json
Rekard0: amended

packages/subgraph/README.md
Rekard0: yes, we can use .env instead. amended

packages/subgraph/src/dao.ts
nivida: I was meaning that we can add a "reference" argument to the deposit method we already have. :-) 

We can't override receive() cause it is a std fallback if someone is sending ETH to this contract.

nivida: But let us create out of this a task and we can if we want change it later on:))

packages/subgraph/README.md
Rekard0: yes, done

packages/subgraph/README.md
Rekard0: yes, now all can be done via docker compose

packages/subgraph/manifest/subgraph.placeholder.yaml
Rekard0: yes removed

packages/subgraph/package.json
Rekard0: i was following the example of subgraph-demo, but yes, it can be devDependency. amended

packages/subgraph/schema.graphql
Rekard0: yes, it will be in PR feature/dao-676

package.json
Rekard0: yes it is needed for running npm scripts such as: lerna run ..etc
however it is not needed if lerna is installed globally.

should we remove it? and assume lerna is installed globally?

packages/subgraph/schema.graphql
Rekard0: ok, i will consider that in its PR

packages/subgraph/Dockerfile
Rekard0: done

package.json
nivida: So let us keep it for now 👍

nivida:
Thanks for all the changes. As stated by you do further improvements come with other smaller PRs. Merging it now.

PR #196, opened 2022-02-02
Feature/Transfer Modal DAO-660

RakeshUP:
- ModalBottomSheetSwitcher handles switching between modal and bottom sheet 
- All other modals are refactored in this PR
- Few styling and naming changes

Didn't do: 
- Fix the styling of bottomsheet in mobile navbar(bgColor and stuff like that)
- DAO Switcher is still not implemented for mobile


packages/web-app/src/containers/transactionModal/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+      title={title}
+      subtitle={subtitle}
+    >
+      <GasCostTableContainer>
+        <GasCostEthContainer>
+          <VStack>
+            <Label>{t('TransactionModal.estimatedFees')}</Label>
+            <p className="text-sm text-ui-500">{`${t(
+              'TransactionModal.synced'
+            )} 30 ${t('labels.sec')} ${t('labels.ago')}`}</p>
```
Fabricevladimir: You can have this as an interpolated string and pass in the seconds like so:  "Synced  {{seconds}} ago"

packages/web-app/src/containers/transactionModal/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+};
+
+const icons = {
+  [TransactionState.WAITING]: undefined,
+  [TransactionState.LOADING]: <Spinner size="xs" color="white" />,
+  [TransactionState.SUCCESS]: undefined,
+  [TransactionState.ERROR]: <IconReload />,
+};
+
+const TransactionModal: React.FC<TransactionModalProps> = ({
```
Fabricevladimir, line 41: Should this be in web-app or ui-components?

packages/web-app/src/hooks/useMediaQuery.tsx
```diff
@@ -0,0 +1,36 @@
+import {useEffect, useState} from 'react';
+
+function useMediaQuery(query: string): boolean {
```
Fabricevladimir: Thanks for this. The initial plan was to use TW instead of explicit media queries, but there are quite a few cases where that's just not possible. For convenience's sake though since we already know which screens we'll support, it would be nice if we could just get the screen from the hook. Something like:
`const {isTablet} = useScreenSizes(); // or: isMobile, isDesktop and so on` 
What do you think?

packages/web-app/tailwind.config.js
```diff
@@ -96,6 +96,9 @@ module.exports = {
         16: '128px',
         25: '200px',
       },
+      borderRadius: {
```
Fabricevladimir: Thanks, I forgot to update this in the PR that added it in ui-components. Perhaps one TW config would be a better approach...

packages/web-app/src/containers/transactionModal/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+  title: string;
+  footerButtonLabel: string;
+  state: TransactionState;
+  callback: () => void;
+  subtitle?: string;
+  successLabel?: string;
+  errorLabel?: string;
+  approveStepNeeded?: boolean;
+  approveCallback?: () => void;
+};
```
Fabricevladimir, line 32: Missing cost and gas fee props I think

packages/web-app/src/containers/transactionModal/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+          </VStack>
+          <VStack>
+            <StrongText>{'0.001ETH'}</StrongText>
+            <p className="text-sm text-right text-ui-500">{'127gwei'}</p>
+          </VStack>
+        </GasCostEthContainer>
+
+        <GasCostUSDContainer>
+          <Label>{t('TransactionModal.totalCost')}</Label>
+          <StrongText>{'$16.28'}</StrongText>
```
Fabricevladimir, line 93: The total cost should be calculated dynamically.

packages/web-app/src/containers/transactionModal/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+        <GasCostEthContainer>
+          <VStack>
+            <Label>{t('TransactionModal.estimatedFees')}</Label>
+            <p className="text-sm text-ui-500">{`${t(
+              'TransactionModal.synced'
+            )} 30 ${t('labels.sec')} ${t('labels.ago')}`}</p>
+          </VStack>
+          <VStack>
+            <StrongText>{'0.001ETH'}</StrongText>
+            <p className="text-sm text-right text-ui-500">{'127gwei'}</p>
```
Fabricevladimir, line 87: This should probably come as props as well, or at the very least derived from the ETH value. Ethers.js has a utility function for that I believe...

Fabricevladimir (changes requested):
Massive work. Thanks, the modal system needed an overhaul. Left some comments.

packages/web-app/tailwind.config.js
RakeshUP: The bigger issue is that styles are not packaged with the ui-components package. The package should contain its own styles and should be able to work independently, but that's not happening now

packages/web-app/src/containers/transactionModal/index.tsx
RakeshUP: Estimating gas fee, transaction cost in native token, and cost in USD are to be done in a separate task. I've hardcoded those values for now. I didn't add the props, because was not sure what props are needed. I think it needs contract address and function name too for estimating the prices

packages/web-app/src/containers/transactionModal/index.tsx
RakeshUP: Every value is hardcoded for now, till it's fixed in other tickets DAO-661 and DAO-662

packages/web-app/src/containers/transactionModal/index.tsx
RakeshUP: I am putting components with translated strings in the web-app package for now

packages/web-app/src/containers/transactionModal/index.tsx
RakeshUP: Cool, I am taking a look at that

packages/web-app/src/hooks/useMediaQuery.tsx
RakeshUP: A much better way to do the way you think is to use have a listener on document.innerWidth and returning issTablet, isMobile, etc values based on that. Because to export the same structure we would need three matchMedia listeners, which would result in more re-renders. 

We can change this in the future to use a listener on document.innerWidth

packages/web-app/src/containers/transactionModal/index.tsx
RakeshUP: Changed it

packages/web-app/src/containers/transactionModal/index.tsx
Fabricevladimir: Fair enough

packages/web-app/tailwind.config.js
Fabricevladimir: Yup, you're right. This is something we definitely need to revisit.

packages/web-app/src/hooks/useMediaQuery.tsx
Fabricevladimir: Ah, yes. I would like this yes. innerWidth would be compared to our screen sizes and update memoized boolean values. If it takes too much time, we can leave it for later, but it would really be helpful in terms of dev experience.

sepehr2github (approved):
Massive Work Rakash! Well Done! honestly the reason I had to make bottom sheet to web app is that the storybook couldn't support framer motion and we should find a solution later to move the whole modal to UI components. But anyway Thank you!

packages/web-app/src/hooks/useMediaQuery.tsx
RakeshUP: Changed the api to the one you suggested

Fabricevladimir (approved):

PR #198, opened 2022-02-02
chore/ DAO-655 & fix single image input default limitations

sepehr2github:
- percentage state added to numeric input
- full-width style added to default mode
- remove input single image default limitations


packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -27,13 +29,17 @@ export const NumberInput: React.FC<NumberInputProps> = ({
...
-        placeholder={'0'}
-      />
+      <InputWrapper>
+        <StyledNumberInput
+          {...props}
+          {...(percentage && {percentage, min: 0, max: 100})}
+          ref={inputRef}
+          disabled={disabled}
+          type={'number'}
+          placeholder={'0'}
```
Fabricevladimir: Are we against a user input as placeholder?

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -27,13 +29,17 @@ export const NumberInput: React.FC<NumberInputProps> = ({
...
+        />
+        {percentage && <Percent disabled={disabled}>%</Percent>}
```
Fabricevladimir: There is a weird state when the value is undefined and the placeholder shows along with the %; the placeholder text color and % text color are different.

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -27,13 +29,17 @@ export const NumberInput: React.FC<NumberInputProps> = ({
         disabled={disabled}
         onClick={() => inputRef.current?.stepDown()}
```
Fabricevladimir: It doesn't seem as if stepDown and stepUp trigger an onChange event on the input by default, so if this component is controlled, when the user clicks on either button, the controller value will not be updated.

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -27,13 +29,17 @@ export const NumberInput: React.FC<NumberInputProps> = ({
...
-        {...props}
-        ref={inputRef}
-        disabled={disabled}
-        type={'number'}
-        placeholder={'0'}
-      />
+      <InputWrapper>
+        <StyledNumberInput
+          {...props}
+          {...(percentage && {percentage, min: 0, max: 100})}
```
Fabricevladimir, line 46: When the percentage prop is set to true, the input can only show 2 characters no matter how wide it is. So if you insert 100, you only see 00

Fabricevladimir (changes requested):
Thanks! Left some comments.

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -9,43 +9,54 @@ export type NumberInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
...
-      <ButtonIcon
+      <StyledIconButton
+        name="down"
         mode="ghost"
-        size="medium"
+        size="small"
         icon={<IconRemove />}
         disabled={disabled}
-        onClick={() => inputRef.current?.stepDown()}
+        onClick={() => handleChange('up')}
```
Fabricevladimir: Small comment, the name is reversed here. Name is down, but handleChange is called with 'up'.

packages/ui-components/src/components/input/numberInput.tsx
```diff
@@ -9,43 +9,54 @@ export type NumberInputProps = React.InputHTMLAttributes<HTMLInputElement> & {
...
   disabled,
   width,
   percentage = false,
+  value,
   ...props
 }) => {
   const inputRef = useRef<HTMLInputElement>(null);
 
+  const handleChange = (mode: 'up' | 'down') => {
+    mode === 'up' ? inputRef.current?.stepDown() : inputRef.current?.stepUp();
```
Fabricevladimir: The condition is flipped.. we're calling stepDown() when 'up' is passed in.

packages/ui-components/src/components/button/buttonIcon.tsx
```diff
@@ -36,9 +36,9 @@ export const ButtonIcon: React.FC<ButtonIconProps> = ({
 };
 
 const paddingStyles = {
-  small: 'p-1',
-  medium: 'p-1.5',
-  large: 'p-2',
+  small: 'h-4 w-4 p-1',
```
Fabricevladimir: Height is technically unnecessary here because the ButtonBase takes care of that for all the buttons.

Fabricevladimir (approved):
Thanks! This can be merged once the up and down comments are addressed.

RakeshUP (approved):

PR #200, opened 2022-02-03
Feature/Links adding module - DAO-638

Fabricevladimir:
- adds AddLinks component and corresponding tests
- adds component to Create DAO -> Metadata screen


RakeshUP (approved):

packages/web-app/src/components/addLinks/row.tsx
```diff
@@ -0,0 +1,135 @@
...
+                  <AlertInline label={error.message} mode="critical" />
+                </ErrorContainer>
+              )}
+            </>
+          )}
+        />
+      </LabelContainer>
+
+      <LinkContainer>
+        <Popover
```
sepehr2github, line 52: Screen Shot 2022-02-07 at 12 33 42 AM

It's not a big deal but It always remains selected!

sepehr2github (commented):
Thank you Fabrice just a few comments:

Screen Shot 2022-02-07 at 12 44 17 AM

When the user added a new field. It can't be empty!

packages/web-app/src/components/addLinks/row.tsx
RakeshUP: I think the dropdown component tries to focus the first focussable element. Even our modal component focusses the close button always when modal opens

Fabricevladimir:
@sepehr2github The popover focuses the first focus-able element. As for the message, there is no validation for that form yet.

sepehr2github (approved):

PR #201, opened 2022-02-03
Feature/ Description list and Checkbox->Simple components DAO-657

RakeshUP:


packages/ui-components/src/components/checkbox/checkboxSimple.tsx
```diff
@@ -0,0 +1,54 @@
...
+  state = 'default',
+  onClick,
+}) => {
+  return (
+    <Container
+      data-testid="checkboxSimple"
+      disabled={disabled}
+      state={state}
+      labelLeft={labelLeft}
+      onClick={onClick}
```
Fabricevladimir, line 28: Can you disable clicking on it when it's disabled?

packages/ui-components/src/components/checkbox/checkboxSimple.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import styled from 'styled-components';
+import {Icons} from './checkboxListItem';
+
+export type CheckboxSimpleProps = {
+  label: string;
+  labelLeft?: boolean;
```
Fabricevladimir: Can you ask Selim to change this to IconLeft instead just to be consistent with the other components or do you think it makes more sense to name it like so?

Fabricevladimir (approved):
Looks good! Thanks. Can be merged once the clicking state is fixed.

packages/ui-components/src/components/checkbox/checkboxSimple.tsx
RakeshUP: Ohh yes, even I was doubtful of the naming convention. I too think it would be consistent to make the prop iconLeft. I'll change it from labelLeft to iconLeft

sepehr2github (approved):
Thank you Rakash. Looks Good to me but please take care of fabrice comments thanks!

PR #202, opened 2022-02-06
Assemble/ DAO-656 

sepehr2github:
- assemble page
- edit tailwind configuration

Also, I ignored the minimum duration description bold part because It needed to change the label component and I asked Selim if required I'm gonna do that otherwise we will tackle it later!

Note: you can ignore the number input because all changes exist on #198. I only merged those changes into this one!


RakeshUP (approved):

Fabricevladimir (commented):
Thanks, looks good. Can you handle the mobile view as well?

Fabricevladimir (approved):

PR #204, opened 2022-02-07
TokenFactory tests

mathewmeconry:
## Changes
- adds node engine restriction to package.json
- adds smock as a dependency for contracts and configures hardhat properly
- adds tests for TokenFactory.sol
- removes unused event `DAOCreated` from TokenFactory.sol

### DIFF
--- a/packages/contracts/test/core/token-factory.ts
+++ b/packages/contracts/test/core/token-factory.ts
@@ -159,6 +159,23 @@ describe('Core: TokenFactory', () => {
       expect(tx[1]).not.to.be.eq(merkleMinterBase.address);
     });
 
+    it('should emit TokenCreated', async () => {
+      const config: TokenConfig = {
+        addr: '0x0000000000000000000000000000000000000000',
+        name: 'FakeToken',
+        symbol: 'FT',
+      };
+
+      const mintConfig: MintConfig = {
+        receivers: ['0x0000000000000000000000000000000000000002'],
+        amounts: [1],
+      };
+
+      await expect(
+        tokenFactory.newToken(dao.address, config, mintConfig)
+      ).to.emit(tokenFactory, 'TokenCreated');
+    });
+
     it('should mint tokens', async () => {
       const config: TokenConfig = {
         addr: '0x0000000000000000000000000000000000000000',
@@ -176,19 +193,18 @@ describe('Core: TokenFactory', () => {
 
       const tx = await tokenFactory.newToken(dao.address, config, mintConfig);
       const rc = await tx.wait();
-      const IERC20Upgradeable = ERC20Upgradeable__factory.createInterface();
-      const transferTopic = IERC20Upgradeable.getEventTopic('Transfer');
+      const eventArgs =
+        rc.events?.find(e => e.event === 'TokenCreated')?.args || [];
 
-      const filteredLogs = rc.logs.filter(
-        log => log.topics[0] === transferTopic
+      const erc20Contract = await ethers.getContractAt(
+        'ERC20Upgradeable',
+        eventArgs[0]
       );
 
-      expect(filteredLogs.length).to.be.eq(2);
-      for (const i in filteredLogs) {
-        const log = filteredLogs[i];
-        const event = IERC20Upgradeable.parseLog(log);
-        expect(event.args.to).to.be.eq(mintConfig.receivers[i]);
-        expect(event.args.value).to.be.eq(mintConfig.amounts[i]);
+      for (const i in mintConfig.receivers) {
+        expect(
+          await erc20Contract.callStatic.balanceOf(mintConfig.receivers[i])
+        ).to.be.eq(mintConfig.amounts[i]);
       }
     });
 
--- a/packages/contracts/test/core/token-factory.ts
+++ b/packages/contracts/test/core/token-factory.ts
@@ -192,131 +192,49 @@ describe('Core: TokenFactory', () => {
       }
     });
 
-    describe('DAO Permissions', () => {
-      it('should grant TOKEN_MINTER_ROLE for TokenFactory', async () => {
-        const config: TokenConfig = {
-          addr: '0x0000000000000000000000000000000000000000',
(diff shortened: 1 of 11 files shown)


packages/contracts/test/core/token-factory.ts
```diff
@@ -0,0 +1,322 @@
...
+        amounts: [1],
+      };
+
+      const tx = await tokenFactory.callStatic.newToken(
+        dao.address,
+        config,
+        mintConfig
+      );
+
+      expect(tx[0]).not.to.be.eq(governanceWrappedBase.address);
```
novaknole, line 95: * This seems that the check doesn't serve any strong purpose.  Whatever I almost do/change in `TokenFactory`'s contract, the test would still succeed.

packages/contracts/test/core/token-factory.ts
```diff
@@ -0,0 +1,322 @@
...
+          '0x0000000000000000000000000000000000000002',
+          '0x0000000000000000000000000000000000000003',
+        ],
+        amounts: [1, 5],
+      };
+
+      const tx = await tokenFactory.newToken(dao.address, config, mintConfig);
+      const rc = await tx.wait();
+      const IERC20Upgradeable = ERC20Upgradeable__factory.createInterface();
+      const transferTopic = IERC20Upgradeable.getEventTopic('Transfer');
```
novaknole: In this test , can't we use `to.emit` and `withArgs` solution ? Much cleaner and easier.

packages/contracts/test/core/token-factory.ts
```diff
@@ -0,0 +1,322 @@
...
+          amounts: [1],
+        };
+
+        const tx = await tokenFactory.callStatic.newToken(
+          dao.address,
+          config,
+          mintConfig
+        );
+
+        expect(dao.grant).to.have.been.calledWith(
```
novaknole: In this test, we should also have the check that `dao.revoke` was called with those arguments. no need to have another test which is below this one.

packages/contracts/test/core/token-factory.ts
```diff
@@ -0,0 +1,322 @@
...
```
novaknole: This can also be in the same test above.

packages/contracts/test/core/token-factory.ts
```diff
@@ -0,0 +1,322 @@
...
```
novaknole: same goes for this. Idea is that 

```
expect(dao.grant).to.have.been.calledWith
```

all the calls  can be inside one test, since `tokenConfig.addr` is 0x...000 for all of them.

packages/contracts/test/core/token-factory.ts
```diff
@@ -0,0 +1,322 @@
...
```
novaknole: same for this as well.

packages/contracts/test/core/token-factory.ts
```diff
@@ -0,0 +1,322 @@
...
+        amounts: [1],
+      };
+
+      let tx = await tokenFactory.callStatic.newToken(
+        dao.address,
+        config,
+        mintConfig
+      );
+
+      expect(tx[0]).not.to.be.eq(governanceBase.address);
```
novaknole, line 137: This seems that the check doesn't serve any strong purpose. Whatever I almost do/change in TokenFactory's contract, the test would still succeed.

novaknole (changes requested):
Some stuff can be added in the tests as well and distributorBase is not mentioned anywhere.

For example, 

```
MerkleMinter(merkleMinter).initialize(
            _dao,
            GovernanceERC20(token),
            distributorBase
        );
  ```
  
  We can use `toHaveBeenCalledWith` here for the merkle minter and so on.

**Comment by mathewmeconry** [comment]
> Some stuff can be added in the tests as well and distributorBase is not mentioned anywhere.
> 
> For example,
> 
> ```
> MerkleMinter(merkleMinter).initialize(
>             _dao,
>             GovernanceERC20(token),
>             distributorBase
>         );
> ```
> 
> We can use `toHaveBeenCalledWith` here for the merkle minter and so on.

unfortunately not. Because of the call to `clone` we cannot deploy a mocked or faked contract to this address otherwise `clone` will fail. And hooking into after a contract is deployed is with smock not possible (yet)

File: packages/contracts/test/core/token-factory.ts (hunk 1 above)
Comment (mathewmeconry) [reply packages/contracts/test/core/token-factory.ts:95]: at least it ensures that we don't use the contract deployed in `setupBases`. So I would like to leave this test in

File: packages/contracts/test/core/token-factory.ts (hunk 2 above)
Comment (mathewmeconry) [reply packages/contracts/test/core/token-factory.ts]: we can but then we cannot test multiple mints in one call. As far as I know `to.emit` and `withArgs` can only test for one event and not multiple at the same time

File: packages/contracts/test/core/token-factory.ts (hunk 7 above)
Comment (mathewmeconry) [reply packages/contracts/test/core/token-factory.ts:137]: at least it ensures that we don't use the contract deployed in `setupBases`. So I would like to leave this test in

File: packages/contracts/test/core/token-factory.ts (hunk 3 above)
Comment (mathewmeconry) [reply packages/contracts/test/core/token-factory.ts]: done

File: packages/contracts/test/core/token-factory.ts (hunk 2 above)
Comment (novaknole) [reply packages/contracts/test/core/token-factory.ts]:
```
tx.to.emit('Transfer').withArgs(var1, var2, var3).to.emit('Transfer).withArgs(var4, var5, var6)
```

This should work. Have done it before.

mathewmeconry: changed to use the erc20 contract address which gets emitted in the new `TokenCreated` event

packages/contracts/test/core/token-factory.ts
nivida: more assertions are always good :)

nivida:
@novaknole Discussed it with @mathewmeconry and merged this PR after.

nivida (approved):

PR #206, opened 2022-02-07
Feature/Assemble Define Proposal page DAO-645

RakeshUP:


packages/web-app/src/components/addLinks/index.tsx
```diff
@@ -19,19 +23,22 @@ const AddLinks: React.FC = () => {
...
-      </ListGroup>
+      {fields.length > 0 && (
+        <ListGroup>
+          <Header />
+          {fields.map((field, index) => (
+            <Row
+              key={field.id}
+              index={index}
+              control={control}
+              {...(fields.length > 1 ? {onDelete: () => remove(index)} : {})}
```
Fabricevladimir: Currently, this disables the remove button if there is only one row. Do you want me modify it since it's not necessary to show at least one row (we can show add link button)?

packages/web-app/src/containers/defineProposal/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+const DefineProposal: React.FC = () => {
+  const {t} = useTranslation();
+  const {account, ensAvatarUrl}: useWalletProps = useWallet();
+
+  return (
+    <>
+      <FormItem>
+        <Label label={t('labels.author')} />
+
+        <ButtonWallet
```
Fabricevladimir, line 24: Can you add the form controller stuff to this? Or better yet, maybe create a ticket for form integration and validation

Fabricevladimir (approved):
Thanks! Looking good

packages/web-app/src/components/addLinks/index.tsx
```diff
@@ -19,19 +23,22 @@ const AddLinks: React.FC = () => {
...
-        {fields.map((field, index) => (
-          <Row
-            key={field.id}
-            index={index}
-            control={control}
-            {...(fields.length > 1 ? {onDelete: () => remove(index)} : {})}
-          />
-        ))}
-      </ListGroup>
+      {fields.length > 0 && (
```
sepehr2github, line 26: So why did you hide the list when the length of the fields is 0? Can explain. Thanks

sepehr2github (approved):
Thank you Rakash. Please add form controllers and you're good to go

packages/web-app/src/components/addLinks/index.tsx
RakeshUP: The figma design was like that

packages/web-app/src/containers/defineProposal/index.tsx
RakeshUP: Created a new ticket

packages/web-app/src/components/addLinks/index.tsx
RakeshUP: I have done the change. Removed the condition. 
Add link button seems like overkill for me

PR #212, opened 2022-02-08
feature / DAO-677

Rekard0:
- Aggregate balances
- Add timestamp to vault transfers & balances
- Update unit testing

### DIFF
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -10,9 +10,10 @@ type ERC20Token @entity {
 
 type ERC20Balance @entity {
   id: ID! # token address + dao address
-  token: ERC20Token!
+  token: ERC20Token! # address(0) will be considered as ETH (or canonical token)
   dao: Dao!
   balance: BigInt!
+  lastUpdated: BigInt!
 }
 
 # Dao
@@ -32,13 +33,7 @@ type VaultWithdraw @entity {
   to: Bytes!
   amount: BigInt!
   reference: String!
-}
-
-type VaultEthDeposit @entity {
-  id: ID!
-  dao: Dao!
-  sender: Bytes!
-  amount: BigInt!
+  createdAt: BigInt!
 }
 
 type VaultDeposit @entity {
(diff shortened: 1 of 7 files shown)


packages/subgraph/schema.graphql
```diff
@@ -10,9 +10,10 @@ type ERC20Token @entity {
 
 type ERC20Balance @entity {
```
nivida: Cause we have ETH handled with the same struct can we also name it ``Balance``.. wdyt?

packages/subgraph/src/dao.ts
```diff
@@ -28,22 +29,105 @@ export function handleExecuted(event: Executed): void {
   // TODO:
 }
 
+function updateERC20Balance(
```
nivida: Same naming question here :))

packages/subgraph/src/dao.ts
```diff
@@ -28,22 +29,105 @@ export function handleExecuted(event: Executed): void {
...
+  let tokenContract = GovernanceWrappedERC20.bind(token);
+  let daoBalance = tokenContract.try_balanceOf(daoAddress);
+  if (!daoBalance.reverted) {
+    entity.balance = daoBalance.value;
+  }
+  entity.lastUpdated = timestamp;
+  entity.save();
+}
+
+function updateEthBalance(
```
nivida: I think we have here some code duplication we could remove :-)

nivida (changes requested):

packages/subgraph/src/dao.ts
Rekard0: amended

packages/subgraph/schema.graphql
Rekard0: yes, makes sense , amended

packages/subgraph/src/dao.ts
Rekard0: yes, amended

nivida (approved):

PR #219, opened 2022-02-09
Assemble/ DAO-652 

sepehr2github:
- New token creation added
- responsive view added
- new token fields validation added


packages/ui-components/src/components/input/textInput.tsx
```diff
@@ -40,7 +40,7 @@ export const Container = styled.div.attrs(
       disabled ? 'bg-ui-100 border-ui-200 border-2' : 'bg-ui-0'
```
Fabricevladimir, line 40: This should probably be updated in a separate ticket, but I noticed that on the designs, the adornment button isn't actually disabled when the input itself is disabled...

packages/web-app/src/components/addWallets/row.tsx
```diff
@@ -0,0 +1,202 @@
...
+              <ListItemText
+                title={t('labels.removeWallet')}
+                {...(typeof onDelete === 'function'
+                  ? {mode: 'default', onClick: () => onDelete(index)}
+                  : {mode: 'disabled'})}
+              />
+            </div>
+          }
+        >
+          <ButtonIcon
```
Fabricevladimir, line 129: It seems for the DAO Treasury, this button is not present, well at least not on the desktop view.

packages/web-app/src/components/addWallets/row.tsx
```diff
@@ -0,0 +1,202 @@
...
+export type WalletField = {
+  id: string;
+  address: string;
+  amount: string;
+};
+
+const LinkRow: React.FC<LinkRowProps> = ({control, index, onDelete}) => {
+  const {t} = useTranslation();
+  const {account} = useWallet();
+  const {getValues, setValue} = useFormContext();
```
Fabricevladimir: If you're using `useFormContext`, you can simplify the props by removing control and getting it from the `useFormContext` hook.

packages/web-app/src/locales/en/common.json
```diff
@@ -245,10 +258,15 @@
...
+      "nameSubtitle": "The full name of the token eg. Aragon",
+      "symbolSubtitle": "The symbol of your token e.g. ANT (good practice is to use 3-4 letters)",
       "addExistingTokenHelptext": "Hinterlege einen bestehenden Token, welcher deine Community für die Governance verwendet. To learn more about what it’s important if you use an existing Token, read first the ",
       "tokenHelptextLink": "Token Distribution Guide",
       "tokenContractSubtitlePart1": "Copy the contract address directly from the ",
-      "tokenContractSubtitlePart2": "This way you can ensure that the address is correct."
+      "tokenContractSubtitlePart2": "This way you can ensure that the address is correct.",
+      "distributeTokensHelpertext": "Determine who receives how many tokens. Unless otherwise specified, the treasury receives the rest tokens for the future distribution of tokens (these tokens have no voting rights and do not influence the quorum).",
+      "createTokenHelptext": "Mint a community token on a variable or fixed supply. To learn more about how to mint the token and how to use it, read the ",
+      "createTokenHelplink": "Token Minting Guide."
```
Fabricevladimir: Small typo. The period should be removed.

Fabricevladimir (changes requested):
Thanks! Good work. Left a couple of comments. The only major issue I saw was that when a validation message is shown (on desktop) it causes the corresponding percentage input to drop down.
image

RakeshUP (approved):

packages/ui-components/src/components/input/textInput.tsx
sepehr2github: actually, there are two versions of the designs. I implement this based on the user testing branch and on that branch the button is disabled (actually there is no adornment at all but I left the refactor for later because the functionality is the same)

packages/web-app/src/components/addWallets/row.tsx
sepehr2github: Yes and no. You. can see the button here but in general, you're right
Screen Shot 2022-02-11 at 3 55 46 PM

packages/web-app/src/locales/en/common.json
sepehr2github: Which period?

Fabricevladimir: The one for createTokenHelpLink. It's displaying a period with the link

sepehr2github: Screen Shot 2022-02-11 at 5 53 31 PM
Can you commit the changes there is no period here? I don't understand :)))

Fabricevladimir (approved):

PR #221, opened 2022-02-10
Feature/ DAO-675

Rekard0:
- implement ACL mapping
- implement Governance entities
- implement unit testing for ACL


packages/subgraph/schema.graphql
```diff
@@ -1,6 +1,20 @@
 # Types
 
+# Governances
+
+type DaoGovernance @entity {
```
nivida: Can you elaborate closer on what the idea behind ``DAOGovernance`` and ``Governance`` is?

Rekard0: Governance is a package or an app like SimpleVoting
DAOGovernance is the Many-to-Many relationship : https://thegraph.com/docs/en/developer/create-subgraph-hosted/#many-to-many-relationships

nivida: I see. Thanks. Wdyt about calling it already packages and we eg have a type field to identify packages/contracts that have the execution role?

Rekard0: yes, amended

nivida (approved):
LGTM! Thanks for applying the requested changes! :)

PR #226, opened 2022-02-11
Feature: Assemble voting setup page - Dao 621

WissenIstNacht:
This PR implements step 2 of the withdrawal form, voting setup:
- assemble necessary UI components on the page
- perform validation on user input


packages/web-app/src/components/inputTime/inputTime.tsx
```diff
@@ -0,0 +1,33 @@
+import styled from 'styled-components';
```
sepehr2github, line 1: I'm not sure if It's the right place for this comment :). but, the date format in designs is yyyy/mm/dd but in this pr is mm/dd/yyyy.

packages/web-app/src/containers/setupVotingForm/index.tsx
```diff
@@ -0,0 +1,375 @@
...
+   *************************************************/
+
+  const [endDateType, setEndDateType] = useState<EndDateType>('duration');
+  const [utcInstance, setUtcInstance] = useState<UtcInstance>('first');
+  const [utcStart, setUtcStart] = useState('');
+  const [utcEnd, setUtcEnd] = useState('');
+
+  // Initializes values for the form
+  // This is done here rather than in the defaulValues object as time can
+  // ellapse between the creation of the form context and this stage of the form.
```
sepehr2github, line 48: I know. implementing these validators weren't easy at all. Good job! 👍

nivida (commented):
I've seen it does add 18 code smells to the code base. Would be great to at least address the Major and Critical code smells. 

Hint: I think the coverage generation is missing so that SonarQube can analyze it as well.

sepehr2github (changes requested):
Well Done Vali! you did a great job. I left A few comments.

1) The time picker looks broken on my screen. maybe it needs some width adjustment
Screen Shot 2022-02-14 at 3 07 18 PM
Also, we need to fix the alignment issues!

2) I think you mention this before. I don't remember if the issue was this or something else but when I selected my start date and end date the next button was still disabled. (without any alert messages) 
Screen Shot 2022-02-14 at 3 13 30 PM

packages/web-app/src/components/inputTime/inputTime.tsx
WissenIstNacht: The date displayed in the input depends on the browser's locale :)

I already notified Selim about it a while ago 👍🏼

WissenIstNacht:
Hey @sepehr2github, thanks for the review! About the two points you mentioned:

1. I forgot to check chrome before creating the PR, sorry 🙈 The alignment issues should be resolved now 👍🏼 
2. Yes, that's what I explained on a daily a few days back. The form thinks that something is wrong, which is why it does not let you go on. However, I'm not 100% sure what it thinks is wrong. Interestingly, if you go back a step, the button is still disabled. If you then change the token amount input, though, it will turn back to normal (also in the next step).

packages/web-app/src/components/inputTime/inputTime.tsx
sepehr2github: Ah. Thanks 👍

RakeshUP (approved):

RakeshUP:
Looks good. For me in Chrome browser, the `M` in `PM` seems to be cut off

Screenshot 2022-02-15 at 6 43 57 PM

sepehr2github (approved):

PR #233, opened 2022-02-14
Feature / DAO-629

Rekard0:
- Implement Simple Voting Template
- Unit testing
- Updating Contracts for Execution and White List Voting
- Implement Execution Mapping


nivida (approved):
LGTM! The only thing we could do here as we talked about on the sync is renaming SimpleVoting to ERC20Voting.

PR #235, opened 2022-02-15
Chore/Apollo client DAO-682

RakeshUP:
Only the fetchTokenData is changed to use Apollo Client REST Link, because that data seems to be constant for a chain and can be persisted. To persist a query, you need to add the typename to entitiesToPersist in apolloClient.tsx 

The token data loading in finance page is wee bit faster from the subsequent loads

### DIFF
--- a/packages/web-app/src/context/apolloClient.tsx
+++ b/packages/web-app/src/context/apolloClient.tsx
@@ -1,0 +1,23 @@
+import {ApolloClient, InMemoryCache} from '@apollo/client';
+import {RestLink} from 'apollo-link-rest';
+import {CachePersistor, LocalStorageWrapper} from 'apollo3-cache-persist';
+
+const restLink = new RestLink({
+  uri: 'https://api.coingecko.com/api/v3',
+});
+
+const cache = new InMemoryCache();
+
+const persistor = new CachePersistor({
+  cache,
+  maxSize: 5242880, // 5 MiB
+  storage: new LocalStorageWrapper(window.localStorage),
+  debug: process.env.NODE_ENV === 'development',
+});
+
+await persistor.restore();
+
+export const client = new ApolloClient({
+  cache,
+  link: restLink,
+});
--- a/packages/web-app/src/context/apolloClient.tsx
+++ b/packages/web-app/src/context/apolloClient.tsx
@@ -14,6 +14,7 @@ const entitiesToPersist = ['tokenData'];
 
 const persistor = new CachePersistor({
   cache,
+  // TODO: Check and update the size needed for the cache
   maxSize: 5242880, // 5 MiB
   storage: new LocalStorageWrapper(window.localStorage),
   debug: process.env.NODE_ENV === 'development',
--- a/packages/web-app/src/context/apolloClient.tsx
+++ b/packages/web-app/src/context/apolloClient.tsx
@@ -1,18 +1,57 @@
 import {ApolloClient, InMemoryCache} from '@apollo/client';
 import {RestLink} from 'apollo-link-rest';
 import {CachePersistor, LocalStorageWrapper} from 'apollo3-cache-persist';
+import {BASE_URL} from 'utils/constants';
 
 const restLink = new RestLink({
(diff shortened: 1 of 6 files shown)


packages/web-app/src/context/apolloClient.tsx
```diff
@@ -0,0 +1,62 @@
...
+});
+
+const cache = new InMemoryCache();
+
+// add the REST API's typename you want to persist here
+const entitiesToPersist = ['tokenData'];
+
+const persistor = new CachePersistor({
+  cache,
+  maxSize: 5242880, // 5 MiB
```
sepehr2github, line 18: Why did you set max size 5 mb?

sepehr2github (commented):
Thank you Rakash. There are some lint errors. Can you fix them?! 👍

packages/web-app/src/context/apolloClient.tsx
RakeshUP: It was a random number that came to my mind while working on this. We would certainly have to change this when we come to know about the static coingecko data size(token metadata in every chain)

RakeshUP:
Fixed the lint errors @sepehr2github

packages/web-app/src/context/apolloClient.tsx
```diff
@@ -0,0 +1,65 @@
+import {ApolloClient, InMemoryCache} from '@apollo/client';
+import {RestLink} from 'apollo-link-rest';
+import {CachePersistor, LocalStorageWrapper} from 'apollo3-cache-persist';
+import {BASE_URL} from 'utils/constants';
+
+const restLink = new RestLink({
+  uri: BASE_URL,
```
WissenIstNacht, line 7: just out of couriosity, what do we do if we need to query another API? Do we need to create another apollo client object with a different link?

packages/web-app/src/services/prices.ts
```diff
@@ -76,24 +78,41 @@ type FetchedTokenData = Promise<TokenData | undefined>;
...
+  if (!error && data.tokenData) {
+    return {
+      id: data.tokenData.id,
+      name: data.tokenData.name,
+      symbol: (data.tokenData.symbol as string).toUpperCase(),
+      imgUrl: data.tokenData.image.large,
+      address: address,
+    };
   }
+  console.error('Error fetching token price', error);
```
WissenIstNacht, line 115: Any partiular reason you're logging the error instead of throwing the one (potentially) returned by the `query`? I personally would prefer for this to throw an error as they are a) easier to work with (clear structure, typed, etc.) and b) we will likely not perform error handling in here.

WissenIstNacht (commented):
Thanks, seems to work well :)

Let me know what you think on the comments.

packages/web-app/src/context/apolloClient.tsx
RakeshUP: We define a new endpoint like in this [line](https://github.com/RakeshUP/apollo-client-poc/blob/main/src/cache.ts#L105) and pass the endpoint argument to queries whenever needed like [here](https://github.com/RakeshUP/apollo-client-poc/blob/main/src/pages/Home.tsx#L50)

packages/web-app/src/services/prices.ts
RakeshUP: All the API calls are handled this way and I agree that we shouldn't be catching the errors here. Even better we should just return the apollo client query promise instead of using await here.

Also, this very fetchTokenData function is soon to be refactored to support multiple chains, maybe we take care of the error handling then?

WissenIstNacht: Yup, just after posting this I realized that you probably just followed what was there. Thanks for the reply anyway. 👍🏼

WissenIstNacht (approved):

sepehr2github (approved):

packages/web-app/src/context/apolloClient.tsx
sepehr2github: But Please add a proper comment as a reminder for the future!

PR #236, opened 2022-02-15
Feature/ DAO-694

Rekard0:
- refactor SimpleVoting incontracts
- refactor SimpleVoting in Subgraph
- refactor subgraph constants
- redeploy contracts


packages/subgraph/schema.graphql
```diff
@@ -115,40 +115,40 @@ interface Proposal {
...
+  proposals: [ERC20VotingProposal!]! @derivedFrom(field: "erc20VotingPkg")
   supportRequiredPct: BigInt
   minAcceptQuorumPct: BigInt
   voteTime: BigInt
   votesLength: BigInt
   token: ERC20Token
 }
 
-type EVVoter @entity {
+type ERC20VotingVoter @entity {
```
nivida: Can't we call this also just voter?

packages/subgraph/schema.graphql
```diff
@@ -115,40 +115,40 @@ interface Proposal {
...
   id: ID! # address
-  proposals: [EVVoterProposal!]! @derivedFrom(field: "voter")
+  proposals: [ERC20VotingVoterProposal!]! @derivedFrom(field: "voter")
 }
 
-type EVVoterProposal @entity {
+type ERC20VotingVoterProposal @entity {
```
nivida: This probably could be called ``ERC20Vote``?

nivida (changes requested):
LGTM! Only the naming of the ERC20Voting entities in the schema could get simplified slightly. Would be great if you could also elaborate closer on why ``ERC20VotingVoter`` and ``ERC20VotingVoterProposal`` is required? :) 

``ERC20VotingVoter`` & ``ERC20VotingVoterProposal`` are as commented a bit strange names. :)

packages/subgraph/schema.graphql
```diff
@@ -115,40 +115,40 @@ interface Proposal {
...
 
-type EVProposal implements Proposal @entity {
+type ERC20VotingProposal implements Proposal @entity {
   id: ID! # package + voteId
   dao: Dao!
   createdAt: BigInt!
   actions: [Action!]! @derivedFrom(field: "proposal")
   pkg: Package!
-  evPkg: EVPackage!
+  erc20VotingPkg: ERC20VotingPackage!
```
nivida: Here the ``erc20VotingPkg`` field probably can just be called ``contract``? Totally open to hear about other suggestions :)

nivida (changes requested):

packages/subgraph/schema.graphql
Rekard0: i don't think we can, because later we will have whitelistVoter, but instead it can implements Voter interface.

packages/subgraph/schema.graphql
Rekard0: yes we can, done

packages/subgraph/schema.graphql
Rekard0: thanks,  i noticed basically pkg and erc20VotingPkg are essentially the same, so now we can only use pkg.
i prefer to keep the field name as pkg, because it is of package interface

packages/subgraph/schema.graphql
nivida: Yeah.. I think we can fix this as soon the package registry is existing and those voting types are handled as package with a separate subgraph 👍

packages/subgraph/schema.graphql
nivida: agreed :-)

nivida (approved):

PR #240, opened 2022-02-16
Assemble page/ DAO-658 

sepehr2github:
- add a review page
- review checks validation added
- change chain selection datastore
- add controlled singleInputImage
- add create Dao page validations


RakeshUP (commented):
@sepehr2github Did you take a look at the Description List component that exists in the web-app package? It could be used like [this](https://github.com/aragon/zaragoza/blob/cf239cb1dff9f74f611e1967e5a78fc15a0c5401/packages/web-app/src/pages/home.tsx#L32).

sepehr2github:
> @sepehr2github Did you take a look at the Description List component that exists in the web-app package? It could be used like [this](https://github.com/aragon/zaragoza/blob/cf239cb1dff9f74f611e1967e5a78fc15a0c5401/packages/web-app/src/pages/home.tsx#L32).

No @RakeshUP I wasn't aware of that :/

packages/web-app/src/locales/en/common.json
```diff
@@ -83,6 +85,16 @@
...
       "addresses": "{{count}} Addresses"
     },
+    "review": {
+      "blockchain": "Blockchain",
+      "daoMetadata": "DAO Metadata",
+      "community": "Community",
+      "governance": "Governance",
+      "network": "Network",
+      "eligibleMembers": "Eligible Members",
+      "Supply": "Supply",
```
WissenIstNacht: two things here:

1) Keep the variable names lower case
2) The string "Supply" is already defined a few lines above. I suggest keeping the other one, but renaming it from `existingTokenSupply` to just `supply`.

packages/ui-components/src/components/input/inputImageSingle.tsx
```diff
@@ -31,6 +31,10 @@ export type InputImageSingleProps = {
    * Alow Square image only
    */
   onlySquare?: boolean;
+  /**
+   * Passing image src for preview
+   */
+  preview?: string | null;
```
WissenIstNacht: Not super important, but if the preview source is a string, `''` might be better than `null` (you can still treat `''` as not having a preview, but you wont ever get into trouble with null.)

packages/web-app/src/components/addWallets/footer.tsx
```diff
@@ -11,7 +11,7 @@ type WalletsFooterProps = {
 const AddWalletsFooter: React.FC<WalletsFooterProps> = ({totalAddresses}) => {
   const {t} = useTranslation();
   const {getValues} = useFormContext();
-  const TotalSupply = getValues('totalTokenSupply');
+  const TotalSupply = getValues('tokenTotalSupply');
```
WissenIstNacht: lowercase

packages/web-app/src/components/addWallets/row.tsx
```diff
@@ -36,15 +36,14 @@ const LinkRow: React.FC<LinkRowProps> = ({index, onDelete}) => {
   const {control, getValues, setValue, trigger} = useFormContext();
   const walletFieldArray = getValues('wallets');
```
WissenIstNacht, line 37: Not sure where to put this (since it was on an earlier PR). But this component is not a `LinkRow` :p

packages/web-app/src/pages/createDAO.tsx
```diff
@@ -55,6 +64,29 @@ const CreateDAO: React.FC = () => {
     errors.links,
   ]);
 
+  const daoSetupCommunityIsValid = useMemo(() => {
```
WissenIstNacht, line 67: This doesn't work when using an existing token :( (The button is disabled for me, even when the contract address is correct)

packages/web-app/src/containers/setupCommunity/addExistingToken.tsx
```diff
@@ -45,9 +45,9 @@ const AddExistingToken: React.FC = () => {
   );
 
   const explorer = useMemo(() => {
-    if (chainId) {
+    if (chainId.id) {
```
WissenIstNacht, line 48: Just FYI: The form data type doesn't contain a `chainId` (or `blockchain`) field

packages/web-app/src/containers/goLive/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+        </p>
+      </div>
+      <ImageContainer src={goLive} />
+    </div>
+  );
+};
+
+const GoLive: React.FC = () => {
+  return (
+    <Container>
```
WissenIstNacht, line 34: I'm writing this here because it concerns all the cards in this containers: The cards currently grow in width when the screen gets wider, which they should not. (The header should, but not the summary cards.)

packages/web-app/src/containers/goLive/blockchain.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <Header>
+        <Title>{t('labels.review.blockchain')}</Title>
+      </Header>
+      <Body>
+        <Row>
+          <LabelWrapper>
+            <Label>{t('labels.review.network')}</Label>
+          </LabelWrapper>
+          <TextContent>
+            {t('createDAO.review.network', {network: blockchain.network})}
```
WissenIstNacht, line 26: Something is wrong here. The Netowrk is showing "main net", but it should be "Main Net".

packages/web-app/src/containers/goLive/daoMetadata.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <Card>
+      <Header>
+        <Title>{t('labels.review.daoMetadata')}</Title>
+      </Header>
+      <Body>
+        <Row>
+          <LabelWrapper>
+            <Label>{t('labels.logo')}</Label>
+          </LabelWrapper>
+          {daoLogo && (
```
WissenIstNacht, line 43: The logo should be center-aligned with the label "Logo"

packages/web-app/src/containers/goLive/community.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <Card>
+      <Header>
+        <Title>{t('labels.review.community')}</Title>
+      </Header>
+      <Body>
+        <Row>
+          <LabelWrapper>
+            <Label>{t('labels.review.eligibleMembers')}</Label>
+          </LabelWrapper>
+          <TextContent>{membership}</TextContent>
```
WissenIstNacht: Should say "Token Holders" instead of "token".

packages/web-app/src/containers/goLive/community.tsx
```diff
@@ -0,0 +1,109 @@
...
+            {isCustomToken && <Badge label="New" colorScheme="info" />}
+          </BadgeWrapper>
+        </Row>
+        <Row>
+          <LabelWrapper>
+            <Label>{t('labels.review.Supply')}</Label>
+          </LabelWrapper>
+          <BadgeWrapper>
+            <TextContent>
+              {tokenTotalSupply} {tokenSymbol}
```
WissenIstNacht, line 66: This doesn't update corretly. For example, I entered the ANT token to use an existing token. Then I changed to create a token. I added a name, symbol and distribution. But in the end it shows ANT with a distribution, instead of the nem name and symbol I added :)

packages/web-app/src/containers/goLive/governance.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <Label>{t('labels.minimumSupport')}</Label>
+          </LabelWrapper>
+          <TextContent>{support}%</TextContent>
+        </Row>
+        <Row>
+          <LabelWrapper>
+            <Label>{t('labels.minimumDuration')}</Label>
+          </LabelWrapper>
+          <TimeWrapper>
+            <div>{t('createDAO.review.days', {durationDays})}</div>
```
WissenIstNacht: Doesn't show any day/hour/minute.

WissenIstNacht (changes requested):
Hey Sepehr, thanks  a lot for the PR.

The page assembly on the review ("go live" page) looks good visually :muscle: 

However, there are a lot of important things that do not work in this flow. I don't think I did a review on this since the blockchain selection, so it's hard to judge what's going on. I've noted a few things above, but there's more. For example:
- The wallet list form does not complain when a row is empty. 
- The wallet list form does not complain when a row has an amount but no wallet address.
- Deleting a row from the wallet list form does not update the total in the footer.
- Selecting an existing token on the second step leaves the "continue" button disabled.
- The stepper shows a wrong number of steps (this is because there are more pages that steps, since the "intro" and the "go live" page count as part of the stepper as well)
- The minium approval amount does not change in the inline alert on "configure governance"
- ...

I'm sure it's not all your fault, you just have the bad luck of being the one who finishes the flow :p But since this PR is supposed to include validation (DAO 659) AND it's the last PR for this flow, we need to address these changes somehow. My suggestion would be this: remove the 659 ticket from this PR and send it back to "to do". This way, we can merge the "visual" part (DAO 658). Then, we can and fix everything that is currently broken in the form and validation on 659. Wdyt?

packages/web-app/src/containers/goLive/community.tsx
sepehr2github: I will fix it in new PR

packages/web-app/src/containers/goLive/community.tsx
sepehr2github: I will fix it in new PR

packages/web-app/src/containers/goLive/daoMetadata.tsx
sepehr2github: I will fix it in new PR

packages/web-app/src/containers/goLive/blockchain.tsx
sepehr2github: I will fix it in new PR

packages/web-app/src/containers/goLive/index.tsx
sepehr2github: I should fix the wab app grid and I will do it in next PR

packages/web-app/src/containers/setupCommunity/addExistingToken.tsx
sepehr2github: Next PR

sepehr2github:
This Pr is fine in visual view and I will send another PR to fix all other remaining logic issues. Please review this PR (only visual aspect) @WissenIstNacht @Fabricevladimir

packages/web-app/src/containers/goLive/community.tsx
WissenIstNacht: Seems like a visual thing to me, though 😅

packages/web-app/src/containers/goLive/index.tsx
WissenIstNacht: Which next PR? 😅 

No but seriously, 659 is about validation. Introducing a grid system is a bit out of scope there. So you'd have to open a new ticket. But by the time you or someone else finishes it, this will be blocked in QA... Also I feel like this is quite fundamental in this PR. And the fix is literally adding a width class in two places...

But yeah, if all this doesn't convince you add a ticket for later 😉

packages/web-app/src/containers/goLive/community.tsx
sepehr2github: Sure! No worries. I'm gonna fix it now

packages/web-app/src/containers/goLive/index.tsx
sepehr2github: I had a discussion with Fabrice and he will create a separate task for grids. but I fixed it using tailwind classes

WissenIstNacht:
Looking really good, thanks a lot! 💪🏼

WissenIstNacht (approved):

RakeshUP (approved):

PR #241, opened 2022-02-17
Feature/Assemble Withdraw Review Page DAO-592

RakeshUP:


packages/web-app/src/containers/reviewWithdraw/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+          )}
+
+          <VotingTerminal />
+
+          <CardExecution
+            title="Execution"
+            description="These smart actions are executed when the proposal reaches sufficient support. Find out which actions are executed."
+            to={values.to}
+            from={values.from}
+            toLabel="To"
```
Fabricevladimir: Can you update to use the labels already present in the translation files?

packages/web-app/src/containers/reviewWithdraw/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+                onClick={() => setExpandedProposal(false)}
+              />
+            </>
+          )}
+
+          <VotingTerminal />
+
+          <CardExecution
+            title="Execution"
+            description="These smart actions are executed when the proposal reaches sufficient support. Find out which actions are executed."
```
Fabricevladimir: If title and description are going to be the same across all withdraws, can you move the strings into the translations?

packages/web-app/src/containers/reviewWithdraw/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+            to={values.to}
+            from={values.from}
+            toLabel="To"
+            fromLabel="From"
+            tokenName={values.tokenName}
+            tokenImageUrl={values.tokenImgUrl}
+            tokenSymbol={values.tokenSymbol}
+            tokenCount={values.amount}
+            treasuryShare="$TODO:"
```
Fabricevladimir: This is the token price correct? We fetch this back in the token menu, we can either add it to the form, or fetch it again here considering that the user might take eternity before they complete the form. Thoughts? 

Also, there is a fetchTokenPrice function that takes the token address if you want to use that. Currently it's only supporting Ethereum but we'll modify it once the Apollo client cacheing is finalized.

Fabricevladimir (commented):
Good work! Thanks.

Fabricevladimir:
One thing I also noticed is that because the back and continue buttons are outside the individual steps, on the reviewProposal, the Publish Proposal button is off since that page takes up the full width of the container. Should a custom footer be passed into the step instead? Thoughts? 🤔

packages/web-app/src/locales/en/common.json
```diff
@@ -245,8 +245,7 @@
       "title": "Setup Voting",
       "description": "For a withdraw, a vote is necessary so that the majority can determine whether your proposal is legitimate. For this, you need this configuration.",
       "optionLabel": {
-        "title": "Options",
-        "helptext": "This is a placeholder"
```
WissenIstNacht, line 255: Thanks for reminding me of this. I've dropped a question on figma now.

packages/web-app/src/containers/reviewWithdraw/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+                className="mt-3 w-full tablet:w-max"
+                label={t('governance.proposals.buttons.closeFullProposal')}
+                mode="secondary"
+                iconRight={<IconChevronUp />}
+                onClick={() => setExpandedProposal(false)}
+              />
+            </>
+          )}
+
+          <VotingTerminal />
```
WissenIstNacht, line 93: FYI I just realized that table on the "Voters" tab does not fit inside the container.

I'll leave it up to you to judge if you want to change that in this PR or do it in another one.

packages/web-app/src/containers/reviewWithdraw/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+          iconRight={<IconChevronDown />}
+          onClick={() => setExpandedProposal(true)}
+        />
+      )}
+
+      <ContentContainer expandedProposal={expandedProposal}>
+        <ProposalContainer>
+          {expandedProposal && (
+            <>
+              <EditorContent editor={editor} />
```
WissenIstNacht: Links created with the WYSIWYG editor on the previous page do not show/work.

Again, I'll let oyu be the judge of when you want to fix this. If you don't do it now, please add a comment.

WissenIstNacht (commented):
Thanks man. Looks good, works well.

Pleas either address my comments or leave a TODO comment.

packages/web-app/src/containers/reviewWithdraw/index.tsx
RakeshUP: formMethod setValues is not re-rendering the component. So if we fetch the token price on the review withdrawal screen, we need to add a state here, which seems too much. Let's see if someone complains that the price is not updated!

RakeshUP:
@Fabricevladimir I've made the requested changes. Regarding the footer for fullWidth I too think having a custom footer prop would be a nice way. Let's see what @sepehr2github does in his create DAO review page PR!

@WissenIstNacht I've addresses your comments too

WissenIstNacht (approved):
Thanks a lot for the changes.

There's still a small issue with the links in the WYSYWIG editor, though :D The links specified open as paths rather than external urls (e.g.: localhost:3000/google.com).

I'll approve the PR so you don't get blocked by this.

RakeshUP:
@WissenIstNacht what did you put as url for links to open as relative paths? I am not able to reproduce the bug. 
Could you also check the behavior in the deployment preview? 

UPDATE: Is the bug happening in the TextArea or the Add Resource Component?

Fabricevladimir:
Thanks! It would be nice if you could create a ticket for the link fix. Because abc.com is considered as a valid email, it's appended to the current location. This one is on me 😅  The regular expression needs to require a protocol or we can append one if it's not present.

Fabricevladimir (approved):

PR #243, opened 2022-02-17
Feature/Breadcrumb & AlertBanner  - DAO-699

Fabricevladimir:
- adds Breadcrumb and AlertBanner components with corresponding stories and tests
- adds User Interface Home icon


RakeshUP (approved):

packages/ui-components/stories/breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import {Meta, Story} from '@storybook/react';
+import {Badge, Breadcrumb, BreadcrumbProps} from '../src';
+
+export default {
+  title: 'Components/Breadcrumb',
+  component: Breadcrumb,
+} as Meta;
+
+const Template: Story<BreadcrumbProps> = args => <Breadcrumb {...args} />;
```
WissenIstNacht, line 10: Bonus points: Could you add a story with 3 levels? :)

packages/ui-components/src/components/breadcrumb/breadcrumb.tsx
```diff
@@ -0,0 +1,94 @@
...
+   * include a label and its corresponding path
+   */
+  crumbs: Crumbs;
+
+  /**
+   * Whether breadcrumbs are being displayed in a process flow.
+   * If more than one breadcrumb is given when inside a process flow,
+   * only the first crumb will be shown
+   */
+  process?: boolean;
```
WissenIstNacht, line 27: Ah, while we're on the subject, this would be a great place to apply discriminated unions. :D

You could do something like, such that a library client can not do `process crumbs=[...]` in a component:

```
export type Crumb = {
	label: string
	path: string
}

type ProcessBreadcrumbProps = {
	/**
	 * Array of breadcrumbs to be displayed; each breadcrumb should
	 * include a label and its corresponding path
	 */
	crumbs: Crumb

	/**
	 * Whether breadcrumbs are being displayed in a process flow.
	 * If more than one breadcrumb is given when inside a process flow,
	 * only the first crumb will be shown
	 */
	process: true
}

type NormalBreadcrumbProps = {
	/** List of breadcrumbs */
	crumbs: Crumb[]
	/**
	 * Whether breadcrumbs are being displayed in a process flow.
	 * If more than one breadcrumb is given when inside a process flow,
	 * only the first crumb will be shown
	 */
	process: false
}

export type BreadcrumbProps = {
	/** Tag shown at the end of the list of breadcrumbs */
	tag?: React.FunctionComponentElement<BadgeProps>

	/** Callback returning the path value of the breadcrumb clicked */
	onClick?: (path: string) => void
} & (ProcessBreadcrumbProps | NormalBreadcrumbProps)
```

You might need to make some slight adaptation in you code, though. Don't sweat if it's too much work.

WissenIstNacht (approved):
Awesome, thanks.

I've added two possible improvements (from my POV, anyways :joy:). I'll approve the PR already. Up to you if you want to incorporate them.

packages/ui-components/stories/breadcrumbs.stories.tsx
Fabricevladimir: Sure thing

packages/ui-components/src/components/breadcrumb/breadcrumb.tsx
Fabricevladimir: I like this!

PR #244, opened 2022-02-18
Feature/ListItemAction - DAO-698

Fabricevladimir:
- merges ListItemText with ListItemMenu into ListItemAction
- updates corresponding tests and stories
- updates ListItemText references in web-app to use new ListItemAction


sepehr2github (commented):
Thank you Fabrice but What will happen to the web app ListItemTexts? I think we can't approve this PR without web-app refactor?

RakeshUP (approved):

sepehr2github (approved):
Thank you for new changes!

PR #247, opened 2022-02-21
feat: contract deployment

mathewmeconry:


nivida (changes requested):
LGTM! But on merge into main should it only update the factory contracts and not the Registry cause there in case we have to update it are some further steps required. Also it would be great to have the new addresses of the factories added together with a commit hash to a deployments.md or similar in this repository.

mathewmeconry:
> LGTM! But on merge into main should it only update the factory contracts and not the Registry cause there in case we have to update it are some further steps required. Also it would be great to have the new addresses of the factories added together with a commit hash to a deployments.md or similar in this repository.

done
it updates the `Releases.md` file in `packages/contracts` and updates the `active_contracts.json` file in the root of this repository.
`active_contracts.json` can then be reused for other deployments

nivida (approved):
LGTM! thanks! :))

packages/contracts/scripts/deploy.ts
```diff
@@ -3,28 +3,134 @@
 //
 // When running the script with `npx hardhat run <script>` you'll find the Hardhat
 // Runtime Environment's members available in the global scope.
-import { ethers } from "hardhat";
+import {ethers} from 'hardhat';
```
Rekard0: i am a bit confused. why we use hardhat script when we have hardhat-deploy plugin installed and used. check package.json and /deploy

or i am missing something?

mathewmeconry: valid point. haven't noticed this plugin.
I'll restructure it with the refactoring to support multiple networks properly

Rekard0 (approved):

PR #249, opened 2022-02-21
Add Dropdown from Radix-UI

WissenIstNacht:
Adds the proper dropdown component from Radix UI.

### DIFF
--- a/packages/ui-components/__tests__/dropdown.test.tsx
+++ b/packages/ui-components/__tests__/dropdown.test.tsx
@@ -1,0 +1,17 @@
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+
+import {Dropdown} from '../src/components/dropdown';
+
+describe('Dropdown', () => {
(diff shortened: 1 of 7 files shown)


RakeshUP (approved):

sepehr2github (commented):
Thanks, Vali. Clean and well commented. I'm wondering should the drop-down always be a menu? because you rendered a list and added a click event for each of them!

WissenIstNacht:
Hey @sepehr2github, thanks thanks for the review.

> I'm wondering should the drop-down always be a menu? 

I'm not sure I understand this question. What else could it be? Do you have something else in mind?

> added a click event for each of them!

I don't think there's a "click event" per se. There is a `onSelected` event that is provided by the RadixUI library. The list contains the component you want to be rendered, as well as the callback you want to execute when the `onSelect` is triggered. Does that make sense?

sepehr2github (approved):

PR #253, opened 2022-02-22
Chore/Disabled options in ButtonGroup, Link url fixes DAO-701 DAO-592

RakeshUP:
- Makes few tabs in VotingTerminal disabled
- Adds protocol(`http://` - this is the default behaviour in almost every app and also tiptap editor) to links to prevent opening them as relative paths

### DIFF
--- a/packages/web-app/src/components/addLinks/row.tsx
+++ b/packages/web-app/src/components/addLinks/row.tsx
@@ -12,7 +12,11 @@ import styled from 'styled-components';
 import {useTranslation} from 'react-i18next';
 import {Controller, useFormContext, useFormState} from 'react-hook-form';
 
-import {EMAIL_PATTERN, URL_PATTERN} from 'utils/constants';
+import {
+  EMAIL_PATTERN,
+  URL_PATTERN,
+  URL_WITH_PROTOCOL_PATTERN,
+} from 'utils/constants';
 
 type LinkRowProps = {
   index: number;
@@ -21,7 +25,7 @@ type LinkRowProps = {
 
 const LinkRow: React.FC<LinkRowProps> = ({index, onDelete}) => {
   const {t} = useTranslation();
-  const {control, clearErrors, getValues, trigger} = useFormContext();
+  const {control, clearErrors, getValues, trigger, setValue} = useFormContext();
   const {errors} = useFormState();
 
   /*************************************************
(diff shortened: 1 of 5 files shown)


packages/web-app/src/components/addLinks/row.tsx
```diff
@@ -57,6 +61,28 @@ const LinkRow: React.FC<LinkRowProps> = ({index, onDelete}) => {
...
+  const addProtocolToLinks = useCallback(
+    (name: string) => {
+      const url = getValues(name);
+
+      if (
+        new RegExp(URL_PATTERN).test(url) ||
+        new RegExp(EMAIL_PATTERN).test(url)
+      ) {
+        if (
+          new RegExp(URL_PATTERN).test(url) &&
```
Fabricevladimir: I understand why this is being done. Do you think it's perhaps better to limit the user input to only a url that contains the protocol in the first place so that we don't have to do this regex matching multiple times?

Fabricevladimir (approved):
Thanks for the fixes!

packages/web-app/src/components/addLinks/row.tsx
```diff
@@ -57,6 +61,28 @@ const LinkRow: React.FC<LinkRowProps> = ({index, onDelete}) => {
     [linkedFieldsAreValid, t]
   );
 
+  const addProtocolToLinks = useCallback(
+    (name: string) => {
+      const url = getValues(name);
+
+      if (
+        new RegExp(URL_PATTERN).test(url) ||
```
WissenIstNacht: Could you extract the RegExp objects into variables and reuse them? Creating new ones for each test is kind of (unnecessarily) expensive.

WissenIstNacht (changes requested):
Thanks Rakesh!

packages/web-app/src/components/addLinks/row.tsx
RakeshUP: I did this because all the other apps I checked like Notion, Jira, etc adds http:// to every URL without a protocol. Also this is the default behaviour in our Tiptap editor

packages/web-app/src/components/addLinks/row.tsx
RakeshUP: Fixed it

packages/web-app/src/utils/constants.ts
```diff
@@ -26,7 +26,10 @@ export const enum TransferTypes {
 }
 
 export const URL_PATTERN =
-  /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-z0-9]+([-.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?$/g;
+  /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-z0-9]+([-.]{1}[a-z0-9]+)*\.[a-z]{2,5}(:[0-9]{1,5})?(\/.*)?$/i;
```
WissenIstNacht, line 29: Nice catch. Could have been an annoying bug to find later 👍🏼

WissenIstNacht (approved):

packages/web-app/src/utils/constants.ts
RakeshUP: It already created issues. When I put regex declaration to a constant and reused it, instead of instantiating the regex multiple times, the /g started giving problems and this StackOverflow post saved me. https://stackoverflow.com/questions/2851308/why-does-my-javascript-regex-test-give-alternating-results

PR #256, opened 2022-02-23
Feature/ DAO-696 

sepehr2github:
- add withdraw actions
- create basic actions context

Note: I've tried to create a simple pipeline state structure with proper methods but regarding the best context structure we need a discussion


RakeshUP (approved):
IMO the `Reset Action` button should only clear the states in that particular action. When I have 2 withdraw asset actions and click reset on any, both the form's state is reset. If it makes sense to make another copy of `ConfigureWithdrawForm`, please do. 

But otherwise, this PR is very good. ActionContext 👌👏

packages/web-app/src/containers/actionBuilder/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * This Component is responsible for generating all actions that append to pipeline context (actions)
+ * In future we can add more action template like: mint token Component
+ * or custom action component (for smart contracts methods)
+ * @returns List of actions
+ */
+
+const ActionBuilder: React.FC = () => {
+  const {actions} = useActionsContext();
+
+  const actionsComponents = (name: ActionsTypes, index: number) => {
```
Fabricevladimir: Why not make this an actual component with props and such?

packages/web-app/src/containers/actionBuilder/withdrawAction.tsx
```diff
@@ -0,0 +1,103 @@
...
+import {useFormContext} from 'react-hook-form';
+
+import {useActionsContext} from 'context/actions';
+import ConfigureWithdrawForm from '../configureWithdraw';
+
+type Props = {
+  index: number;
+};
+
+const WithdrawAction: React.FC<Props> = ({index}) => {
```
Fabricevladimir, line 19: Because a global form is being used for multiple withdraw actions, the fields (to, amount, etc.) will all refer to same values, meaning when you have multiple withdraws, a change in one will be produced in the other. A solution might be to make each action its own form, maybe?

packages/web-app/src/containers/addActionMenu/index.tsx
```diff
@@ -4,21 +4,19 @@ import {useTranslation} from 'react-i18next';
...
 import {useGlobalModalContext} from 'context/globalModals';
+import {useActionsContext} from 'context/actions';
 import ModalBottomSheetSwitcher from 'components/modalBottomSheetSwitcher';
 
 export enum AddActionItems {
-  ADD_REMOVE_TOKENS = 'ADD_REMOVE_TOKENS',
-  MINT_TOKENS = 'MINT_TOKENS',
-  WITHDRAW_ASSETS = 'WITHDRAW_ASSETS',
-  EXTERNAL_CONTRACT = 'EXTERNAL_CONTRACT',
+  ADD_REMOVE_TOKENS = 'add_remove_address',
```
Fabricevladimir: Small typo? Did you mean `ADD_REMOVE_ADDRESS`?

packages/web-app/src/containers/addActionMenu/index.tsx
```diff
@@ -4,21 +4,19 @@ import {useTranslation} from 'react-i18next';
 import {ActionListItem, IconChevronRight} from '@aragon/ui-components';
```
Fabricevladimir: This one is on me. This should be `ListItemAction` and not `ActionListItem`. There probably is no difference in actual styling, but the `ActionListItem` will be removed since it's no longer in the Components designs. This confusion was introduced when we merged `ListItemMenu` and `ListItemText` into `ListItemAction`. I had forgotten to remove the already present `ActionListItem` component.

packages/web-app/src/pages/newProposal.tsx
```diff
@@ -1,17 +1,71 @@
...
+import {formatUnits} from 'utils/library';
+import {useWalletProps} from 'containers/walletMenu';
+import {TransferTypes} from 'utils/constants';
+import {BaseTokenInfo} from 'utils/types';
+import {fetchTokenPrice} from 'services/prices';
 
 const NewProposal: React.FC = () => {
   const {t} = useTranslation();
-  const formMethods = useForm();
+  const formMethods = useForm<WithdrawFormData>({
```
Fabricevladimir: Using WithdrawFormData here leads to believe that a New Proposal has the same values as a Withdraw and we also lose the actual New Proposal data (Author, title, summary and the rest). Perhaps this should be moved to the Withdraw assets option.

packages/web-app/src/pages/newProposal.tsx
```diff
@@ -1,17 +1,71 @@
...
+    defaultValues,
+    mode: 'onChange',
+  });
+  const {data: tokens} = useDaoTokens('myDaoAddress');
+  const {account}: useWalletProps = useWallet();
+
+  useEffect(() => {
```
Fabricevladimir, line 29: This is an extension of the previous comment. As it is, once the user hits the New Proposal page, the form is populated with `from` and `type` even if the proposal doesn't include an actual Withdraw action.

packages/web-app/src/pages/newProposal.tsx
```diff
@@ -1,17 +1,71 @@
...
+      formMethods.setValue('from', constants.AddressZero);
+      formMethods.setValue('type', TransferTypes.Withdraw);
+    }
+  }, [account, formMethods]);
+
+  /*************************************************
+   *             Callbacks and Handlers            *
+   *************************************************/
+
+  const handleTokenSelect = (token: BaseTokenInfo) => {
```
Fabricevladimir, line 41: Same here, this is not necessary unless the user actually wants to include a Withdraw action. I understand though that Token Menu requires the handler to be on the same page. If we can move the Token Menu to where the withdraw action is actually taking place, we can remove this from here. Better yet, at some point we need to move this handler into the Token Menu itself. There's a lot of duplication going around with `handleTokenSelect` 😄

Fabricevladimir (changes requested):
Thanks Sepehr! Moving the actions into the context as opposed to leaving them in the New Proposal form context very much simplifies things 👍🏾 
Left a couple of comments about the position of things.

packages/web-app/src/containers/actionBuilder/withdrawAction.tsx
sepehr2github: It's just a placeholder somehow. Now the index is accessible for withdraw form and we can use formArreyField later

packages/web-app/src/pages/newProposal.tsx
sepehr2github: We need to fix this based on actions pipeline structure

Fabricevladimir (approved):

PR #259, opened 2022-02-28
Chore/ DAO-706 

sepehr2github:
Fix blow issues: 
- The wallet list form does not complain when a row is empty.
- The wallet list form does not complain when a row has an amount but no wallet address.
- Deleting a row from the wallet list form does not update the total in the footer.
- Selecting an existing token on the second step leaves the "continue" button disabled.
- The stepper shows a wrong number of steps (this is because there are more pages that steps, since the "intro" and the "go live" page count as part of the stepper as well)
- The logo should be center-aligned with the label "Logo"

### DIFF
--- a/packages/web-app/src/components/fullScreenStepper/fullScreenStepper.tsx
+++ b/packages/web-app/src/components/fullScreenStepper/fullScreenStepper.tsx
@@ -13,7 +13,6 @@ import {StepProps} from './step';
 
 export type FullScreenStepperProps = {
   wizardProcessName: string;
-  totalFormSteps?: number;
   children: React.FunctionComponentElement<StepProps>[];
 };
 
@@ -32,12 +31,10 @@ export const useFormStep = () =>
   useContext(FullScreenStepperContext) as FullScreenStepperContextType;
 
 export const FullScreenStepper: React.FC<FullScreenStepperProps> = ({
-  totalFormSteps,
   wizardProcessName,
   children,
 }) => {
   const {t} = useTranslation();
-
   const {currentStep, prev, next, setStep} = useStepper(children.length);
 
   const currentIndex = currentStep - 1;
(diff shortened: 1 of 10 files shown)


packages/web-app/src/pages/createDAO.tsx
```diff
@@ -66,8 +74,14 @@ const CreateDAO: React.FC = () => {
 
   const daoSetupCommunityIsValid = useMemo(() => {
```
Fabricevladimir, line 75: At some point we should try to extract these validators into a hook or the like

Fabricevladimir (approved):
Thanks!

RakeshUP (approved):

packages/web-app/src/components/fullScreenStepper/fullScreenStepper.tsx
```diff
@@ -32,12 +31,10 @@ export const useFormStep = () =>
   useContext(FullScreenStepperContext) as FullScreenStepperContextType;
 
 export const FullScreenStepper: React.FC<FullScreenStepperProps> = ({
-  totalFormSteps,
```
WissenIstNacht, line 35: I'm not sure you're doing yourself a favour by removing this. It was added by Rakesh to take care of forms that had more pages than the form (for example, in the case where there was an intro page, like on the creation flow).

If you keep this remove, you'll walk into trouble in existing flows like the deposit flow. For example, with your current changes, the stepper starts counting from 0. I.e., the "Configure Deposit" step is "step 0 of 2", but it should be "step 1 of 2".

WissenIstNacht (changes requested):
Hey Sepehr, thanks for the awesome work! Validation really is a pain.

Aside from the issue above with the stepper numbers, there's a few more details I've found. Most of them are really small issues. I've separated them into things that need to be changed, and things that we should keep in mind and maybe address later. 

Necessary changes:

- Wallet list:
  - first wallet address input has “Your Wallet” as value intead of placeholder. This leads to problems like: User can not directly paste. Text doesn't vanish on click. Validation doesn't trigger. Etc.
  - Empty addresses rows are currently allowed (i.e., rows with no address AND no token amount). This is not a problem in the form itself. But on the review page, it will count the empty row as an address. I.e., if you enter one valid address and and empty one, it will say "See 3 addresses" on the review. 
  - There's a typo on the double wallet error. Should be "You can't enter an address twice".  That being said, I would use a different copy: “Address must not be reused” (because in the rest of the app errors never speak to the user directly. Also, it’s not only that the user can’t enter the address twice, but any number of times more than one) 
- Review:
  - Entering super long url in links breaks the layout of DAO Metadata component.
  - Fallback DAO icon not shown in DAO Metadata
- Setup Community
  - Using an existing token, and then switch to new token doesn't fully reset the tokens. For example, if you use ANT and then change to create a new Token, the name and Symbol will be "Aragon Network Token" and "ANT". Also, the wallet list component will show the total supply of ANT.

Optional changes:

- Link List:
    - “Links is mandatory” error appears as soon as user starts typing in title. Maybe should only appear when user leaves the field?
- Wallet list:
    - Rounding percentages (equal distribution among 3 addresses gives 33%, which adds up to 99%)
    - Going back to edit a part of the form does not reset the check (I’m not sure if it should, so take this with a grain of salt)

PR #264, opened 2022-03-02
Fix/Navbar Fixes DAO-737

Fabricevladimir:
- adds navbar into Wizard Create
- updates FullScreenStepper to take process crumb label and path
- updates mobile nav to show network indicator on process paths
- removes (temporarily) options menu

Note: A couple of failing tests are fixed (CardDao, ListItemDao)


RakeshUP (approved):
Just noticed: the ChevronLeft button in breadcrumb to go back, I think it should be an anchor tag with path as href. This just makes it more accessible than navigating on buttonClick. As a matter of fact all links should be a tags instead of button in breadcrumb.

Fabricevladimir:
> Just noticed: the ChevronLeft button in breadcrumb to go back, I think it should be an anchor tag with path as href. This just makes it more accessible than navigating on buttonClick. As a matter of fact all links should be a tags instead of button in breadcrumb.

I agree, yes. The reason why I went with buttons instead is because I didn't want a whole page refresh going back.

sepehr2github (approved):
Thanks Fabrice. 
Screen Shot 2022-03-08 at 11 52 10 AM
the finance page mobile broken and You fixed this issue on other PR am I right?

Except this one looks good to me!

PR #266, opened 2022-03-03
Chore/ DAO-706

sepehr2github:
these comments fixed:
Wallet list:
- first wallet address input has “Your Wallet” as value intead of placeholder. This leads to problems like: User can not directly paste. Text doesn't vanish on click. Validation doesn't trigger. Etc. _NOTE: This is currently according to the designs and was therefore not changed [VR]_
- Empty addresses rows are currently allowed (i.e., rows with no address AND no token amount). This is not a problem in the form itself. But on the review page, it will count the empty row as an address. I.e., if you enter one valid address and and empty one, it will say "See 3 addresses" on the review.
- There's a typo on the double wallet error. Should be "You can't enter an address twice". That being said, I would use a different copy: “Address must not be reused” (because in the rest of the app errors never speak to the user directly. Also, it’s not only that the user can’t enter the address twice, but any number of times more than one)
Review:
- Entering super long url in links breaks the layout of DAO Metadata component.
- Fallback DAO icon not shown in DAO Metadata
Setup Community
- Using an existing token, and then switch to new token doesn't fully reset the tokens. For example, if you use ANT and then change to create a new Token, the name and Symbol will be "Aragon Network Token" and "ANT". Also, the wallet list component will show the total supply of ANT.

Regarding these two:
- “Links is mandatory” error appears as soon as user starts typing in title. Maybe should only appear when user leaves the field? -> It takes longer than I thought so I will fix it later also I think we can ask Selim idea about the right functionality
- Rounding percentages (equal distribution among 3 addresses gives 33%, which adds up to 99%) -> I think It's UX more than technical issue
- Going back to edit a part of the form does not reset the check (I’m not sure if it should, so take this with a grain of salt) -> remember I asked a question regarding this and Ramon told me checkboxes remain


RakeshUP (approved):

Fabricevladimir (commented):
Thanks for the fixes, good work! I noticed a couple of things:

1. I can duplicate my wallet when adding addresses (Setup Community)
2. I can add a wallet without tokens without an issue (Setup Community). Is this proper behavior? 
3. I can add more than 100% Support (Configure Governance)
4. I can add more than 60 hours & 60 minutes for the Duration (Configure Governance)

For 3 & 4, perhaps it's not a flow issue and more of a slight change needed in the input value instead? 🤔

WissenIstNacht:
Thanks, good job man!

There is still a bit of an issue when changing from creating a token to using one that already exists (and vice-versa). For example, when entering ANT and then changing to create a token, it still prefills the name of the token. Also it keeps the total. Another example is if you start by creating a token and then switch to an existing one. If you entere a few addresses, it will tell you " See n Addresses" on the rewview.

Fabricevladimir (approved):
Good work! Once you've addressed the reset on existing vs new token select, good to go!

WissenIstNacht (approved):
Seems to work! 🥳 

One last thing: Please make sure to contact someone on the design side about the "See 2 addresses" when using an existing contract. I still they should come up with an alternate state for this component.

PR #268, opened 2022-03-03
Feature/Page headers - DAO-739

Fabricevladimir:
- adds breadcrumbs to mobile page headers 
- makes breadcrumbs bold

Note: Per the design system, page headers should technically be components. I've embedded them directly into the web-app because I disagree 🥊

### DIFF
--- a/packages/web-app/src/pages/proposal.tsx
+++ b/packages/web-app/src/pages/proposal.tsx
@@ -192,8 +192,7 @@ const Content = styled.div.attrs({
 })``;
 
 const HeaderContainer = styled.div.attrs({
-  className:
-    'flex flex-col gap-y-2 desktop:p-0 px-2 tablet:px-3 pt-2 border border-red-500',
+  className: 'flex flex-col gap-y-2 desktop:p-0 px-2 tablet:px-3 pt-2',
 })``;
 
 const ProposalTitle = styled.p.attrs({
--- a/packages/web-app/src/pages/proposal.tsx
+++ b/packages/web-app/src/pages/proposal.tsx
@@ -89,7 +89,6 @@ const Proposal: React.FC = () => {
             <Link
               external
               label={publisher}
-              // label={shortenAddress(publisherAddress || '')}
               // href={`${explorers[chainId]}${publisherAddress}`}
             />
           </ProposerLink>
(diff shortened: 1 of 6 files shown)


packages/web-app/src/pages/proposal.tsx
```diff
@@ -35,51 +39,79 @@ const stepDataRunning: ProgressStatusProps[] = [
...
 const Proposal: React.FC = () => {
   const [expandedProposal, setExpandedProposal] = useState(false);
   const {t} = useTranslation();
-  const navigate = useNavigate();
   const {id} = useParams();
+  const navigate = useNavigate();
+  const {isDesktop} = useScreen();
+
+  const publisher = useMemo(() => {
+    // if (publisherAddress === account) return 'you';
```
sepehr2github, line 53: are These comments required?

sepehr2github (commented):
Thanks Fabrice! good job man, Just a small thing.
Screen Shot 2022-03-08 at 11 36 20 AM
On the token page when I click on the finance page It won't come back! and Also we should fix this padding issue

packages/web-app/src/pages/proposal.tsx
```diff
@@ -159,18 +191,28 @@ const Content = styled.div.attrs({
   className: 'tablet:w-4/5 mx-auto px-2 pt-3 pb-14',
 })``;
 
-const Header = styled.p.attrs({className: 'font-bold text-ui-800 text-3xl'})``;
+const HeaderContainer = styled.div.attrs({
+  className: 'flex flex-col gap-y-2 desktop:p-0 px-2 tablet:px-3 pt-2',
```
WissenIstNacht, line 195: Detail regarding tablet layout: The Header section is narrower than the content section (ecause of the 3 unit padding in the x direction). Would be nice to have their width aligned. Not sure if you want to remove the padding here or add it to the container. I'll let you figure it out :D

WissenIstNacht (changes requested):
Thanks man, look great!

packages/web-app/src/pages/proposal.tsx
Fabricevladimir: I'll remove them, but they were code to be used once backend is integrated

packages/web-app/src/pages/proposal.tsx
Fabricevladimir: Yup, good catch. For the header component, the padding is correct per ui-components->header->proposal. I didn't modify the page content size because it's an app-wide issue that warrants its own ticket I felt.

WissenIstNacht: Right, all good 👍🏼

WissenIstNacht (approved):

RakeshUP (approved):

PR #269, opened 2022-03-04
resolve #225: use solidity custom errors

### LINKED ISSUE #225: Use Solidity Custom Errors
[CustomErrors are available since solc 0.8.4](https://blog.soliditylang.org/2021/04/21/custom-errors/), are gas-efficient, make the bytecode smaller, and increase readability.

https://github.com/nomiclabs/hardhat/issues/1428 and https://github.com/ethers-io/ethers.js/issues/1493 support it already, while https://github.com/ChainSafe/web3.js/issues/4235 is working on it.
Because of web3.js, https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2839 is pending.

heueristik:
This feature replaces errors strings by custom errors, as described in #225 in our code.

E.g.,
`string private constant ERROR_CAN_NOT_EXECUTE = "VOTING_CAN_NOT_EXECUTE";`
is replaced by
`error VoteExecutionForbidden(uint256 voteId);`.

Then,
`require(_canExecute(_voteId), ERROR_CAN_NOT_EXECUTE);`
becomes
`if(!_canExecute(_voteId)) revert VoteExecutionForbidden(_voteId);`.

### DIFF
--- a/packages/contracts/contracts/core/DAO.sol
+++ b/packages/contracts/contracts/core/DAO.sol
@@ -30,14 +30,20 @@ contract DAO is IDAO, Initializable, UUPSUpgradeable, ACL, ERC1271, AdaptiveERC1
     bytes32 public constant SET_SIGNATURE_VALIDATOR_ROLE = keccak256("SET_SIGNATURE_VALIDATOR_ROLE");
 
     // Error msg's
-    string internal constant ERROR_ACTION_CALL_FAILED = "ACTION_CALL_FAILED";
-    string internal constant ERROR_WITHDRAW_RECIPIENT_ZERO = "WITHDRAW_RECIPIENT_ZERO";
-    string internal constant ERROR_DEPOSIT_AMOUNT_ZERO = "DEPOSIT_AMOUNT_ZERO";
-    string internal constant ERROR_WITHDRAW_AMOUNT_ZERO = "WITHDRAW_AMOUNT_ZERO";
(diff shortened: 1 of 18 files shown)


novaknole:
Thanks for this, Definatelly needed.

A couple of points:

1. There're still `requires` remaining in our contracts. For example - in ACL.sol. Better to search for all `require` and update it the same way.  
2. As a common way, the `error` in the contracts should be put above `contract` keyword, not in the contract itself.  An example - https://docs.soliditylang.org/en/v0.8.12/structure-of-a-contract.html#errors

heueristik:
> Thanks for this, Definatelly needed.
> 
> A couple of points:
> 
> 1. There're still `requires` remaining in our contracts. For example - in ACL.sol. Better to search for all `require` and update it the same way.
> 2. As a common way, the `error` in the contracts should be put above `contract` keyword, not in the contract itself.  An example - https://docs.soliditylang.org/en/v0.8.12/structure-of-a-contract.html#errors

Good points, I will make these changes.

nivida (changes requested):
LGTM! By side of the missing custom errors in those files: 

- ACL
- AdaptiveERC165
- Permissions
- Registry
- MerkleDistributor

I would keep the errors within the contract closure as they are. I saw both "patterns" used when looking at other projects and I personally think it is way cleaner to have it within the closure.

nivida (approved):
LGTM

novaknole (approved):

PR #273, opened 2022-03-07
fix withdraw flow validation issues

josemarinas:


sepehr2github (commented):
Thank you, Jose! Well Done!
Just one minor thing. On Withdraw flow page two I selected Date + Time and I changed the values then I went to next page but I forgot to do sth. when I came back to page two the end tab was on days not Date + Time

packages/web-app/src/containers/configureWithdraw/index.tsx
```diff
@@ -97,6 +97,13 @@ const ConfigureWithdrawForm: React.FC = () => {
     trigger,
   ]);
 
+  // Manually trigger the amount validation when the coin is changed
+  useEffect(() => {
+    if (symbol !== '') {
+      trigger('amount');
```
Fabricevladimir, line 103: While this triggers for the amount properly, we might not want to do that on the very first render. Since the amount field isn't yet dirty, we'll be showing an error the second the user selects a token. Thoughts?

Fabricevladimir (commented):
Thank you! This was an annoying bug

packages/web-app/src/containers/setupVotingForm/index.tsx
```diff
@@ -137,37 +140,15 @@ const SetupVotingForm: React.FC = () => {
 
     // check start constraints
     if (startMills < currMills) {
-      setError('startTime', {
```
WissenIstNacht, line 140: So there's a couple of issues with letting react hook form assign errors by returning them from the validator function:
- If you change the start date to be after 5 days before the end date (e.g., you land on the page and increase the start date by one day), the error will appear on the start date. This doesn't really make sense.
- Also, the errors are not properly cleared. For example, after the above scenario, increasing the enddate by at least one does not reset the error.

// Edit:
Thinking back to our call on the issue, I had an idea. You said that returning `true` was solving some of the issues you observed. Based on that, I tried the following, which seems to work. First, I reverted back to setting/clearing errors "manually". Then, I added variables that track whether errors were set. If that's the case, I return `''` (basically, like before). If not, however, I now return `true`. Below is the code for the entire validator function, including the changes I just mentioned. Could you test it out and let me know what you think? Thanks a lot 🙏🏼 

```ts
const dateTimeValidator = () => {
    //build start date/time in utc mills
    const sDate = getValues('startDate');
    const sTime = getValues('startTime');
    const sUtc = getValues('startUtc');

    const canonicalSUtc = getCanonicalUtcOffset(sUtc);
    const startDateTime = toDate(sDate + 'T' + sTime + canonicalSUtc);
    const startMills = startDateTime.valueOf();

    const currDateTime = new Date();
    const currMills = currDateTime.getTime();

    //build end date/time in utc mills
    const eDate = getValues('endDate');
    const eTime = getValues('endTime');
    const eUtc = getValues('endUtc');
    const canonicalEUtc = getCanonicalUtcOffset(eUtc);
    const endDateTime = toDate(eDate + 'T' + eTime + canonicalEUtc);
    const endMills = endDateTime.valueOf();

    const minEndDateTimeMills = startMills + daysToMills(5);

    let hasStartError = false;
    let hasEndError = false;

    // check start constraints
    if (startMills < currMills) {
      setError('startTime', {
        type: 'validate',
        message: t('errors.startPast'),
      });
      setError('startDate', {
        type: 'validate',
        message: t('errors.startPast'),
      });
      hasStartError = true;
    }
    if (startMills >= currMills) {
      clearErrors('startDate');
      clearErrors('startTime');
    }

    //check end constraints
    if (endMills < minEndDateTimeMills) {
      setError('endTime', {
        type: 'validate',
        message: t('errors.endPast'),
      });
      setError('endDate', {
        type: 'validate',
        message: t('errors.endPast'),
      });
      hasEndError = true;
    }
    if (endMills >= minEndDateTimeMills) {
      clearErrors('endDate');
      clearErrors('endTime');
    }

    return (hasStartError || hasEndError) ? '' : true;
  };
```

WissenIstNacht (changes requested):
Hey there, thanks a lot for the PR :)

Unfurtonately I think your code has solved a few things, but broken others 😅

josemarinas:
> Thank you, Jose! Well Done! Just one minor thing. On Withdraw flow page two I selected Date + Time and I changed the values then I went to next page but I forgot to do sth. when I came back to page two the end tab was on days not Date + Time

Yes, that happens because of the initial state of the field. The first solution that comes to my mind is using a React hook with state for the whole withdrawal process, and start managing the forms in this way, but maybe it is a little overkill, let me know what you think

packages/web-app/src/containers/configureWithdraw/index.tsx
josemarinas: Yes, you are right, it should be solved by adding a check to know if the field is dirty in the if. Thank you for pointing it out

packages/web-app/src/containers/setupVotingForm/index.tsx
josemarinas: Sure, it makes sense, I'll try it and let you know

WissenIstNacht (approved):
Thanks, the validation issue seems to work 👍🏼 

Regarding @sepehr2github 's comment: I will open a new ticket on this issue and deal with it on a separate PR.

Fabricevladimir:
There's a small issue with the amount validation. Validation for amount isn't triggering when the contract address changes. It can be reproduced by following these steps:
1. In token menu, select add custom token
2. Insert incorrect contract address & some value for the amount (just so the field is dirty)
3. Back in token menu, select one of the DAO tokens

I'm not sure if we want to fix this here or merge and resolve the issue in a separate ticket.

WissenIstNacht:
Closing the PR as it's already been merged. We'll have another ticket(s) to tackle pending issues across the various forms.

PR #275, opened 2022-03-08
Feature/ DAO-685 Wallet Network switch

sepehr2github:
Logic Part Implemented!


sepehr2github:
Closed due to new decision for using web3modal instead of use wallet

PR #276, opened 2022-03-08
Fix: Dao 692 arbitrum login

WissenIstNacht:
This PR allows the app to connect to Arbitrum.

### DIFF
--- a/packages/web-app/src/hooks/useEnsData.tsx
+++ b/packages/web-app/src/hooks/useEnsData.tsx
@@ -1,0 +1,65 @@
+import {Address} from '@aragon/ui-components/src/utils/addresses';
+import {useProviders} from 'context/providers';
+import {useState, useEffect} from 'react';
+import {HookData} from 'utils/types';
+
+export function useEnsName(address: Address): HookData<string> {
+  const {web3} = useProviders();
+  const [name, setName] = useState('');
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState<Error | undefined>();
+
+  useEffect(() => {
+    async function fetchEnsData() {
+      setIsLoading(true);
+
+      if (web3 && address) {
+        try {
+          let res = await web3.lookupAddress(address);
+          if (!res) res = '';
+          setName(res);
+        } catch (error) {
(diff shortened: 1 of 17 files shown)


packages/web-app/src/hooks/useWalletTokens.tsx
```diff
@@ -1,27 +1,105 @@
...
-      if (account === null) {
+      setIsLoading(true);
+      if (account === null || provider === null) {
         setWalletTokens([]);
         return;
       }
 
-      const tokenList = await getTokenList();
       if (Number(balance) !== -1 && Number(balance) !== 0)
         await tokenList.push(constants.AddressZero);
```
Fabricevladimir: We can remove the await there, yes?

packages/web-app/src/hooks/useEnsData.tsx
```diff
@@ -0,0 +1,65 @@
...
+      }
+      setIsLoading(false);
+    }
+
+    fetchEnsData();
+  }, [web3, address]);
+  return {data: name, isLoading, error};
+}
+
+export function useEnsAvatar(address: Address): HookData<string> {
```
Fabricevladimir, line 36: What is the reasoning behind not merging these two into one considering most of the time we'll want both?

Fabricevladimir (approved):
Thanks for this. Cleaned up things really well

packages/web-app/src/hooks/useEnsData.tsx
WissenIstNacht: Right, good question. So there's two things that kinda go hand-in-hand here. The first is, as you said, we wont always use both. But more importantly, having an ens record does not imply having an Avatar. In which case you'll only want the name. Now the second, related reason is data and error handling. I think having them separated makes it simpler to use (although there's a slight overhead of characters) when taking into account things like errors. E.g., what do you return when one call succeeds ant the other fails? Which error do you return when both fail?

Fabricevladimir: Makes sense 👍🏾

RakeshUP (approved):

PR #284, opened 2022-03-14
Feature/DAO switcher & abstain vote option 

RakeshUP:
DAO switcher first pass(DAO-715) & abstain voting option added(DAO-745)

The DAO switcher is very basic and is missing out on functionality. We can work on it further after we have some real data and also decide the parameters in urls

### DIFF
--- a/packages/ui-components/src/components/table/votersTable.tsx
+++ b/packages/ui-components/src/components/table/votersTable.tsx
@@ -7,7 +7,7 @@ import {Link} from '../link';
 
 export type VoterType = {
   wallet: string;
-  option: 'Yes' | 'No';
+  option: 'Yes' | 'Abstain' | 'No';
   votingPower: string;
   tokenAmount: string;
 };
@@ -38,7 +38,13 @@ export const VotersTable: React.FC<VotersTableProps> = ({
             <TableCell type="tag">
               <Badge
                 label={voter.option}
-                colorScheme={voter.option === 'Yes' ? 'success' : 'critical'}
+                colorScheme={
+                  voter.option === 'Yes'
+                    ? 'success'
+                    : voter.option === 'No'
+                    ? 'critical'
+                    : 'neutral'
+                }
               />
             </TableCell>
             <TableCell type="text" text={voter.votingPower} rightAligned />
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/table/votersTable.tsx
```diff
@@ -38,7 +38,13 @@ export const VotersTable: React.FC<VotersTableProps> = ({
             <TableCell type="tag">
               <Badge
                 label={voter.option}
-                colorScheme={voter.option === 'Yes' ? 'success' : 'critical'}
+                colorScheme={
```
Fabricevladimir: Might want to extract this into a variable.

packages/web-app/src/containers/navbar/daoSelectMenu.tsx
```diff
@@ -7,26 +7,22 @@ import {
...
 
 // NOTE: the state setting is temporary until backend integration
 const DaoSelectMenu: React.FC = () => {
   const {t} = useTranslation();
-  const [selectedDao, setSelectedDao] = useState(DAOS[0]);
+  const selectedDao = useReactiveVar(selectedDAO);
+  const favoriteDaos = useReactiveVar(favoriteDAOs);
   const {isSelectDaoOpen, close} = useGlobalModalContext();
 
   const handleDaoSelect = (dao: {daoName: string; daoAddress: string}) => {
```
Fabricevladimir, line 24: If a dao is only the name and address, how do we know which network a dao belongs to?

Fabricevladimir (commented):
Thanks! The Apollo client local storage variable mechanism is very handy

WissenIstNacht (changes requested):
Thanks Rakesh. I think Fabrice's comments already cover everything. PIng me when they're resolved.

packages/web-app/src/containers/navbar/daoSelectMenu.tsx
RakeshUP: I am still not sure of the data structure for the favoriteDAOs array. I just copied it from the existing temp data in Navbar DaoSelector. Things like ENS name for DAO might be added too along with network

Fabricevladimir: I guess that's also going to depend on how we figure out which network a DAO is in as well. Fair enough.

Fabricevladimir (approved):

WissenIstNacht (approved):

PR #287, opened 2022-03-16
Chore/ Review proposal step & buttonGroup fix DAO-767, DAO-771

RakeshUP:


Fabricevladimir (commented):
Thanks for adding that final step. I did notice a couple of things though: 
- a user is still able to navigate to the next step even when the current one is invalid
- In the review step, the `CardExecution` data isn't wired up properly (perhaps because actions is an array this time?)

WissenIstNacht (approved):
Thanks, LGTM!

WissenIstNacht:
About @Fabricevladimir 's comments

>     * a user is still able to navigate to the next step even when the current one is invalid

This occurred to me, too. I feel like there's a couple of people and PRs that address (parts of) this at the moment. I suggest we merge all the open PRs that touches any of this and evaluate what's missing. What do you guys think?

>     * In the review step, the `CardExecution` data isn't wired up properly (perhaps because actions is an array this time?)

Nice catch, this is actually a more serious problem. Just to make sure I understand the issue correctly: If we want to keep things as they are RN, we'd have to "artificially" put the data from the withdrawal flow into an array as this is what the components expect from the proposal flow?

Fabricevladimir:
@WissenIstNacht 
Concerning the validation, I don't know if there is a PR that takes care of that specifically, but yes it might be a better idea to get everything merged before we do anything about it.

As for the withdraw flow data, the component is expecting an `actions` array of, er, _actions_, so the withdraw info must be restructured to match. There is a PR about the withdraw validation (#273), so perhaps both these can be merged and then update the structure afterwards. Thoughts?

Fabricevladimir (approved):

PR #288, opened 2022-03-17
Feature/ Dao-754 

sepehr2github:
- contributors address modal added
- refactored voters table (please review this and let me know if this approach is best or not)


packages/web-app/src/containers/communityAddressesModal/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        const re = new RegExp(searchValue, 'i');
+        return wallets?.address?.match(re);
+      }
+      return true;
+    },
+    [searchValue]
+  );
+
+  useEffect(() => {
+    const list = wallets
```
Fabricevladimir: Would this be best memoized instead of calculated inside useEffect and stored as state?

packages/web-app/src/containers/communityAddressesModal/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <ModalHeader>
+        <SearchInput
+          value={searchValue}
+          onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+            setSearchValue(e.target.value)
+          }
+        />
+      </ModalHeader>
+      <Container>
+        <VotersTable
```
Fabricevladimir: Can you take care of an empty state for when no address matches the search query?

Fabricevladimir (changes requested):
Thanks, nice work! Left a couple of comments

RakeshUP (approved):

Fabricevladimir (approved):

PR #289, opened 2022-03-17
Feature/Main Finance Page & All tokens page - DAO-753 DAO-752 

Fabricevladimir:
This PR refactors existing finance flow
- fetches dao balances from subgraph
- fetches token metadata and prices for balances from coingecko api
- calculates token share for dao vault
- displays tokens on main finance page
- displays tokens on all tokens page

Note: This PR only modifies part of the existing pipeline, but the intention is to gradually replace it; broke the task into multiple smaller chunks.  

Also, for demonstration purposes, the pipeline is hardcoded to a specific dao on a test subgraph deployment


RakeshUP (approved):

packages/web-app/src/utils/types.ts
```diff
@@ -53,6 +53,41 @@ export type TokenBalance = {
   address: Address;
   count: bigint;
 };
+//===========================================================
+export type DaoTokenBalance = {
+  id: number;
+  token: {
+    id: string;
```
WissenIstNacht, line 59: Is id the same as address or is it something like an API identifier for Goingecko?

packages/web-app/src/pages/home.tsx
```diff
@@ -1,15 +1,20 @@
...
+import {useDaoVault} from 'hooks/useDaoVault';
+
 const Home: React.FC = () => {
   const {t} = useTranslation();
-  const navigate = useNavigate();
   const {open} = useGlobalModalContext();
+  const navigate = useNavigate();
+
+  const vault = useDaoVault('0x79fde96a6182adbd9ca4a803ba26f65a893fbf4f');
+  console.log('vault', vault);
```
WissenIstNacht: necessary?

packages/web-app/src/hooks/useDaoVault.tsx
```diff
@@ -0,0 +1,48 @@
...
+ * Hook encapsulating the logic for fetching the assets from the DAO vault and mapping them
+ * to their corresponding USD market values.
+ * @param daoAddress Dao address
+ * @param options.filter TimeFilter for market data
+ * @param options.interval Delay in milliseconds
+ * @returns A list of tokens in the DAO treasury, current USD sum value of all assets,
+ * and the price change in USD based on the filter
+ */
+export const useDaoVault = (daoAddress: string, options?: PollTokenOptions) => {
+  const {data: balances, getDaoBalances} = useDaoBalances(daoAddress);
```
WissenIstNacht: Wait, if this (or rather the apollo hook inside `usDaoBalances`) gives you data, why, also explicitly have the getter function?

packages/web-app/src/hooks/useDaoVault.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const {data: tokensWithMetadata} = useTokenMetadata(balances);
+  const {data} = usePollTokenPrices(tokensWithMetadata, options);
+
+  useEffect(() => {
+    getDaoBalances();
```
WissenIstNacht: This looks suspiciously like an infinite loop. Can you guarantee that calling `getDaoBalance` will not change itself?

packages/web-app/src/hooks/useDaoVault.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }, [getDaoBalances]);
+
+  const tokens = useMemo(
```
WissenIstNacht: Would this work the same way? I.e., would token hold the same data for any path of execution of the program?

```
const tokens = data?.tokens.map(token => ({
      ...token,
      ...(token.marketData?.treasuryShare
        ? {
            treasurySharePercentage:
              (token.marketData.treasuryShare / data.totalAssetValue) * 100,
          }
        : {}),
    })) as VaultToken[]
```

The reason I'm asking is because you seem to rely on memo to store state, which (generally) is in an anti-pattern. The more appropriate way to do it would be to use a state and effect hook-combo. I might be wrong here, it's hard to judge just from mental code execution. In any case, the critical question to know which way to go is the one I asked above. If the hooks works exactly the same (ignoring performance) without `useMemo` you're ok.

packages/web-app/src/hooks/usePollTokenPrices.tsx
```diff
@@ -0,0 +1,112 @@
...
+      }
+    } catch (error) {
+      setError(error as Error);
+      console.error(error);
+    } finally {
+      setIsLoading(false);
+    }
+  }, [isMounted, tokenList]);
+
+  const pricing = useMemo((): {
```
WissenIstNacht: Here same question as in `useDaoVault`. Except this time I'm pretty sure the code would work differently without `useMemo` (which indicates that it should not be used).

In case you're worried about performance here: Keep in mind that `useEffect` has this built in as well. It will not run unnecessarily as it will only run when the dependencies change.

WissenIstNacht (commented):
Hey there. thanks for the re-organization and integration of the apollo services.

I left a couple of questions I'd love to hear your opinion on :)

packages/web-app/src/utils/types.ts
Fabricevladimir: `id` on line 58 is coming from the subgraph, I forgot to remove it here, but it is unnecessary here. `id` on line 60 is the same as token address. I tried aliasing it in the gql query but for some reason it wouldn't let me even though I know it should technically be possible to do so

packages/web-app/src/hooks/useDaoVault.tsx
Fabricevladimir: With this one, I was playing around with the idea of having the `DaoVault` in a context considering Finance, All Tokens, and All Transfers use the same data. For All Transfers especially, I found it almost wasteful having to do all the calls and calculations just to get the Total Volume. In the end though, I couldn't find a good enough solution, so I kept things the same. Thoughts on this one?

packages/web-app/src/pages/home.tsx
Fabricevladimir: Forgot this one!

packages/web-app/src/hooks/useDaoVault.tsx
Fabricevladimir: Hehehe, might have been actually. This one is the lazy query from Apollo client, so I had relative certainty (is that a thing?) about it not randomly changing itself. But I have removed this entirely. Thanks for catching it!

WissenIstNacht (approved):
Thanks, good to go! 🚀

PR #294, opened 2022-03-24
Chore/Update token menu - DAO-775 

Fabricevladimir:
- updates token types
- updates token menu to use new types
- completes switch to new pipeline (continuation of #289)


RakeshUP (approved):
On adding a custom token, the Max Balance becomes 4242.4242424. Also, I get "not erc20 token" error for almost all custom tokens I add. I am not sure if that logic is correct. We could merge this and check that separately I guess

packages/web-app/src/containers/configureWithdraw/index.tsx
```diff
@@ -16,16 +17,19 @@ import React, {useCallback, useEffect} from 'react';
...
+import {fetchBalance, getTokenInfo, isETH} from 'utils/tokens';
 import {
   validateAddress,
   validateTokenAddress,
   validateTokenAmount,
 } from 'utils/validators';
-import {useApolloClient} from 'context/apolloClient';
+
+// TO BE REMOVED during integration
+const DAOVaultAddress = '0x79fde96a6182adbd9ca4a803ba26f65a893fbf4f';
```
sepehr2github, line 32: It would be awesome if we can set a constant and set the dao address there

sepehr2github (commented):
Thanks a lot. Fabrice
Just a tiny thing. UI components test didn't pass.

packages/web-app/src/containers/configureWithdraw/index.tsx
Fabricevladimir: Yes, I haven't really done so because the assumption is that the `daoAddress` will be either gotten from the subgraph or from the url as a param.

sepehr2github (approved):
Thanks Alot

PR #306, opened 2022-04-04
Feature: Wallet based DAO

josemarinas:
## Description

Enabled the wallet based DAO flow and added the required changes for it to work. The list of wallets is stored in the `useForm` with the key `walletList`. A component was added to `ui-components` and the `Label` component was modified to accept HTML

Task: [DAO-572](https://aragonassociation.atlassian.net/browse/DAO-572)

## Type of change

## Checklist:

### DIFF
--- a/packages/ui-components/src/components/alerts/alertCard.tsx
+++ b/packages/ui-components/src/components/alerts/alertCard.tsx
@@ -1,0 +1,71 @@
+import React from 'react';
+import styled from 'styled-components';
+
+import {IconInfo, IconSuccess, IconWarning} from '../icons';
+
+const styles = {
+  info: {
+    icon: <IconInfo className="text-info-500" height={20} width={20} />,
+    color: 'text-info-800',
+    borderColor: 'border-info-400',
+  },
+  success: {
+    icon: <IconSuccess className="text-success-500" height={20} width={20} />,
+    color: 'text-success-800',
+    borderColor: 'border-success-400',
+  },
+  warning: {
+    icon: <IconWarning className="text-warning-500" height={20} width={20} />,
+    color: 'text-warning-800',
+    borderColor: 'border-warning-400',
+  },
+  critical: {
+    icon: <IconWarning className="text-critical-500" size={20} height={20} />,
+    color: 'text-critical-800',
+    borderColor: 'border-critical-400',
+  },
+};
+
+export type AlertCardProps = {
+  /** type and severity of alert */
+  mode?: 'info' | 'success' | 'warning' | 'critical';
+  /** card title */
+  title: string;
+  /** card subtitle */
+  helpText?: string;
+};
+
+export const AlertCard: React.FC<AlertCardProps> = ({
+  mode = 'info',
+  title,
+  helpText,
+}) => {
+  return (
+    <Container mode={mode}>
+      <div>{styles[mode].icon}</div>
+      <TextContainer>
+        <Title mode={mode}>{title}</Title>
+        <HelpText>{helpText}</HelpText>
+      </TextContainer>
+    </Container>
+  );
+};
+
+type ContainerProps = {
+  mode: NonNullable<AlertCardProps['mode']>;
+};
+const Container = styled.div.attrs(({mode}: ContainerProps) => ({
+  className: `flex gap-1.5 box-border px-2 py-1.5 border-2 rounded-xl bg-ui-0 ${styles[mode].borderColor}`,
+}))<ContainerProps>``;
+
+const TextContainer = styled.div.attrs(() => ({
+  className: 'flex flex-col gap-0.25 ',
+}))``;
+
+const Title = styled.div.attrs(({mode}: ContainerProps) => ({
+  className: `flex flex-col font-bold  ${styles[mode].color}`,
+}))<ContainerProps>``;
+
+const HelpText = styled.div.attrs(() => ({
+  className: 'text-ui-600 text-sm',
+}))``;
--- a/packages/ui-components/src/components/label/label.tsx
+++ b/packages/ui-components/src/components/label/label.tsx
@@ -7,6 +7,7 @@ export type LabelProps = {
   helpText?: string;
   isOptional?: boolean;
   badgeLabel?: string;
+  renderHtml?: boolean;
 };
 
 export const Label: React.FC<LabelProps> = ({
@@ -14,14 +15,21 @@ export const Label: React.FC<LabelProps> = ({
   helpText,
   isOptional = false,
   badgeLabel,
+  renderHtml = false,
 }) => {
   return (
     <VStack data-testid="label">
       <LabelLine>
-        <Heading>{label}</Heading>
+        {renderHtml?
+          <Heading dangerouslySetInnerHTML={{__html: label}} />:
+          <Heading>{label}</Heading>
+        }
         {isOptional && <Badge label={badgeLabel || 'Optional'} />}
       </LabelLine>
-      {helpText && <HelpText>{helpText}</HelpText>}
+        {renderHtml && helpText?
+          <HelpText dangerouslySetInnerHTML={{__html: helpText}} />:
+          <HelpText>{helpText}</HelpText>
+        }
     </VStack>
   );
 };
@@ -36,8 +44,17 @@ const LabelLine = styled.div.attrs({
 
 const Heading = styled.p.attrs({
   className: 'font-bold text-ui-800',
-})``;
+})`
+  & > a {
+    color: #003bf5;
+  }
+`;
 
 const HelpText = styled.p.attrs({
   className: 'text-sm font-normal text-ui-600',
-})``;
+})`
+  & > a {
+    color: #003bf5;
+    font-weight: 700;
+  }
+`;
--- a/packages/web-app/src/containers/configureCommunity/index.tsx
+++ b/packages/web-app/src/containers/configureCommunity/index.tsx
@@ -1,5 +1,5 @@
 import {AlertInline, Label, NumberInput} from '@aragon/ui-components';
-import React from 'react';
+import React, {useMemo} from 'react';
 import styled from 'styled-components';
 import {useTranslation} from 'react-i18next';
 import {Controller, useFormContext, useWatch} from 'react-hook-form';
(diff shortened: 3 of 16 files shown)


packages/web-app/src/containers/goLive/community.tsx
```diff
@@ -33,8 +33,12 @@ const Community: React.FC = () => {
     isCustomToken,
     tokenSymbol,
     tokenTotalSupply,
+    walletList,
   } = getValues();
-
+  const handleSeeAllWallets = () => {
+    alert('not implemented');
```
brickpop, line 39: Why is that alert and TO DO here?

josemarinas: As far as I know, there is a different ticket to implement the modal with a list of addresses so once this ticket is resolved his function should be updated

packages/ui-components/src/components/label/label.tsx
```diff
@@ -7,21 +7,27 @@ export type LabelProps = {
...
   isOptional = false,
   badgeLabel,
+  renderHtml = false,
 }) => {
   return (
     <VStack data-testid="label">
       <LabelLine>
-        <Heading>{label}</Heading>
+        {renderHtml && <Heading dangerouslySetInnerHTML={{__html: label}} />}
+        {!renderHtml && <Heading>{label}</Heading>}
```
Fabricevladimir: You may be able to simplify (!) this by using the conditional operator.

packages/web-app/src/components/walletList/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+import React, {useEffect} from 'react';
+import {useFormContext, useWatch} from 'react-hook-form';
+import styled from 'styled-components';
+import {useWallet} from 'use-wallet';
+import {Row} from './row';
+
+export const WalletList = () => {
+  const {account} = useWallet();
+  const {setValue} = useFormContext();
+  const walletArray: string[] = useWatch({
```
Fabricevladimir: react-hook-form provides a `useFieldArray` hook to simplify this type of listing of fields. Checkout the `addWallets` or `addLinks` components to take a look

packages/web-app/src/components/walletList/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  useEffect(() => {
+    setValue('walletList.0', account);
+  }, [account, setValue]);
+  return (
+    <Container>
+      <TableContainer>
+        <Header>{t('labels.walletList.address')}</Header>
+        {walletArray &&
+          walletArray.map((_, index) => (
+            <>
```
Fabricevladimir: Might want to wrap this in a div so you can pass it a key. React will complain about it not having a unique key prop

packages/web-app/src/components/walletList/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+          ))}
+        <Divider />
+        <Footer>
+          {t('labels.walletList.addresses', {count: walletArray?.length || 0})}
+        </Footer>
+      </TableContainer>
+      <ActionsContainer>
+        <TextButtonsContainer>
+          <ButtonText
+            label="Add address"
```
Fabricevladimir: Can you use the translation strings for the label here?

packages/web-app/src/components/walletList/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+            onClick={handleAddWallet}
+          />
+          <ButtonText
+            label="Upload CSV"
+            mode="ghost"
+            size="large"
+            onClick={() => alert('upload CSV here')}
+          />
+        </TextButtonsContainer>
+        <Popover
```
Fabricevladimir: There is a pending task to replace the popovers in the application with the dropdown component from ui-components. Could you use that instead here?

packages/web-app/src/components/walletList/row.tsx
```diff
@@ -0,0 +1,119 @@
...
+type WalletListRowProps = {
+  index: number;
+};
+
+export const Row = ({index}: WalletListRowProps) => {
+  const {control, setValue} = useFormContext();
+  // TODO update with useSigner
+  const {account} = useWallet();
+  const walletList = useWatch({name: 'walletList'});
+  const handleDeleteWallet = () => {
```
Fabricevladimir: Blank line between the methods for readability purposes please 😄

packages/web-app/src/components/walletList/row.tsx
```diff
@@ -0,0 +1,119 @@
...
+                );
+              }}
+              mode="default"
+              placeholder="0x..."
+              adornmentText={value ? 'Copy' : 'Paste'}
+              disabled={value === account}
+              onAdornmentClick={() => handleClipboardActions(value, onChange)}
+            />
+            {error?.message && (
+              // <ErrorContainer>
```
Fabricevladimir: Don't forget to delete if comment is unnecessary

packages/web-app/src/components/walletList/row.tsx
```diff
@@ -0,0 +1,119 @@
...
+                  <ListItemAction
+                    title={t('labels.removeWallet')}
+                    onClick={handleDeleteWallet}
+                    bgWhite
+                  />
+                )}
+                {value === account && (
+                  <ListItemAction
+                    title={t('labels.deleteAllAddresses')}
+                    onClick={handleDeleteAllAddresses}
```
Fabricevladimir: Are you certain this is supposed to delete all addresses and not just the corresponding address? It seems a bit counter-intuitive.

packages/web-app/src/components/walletList/row.tsx
```diff
@@ -0,0 +1,119 @@
...
+              disabled={value === account}
+              onAdornmentClick={() => handleClipboardActions(value, onChange)}
+            />
+            {error?.message && (
+              // <ErrorContainer>
+              <AlertInline label={error.message} mode="critical" />
+              // </ErrorContainer>
+            )}
+          </InputContainer>
+          <Popover
```
Fabricevladimir: Would be great if you could use a dropdown here as well instead. If it takes too much time, move on and we'll implement the dropdown in a separate task

packages/web-app/src/components/walletList/row.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    bgWhite
+                  />
+                )}
+              </div>
+            }
+          >
+            <ButtonIcon
+              mode="ghost"
+              size="large"
+              disabled={walletList.length <= 2}
```
Fabricevladimir: Can you clarify why this button is disabled?

packages/web-app/src/containers/configureCommunity/index.tsx
```diff
@@ -18,50 +18,107 @@ const ConfigureCommunity: React.FC = () => {
...
+                      placeholder={t('placeHolders.daoName')}
+                      percentage={true}
+                    />
+                  </FormWrapper>
+                  <AlertInline
+                    label={t(
+                      'createDAO.step4.alerts.minimumParticipationAlert',
+                      {
+                        percentage:
+                          Math.round(
```
Fabricevladimir: Would it be more efficient to memoize this up top?

packages/ui-components/src/components/alerts/alertCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  critical: {
+    icon: <IconWarning className="text-critical-500" height={20} width={20} />,
+    color: 'text-critical-800',
+    borderColor: 'border-critical-400',
+  },
+};
+
+export type AlertCardProps = {
+  /** type and severity of alert */
+  mode?: 'neutral' | 'success' | 'warning' | 'critical';
```
Fabricevladimir: Shouldn't neutral be info?

packages/ui-components/src/components/alerts/alertCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+  /** card title */
+  title: string;
+  /** card subtitle */
+  subtitle?: string;
```
Fabricevladimir: Can you rename this to helptext to align with the designs unless you feel the design should change instead?

packages/web-app/src/components/walletList/row.tsx
```diff
@@ -0,0 +1,119 @@
...
+      walletList.forEach((wallet: string, walletIndex: number) => {
+        if (address === wallet && index !== walletIndex) {
+          validationResult = t('errors.duplicateAddress') as string;
+        }
+      });
+    }
+    return validationResult;
+  };
+  return (
+    <Controller
```
Fabricevladimir: When there's an error it pushes down the menu button.

image

packages/ui-components/src/components/alerts/alertCard.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const AlertCard: React.FC<AlertCardProps> = ({
+  mode = 'neutral',
+  title,
+  subtitle,
+}) => {
+  return (
+    <Container mode={mode}>
+      {styles[mode].icon}
+      <TextContainer>
```
Fabricevladimir, line 46: The text forces the icon to shrink on mobile

image

Fabricevladimir (changes requested):
Good work, thanks, very involved task! Left a few comments.

packages/web-app/src/components/walletList/row.tsx
josemarinas: It does not make much sense to allow the user to delete its own address from this list, but I'm waiting for selim to tell me about how it should be done

packages/ui-components/src/components/alerts/alertCard.tsx
josemarinas: Maybe, I'm not sure, I used the naming convention that is used in the `AlertInline` component, but if you prefer `info` I can change it to info

packages/ui-components/src/components/alerts/alertCard.tsx
josemarinas: Hmm, I checked this, but it seems like a bigger issue with the icon components in general, if you do the same test with the `AlertInline` component it happens the exact same thing.

packages/ui-components/src/components/label/label.tsx
josemarinas: Fixed in the last commit

packages/web-app/src/components/walletList/index.tsx
josemarinas: It took me a while to understand how `useFieldArray` worked, but I believe right now is working properly with this hook

packages/web-app/src/components/walletList/index.tsx
josemarinas: Updated the code with dropdown instead of popovers

packages/web-app/src/components/walletList/row.tsx
josemarinas: I talked with selim, and he updated the design with the new options. I implemented the in them new commit

packages/web-app/src/containers/configureCommunity/index.tsx
josemarinas: Yes, makes sense, is not a super complex operation but makes sense to have it in a different place just for readability, so I updated the code with an `useMemo` :smile:

packages/ui-components/src/components/alerts/alertCard.tsx
Fabricevladimir: Info would be sweet! For the naming convention in general always go with what's on figma (unless you have a reason not, in that case ping one of the designers to discuss it), because sometimes the existing components have been refactored on figma but the changes are not yet reflected in the code itself

packages/ui-components/src/components/alerts/alertCard.tsx
Fabricevladimir: Ah, good catch. If you wrap a div around the icon like so `<div>{styles[mode].icon}</div>` that should fix it I believe. It would be great if you could do that for the AlertInline as well, that one is on me 🙈

sepehr2github (approved):

Fabricevladimir (approved):
Thanks for the changes!

PR #307, opened 2022-04-04
Feature: Create DAO

josemarinas:
## Description

Added `useCache`, `useClient` and `useDao` hooks. 

- `useClient`: hook with a context to create a connection with the SDK
- `useCache`: hook that stores the data requested from the `useClient` with a unique identifier to avoid making the exact same request 2 times
- `useDao`: used to do actions related with a DAO, for now  just createDao, but the idea is to add all the related methods here, it uses both the `useDao` and `useCache` hooks to return a result.

Task: [DAO-721](https://aragonassociation.atlassian.net/browse/DAO-721)

## Type of change

## Checklist:


packages/web-app/src/containers/goLive/index.tsx
```diff
@@ -56,16 +58,69 @@ const GoLive: React.FC = () => {
...
+   */
+  const getMinDuration = (days: number, hours: number, minutes: number): number => {
+    return minutes * 60 + hours * 3600 + days * 86400
+  }
+  const handleGoLive = () => {
+    const minutes = parseInt(getValues('minutes'))
+    const hours = parseInt(getValues('hours'))
+    const days = parseInt(getValues('days'))
+    const wallets: WalletField[] = getValues('wallets')
+    const createDaoForm: CreateDaoForm = {
```
marcvelmer: This is already deprecated. I'm going to send you the new version (which sadly it's still not published)

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export interface Dao {
+  id: string
+  name: string
+  creator: string
+  metadata: string
+  token: string
+}
+export interface CreateDaoForm {
+  id?: string
```
marcvelmer: What means the `id`?

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,94 @@
+import { ethers } from "ethers"
+import { isAddress } from "ethers/lib/utils"
+import { useEffect, useState } from "react"
+import { useCache } from "./useCache"
+import { useClient } from "./useClient"
+
+export interface Dao {
```
marcvelmer: Is that what the UI is expecting the SDK to return? If so, we should document this and achieve and agreement of which data will be returned and why

packages/web-app/src/hooks/useCachedDao.tsx
marcvelmer: Ok I saw it. I suppose it's for the cache. So, my question now is... Why should we use the cache for a creating DAO event?

packages/web-app/package.json
```diff
@@ -17,6 +17,8 @@
   },
   "dependencies": {
     "@apollo/client": "^3.5.8",
+    "@aragon/sdk-client": "file:./aragon-sdk-client-v0.0.1.tgz",
```
marcvelmer: I suppose this is here because it's just a draft PR

packages/web-app/src/containers/goLive/index.tsx
josemarinas: Ok, perfect

packages/web-app/src/hooks/useCachedDao.tsx
josemarinas: Yes, the `id` will be equivalent to the address or hash that will be generated after the creation of the DAO and yes it is also used for the cache, I'm not sure that in the case of creating a DAO make absolute sense but in the end you have all the data from a DAO (daoConfig, tokenConfig and votingConfig)

packages/web-app/src/hooks/useCachedDao.tsx
josemarinas: Yes, we should agree in the format of this, this interface is just a draft, I will change it once we have a common interface

packages/web-app/package.json
josemarinas: Yes, once is published it should be just the version of the sdk-client

packages/web-app/src/hooks/useClient.tsx
```diff
@@ -0,0 +1,50 @@
...
+    // useEffect(() => {
+    //   provider?.getNetwork().then((network) => {
+    //     setClient({ ...client, chainId, chainName: network.name })
+    //   })
+    // }, [chainId])
+  const value = {
+    client
+  }
+  return (
+    <UseClientContext.Provider value={value}>
```
brickpop, line 64: If we only return the `client`, there may be no need to wrap it in an object `{}`, we could return the client itself.

packages/web-app/src/hooks/useCachedDao.tsx
marcvelmer: Could we get this agreement somewhere written? What is the UI expecting to receive when creating a DAO? Currently it just returns the DAO address. Is this OK?

packages/web-app/src/hooks/useClient.tsx
josemarinas: I'm not sure, because the SDK right now has different classes depending on what  you are doing, for example `ClientDaoERC20Voting` and `ClientDaoWhitelistVoting`

brickpop: Agree. Indeed, I suggested Fabrice to potentially reuse this hook to return the Apollo GraphQL client (we're bypassing the SDK in this case)

packages/web-app/src/components/addWallets/index.tsx
```diff
@@ -42,8 +47,16 @@ const AddWallets: React.FC = () => {
...
+  useEffect(() => {
+    if (account) {
+      update(1, {address: account, amount: '0'});
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
   return (
     <Container data-testid="add-wallets">
+      {JSON.stringify(fields)}
```
brickpop: Is this temporary?

packages/web-app/src/hooks/useClient.tsx
```diff
@@ -0,0 +1,60 @@
...
+    useState<ClientDaoWhitelistVoting>();
+
+  useEffect(() => {
+    if (signer) {
+      const context = new SdkContext({
+        network: chainId,
+        web3Providers: [
+          'https://eth-rinkeby.alchemyapi.io/v2/bgIqe2NxazpzsjfmVmhj3aS3j_HZ9mpr',
+        ],
+        daoFactoryAddress: '0xa0b2B729DE73cd22406d3D5A31816985c04A7cdD',
```
brickpop: Shouldn't this be parameterized, and not hardcoded?

packages/web-app/src/index.tsx
```diff
@@ -9,41 +9,48 @@ import {WalletMenuProvider} from 'context/walletMenu';
 import {GlobalModalsProvider} from 'context/globalModals';
 import {ApolloClientProvider} from 'context/apolloClient';
 import {ProvidersProvider} from 'context/providers';
-import {TransactionsProvider} from 'context/transactions';
+// import { TransactionsProvider } from 'context/transactions';
```
brickpop: Will we need this later on?

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
...
+            setCache(cacheKey, cacheDao);
+            resolve(address);
+          })
+          .catch((e: Error) => {
+            console.log(e);
+            // thow error if it fails
+            reject(e);
+          });
+      } else {
+        reject('ERC20 SDK client is not initialized correctly');
```
brickpop: Promise rejections should return an `Error` object, not just a string
```suggestion
        reject(new Error('ERC20 SDK client is not initialized correctly'));
```

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
...
+            // this will be used to identify the dao in the cache
+            const cacheKey = `dao-${address}`;
+            const cacheDao: Dao = {
+              address,
+            };
+            setCache(cacheKey, cacheDao);
+            resolve(address);
+          })
+          .catch((e: Error) => {
+            console.log(e);
```
brickpop: Is this temporary?

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
...
+            };
+            setCache(cacheKey, cachedDao);
+            resolve(address);
+          })
+          .catch((e: Error) => {
+            // thow error if it fails
+            reject(e);
+          });
+      } else {
+        reject('Whitelist SDK client is not initialized correctly');
```
brickpop:
```suggestion
        reject(new Error('Whitelist SDK client is not initialized correctly'));
```

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
...
+            reject(e);
+          });
+      } else {
+        reject('ERC20 SDK client is not initialized correctly');
+      }
+    });
+  };
+  const createWhitelist = (dao: ICreateDaoWhitelistVoting): Promise<string> => {
+    return new Promise((resolve, reject): void => {
+      console.log(whitelist);
```
brickpop: Is this temporary?

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
...
+            resolve(address);
+          })
+          .catch((e: Error) => {
+            // thow error if it fails
+            reject(e);
+          });
+      } else {
+        reject('Whitelist SDK client is not initialized correctly');
+      }
+    });
```
brickpop: We don't need to create a new Promise just for this. Instead, we could make sure to always return an existing one:

```ts
if (bad) {
   return Promise.reject(new Error("Blah"));
}

return client.dao.create(...).then(...)  // Promise
```

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
...
+    // client.getDao(address).then((value) => setCache(cacheKey, value)).catch((e)=> throw Error(e))
+    const value: Dao = {
+      address,
+    };
+    setCache(cacheKey, value);
+    return value;
+  };
+  const createErc20 = (dao: ICreateDaoERC20Voting): Promise<string> => {
+    return new Promise((resolve, reject): void => {
+      if (erc20) {
```
brickpop: We don't need to create a new Promise just for this. Instead, we could make sure to always return an existing one:

```ts
if (bad) {
   return Promise.reject(new Error("Blah"));
}

return client.dao.create(...).then(...)  // Promise
```

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
+import {isAddress} from 'ethers/lib/utils';
+import {useCache} from './useCache';
+import {useClient} from './useClient';
+import {
+  ICreateDaoERC20Voting,
+  ICreateDaoWhitelistVoting,
+} from '@aragon/sdk-client';
+import {Dao} from 'utils/types';
+
+interface useDaoResponse {
```
brickpop: By convention, TS interfaces begin with a capital `i` and are camelcase
```suggestion
interface IUseDaoResponse {
```

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,102 @@
...
+  update: (dao: Dao) => Dao;
+  createErc20: (dao: ICreateDaoERC20Voting) => Promise<string>;
+  createWhitelist: (dao: ICreateDaoWhitelistVoting) => Promise<string>;
+  get: (address: string) => Dao;
+}
+
+export const useDao = (): useDaoResponse => {
+  const {get: getCache, set: setCache} = useCache();
+  const {erc20, whitelist} = useClient();
```
brickpop: For clarity, I would name them `erc20Client`, `whitelistClient`. However, this will just be a transitory fix

packages/web-app/src/context/transactions.tsx
```diff
@@ -42,6 +54,121 @@ const TransactionsProvider: React.FC<Props> = ({children}) => {
...
+          ),
+        },
+        gsnForwarder: constants.AddressZero,
+      };
+      // call create whitelist function
+      createWhitelist(createDaoForm)
+        .then((address: string) => {
+          // if success set state to success
+          setTransactionState(TransactionState.SUCCESS);
+          console.log(address);
```
brickpop, line 161: Temporary?

packages/web-app/src/context/transactions.tsx
```diff
@@ -42,6 +54,121 @@ const TransactionsProvider: React.FC<Props> = ({children}) => {
...
+        })
+        .catch(e => {
+          // if error set state to error to allow retry
+          console.error(e);
```
brickpop, line 166: Temporary?

packages/web-app/src/hooks/useCache.tsx
```diff
@@ -0,0 +1,41 @@
...
+// eslint is disabled to avoid the warning on the use of any
+// this cache is not the final implementation, and when
+// the final implementation is done the any should be updated with generics
+/* eslint-disable @typescript-eslint/no-explicit-any */
+interface CacheContext {
+  set: (key: string, value: any) => void;
+  get: (key: string) => any;
+}
+
+const UseCacheContext = createContext<CacheContext>({} as CacheContext);
```
brickpop: Shouldn't the default value be a `Map` instead of an object `{}`?

packages/web-app/src/context/transactions.tsx
```diff
@@ -42,6 +54,121 @@ const TransactionsProvider: React.FC<Props> = ({children}) => {
...
+    if (membership === 'token') {
+      // if membership is token get token data from form
+      const {isCustomToken, tokenAddress, tokenName, tokenSymbol, wallets} =
+        getValues();
+      const mintConfig = wallets
+        .filter((wallet: WalletField) => {
+          if (isAddress(wallet.address)) {
+            return true;
+          }
+          return false;
```
brickpop:
```suggestion
          return isAddress(wallet.address);
```

packages/web-app/src/components/whitelistWallets/row.tsx
```diff
@@ -29,19 +29,20 @@ export const Row = ({index}: WhitelistWalletsRowProps) => {
...
-  const addressValidator = (address: string, addressIndex: number) => {
-    let validationResult = validateAddress(address);
-    const wallets = whitelistWallets;
-    if (wallets) {
-      wallets.forEach((wallet: WhitelistWallet, walletIndex: number) => {
-        if (address === wallet.address && addressIndex !== walletIndex) {
-          validationResult = t('errors.duplicateAddress') as string;
+  const addressValidator = (address: string, index: number) => {
+    let validationResult =
+      address === 'My Wallet' ? true : validateAddress(address);
```
brickpop: Is this literal hidden by the UI form controls, or does the user explicitly have to write/keep `"My  Wallet"` ?

brickpop:
Besides the review, could you please check this warning here?
https://sonarcloud.io/project/issues?resolved=false&types=BUG&pullRequest=307&id=aragon_zaragoza&open=AYBCkP2Qm85nxQxBl4rO

Should probably look something like `Wallet | {}`, because otherwise the two types will never intersect.

sepehr2github:
> Besides the review, could you please check this warning here? https://sonarcloud.io/project/issues?resolved=false&types=BUG&pullRequest=307&id=aragon_zaragoza&open=AYBCkP2Qm85nxQxBl4rO
> 
> Should probably look something like `Wallet | {}`, because otherwise the two types will never intersect.

Already Fixed on develop branch

packages/web-app/src/components/addWallets/index.tsx
sepehr2github: Already Fixed on other PR

packages/web-app/src/hooks/useClient.tsx
josemarinas: Yes, forgot to add the env variables

packages/web-app/src/hooks/useCachedDao.tsx
josemarinas: Something like this?
```ts

    if (!erc20) {
      return Promise.reject(new Error('ERC20 SDK client is not initialized correctly'));
    }
    return erc20.dao.create(dao).then((address: string) => {
      const cacheKey = `dao-${address}`;
      const cacheDao: Dao = {
        address,
      };
      setCache(cacheKey, cacheDao);
      return Promise.resolve(address);
    }).catch((e: Error) => {
      return Promise.reject(e);
    })
```

packages/web-app/src/components/whitelistWallets/row.tsx
josemarinas: No, if you write the connected walled address the UI will automatically show "My Wallet"

packages/web-app/src/context/transactions.tsx
josemarinas: Yes this one is there on purpose and is temporary until we have the next step after creating a dao to at least know the address of the created dao, so you can use it to check in subgraph for example

packages/web-app/src/context/transactions.tsx
josemarinas: Yes it already shows an error message, but I don't know if this message should change depending on this error

packages/web-app/src/hooks/useCache.tsx
josemarinas: I dont think so, this references `CacheContext` which is the interface with the `set` and `get` methods

josemarinas:
I updated the branch addressing all the comments @brickpop

.github/workflows/webapp-push.yml
```diff
@@ -42,6 +42,9 @@ jobs:
           VITE_REACT_APP_DEPLOY_ENVIRONMENT: ${{ needs.set-environment.outputs.environment }}
           VITE_REACT_APP_ANALYTICS_KEY: ${{ secrets.VITE_REACT_APP_ANALYTICS_KEY }}
           VITE_REACT_APP_RPC: "mainnet.eth.aragon.network"
+          VITE_REACT_APP_SDK_DAO_FACTORY_ADDRESS: "0xa0b2B729DE73cd22406d3D5A31816985c04A7cdD"
```
brickpop, line 47: Shouldn't this be scoped by network?

packages/web-app/src/components/addWallets/index.tsx
```diff
@@ -42,6 +47,13 @@ const AddWallets: React.FC = () => {
     }, 50);
   };
 
+  useEffect(() => {
+    if (account) {
+      update(1, {address: account, amount: '0'});
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
brickpop, line 56: Shouldn't we follow any changes to `accounts`? or we care just once?

packages/web-app/src/components/whitelistWallets/index.tsx
```diff
@@ -32,7 +32,8 @@ export const WhitelistWallets = () => {
     if (account) {
       update(0, {address: account});
     }
-  });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
brickpop, line 39: Same as above

packages/web-app/src/components/whitelistWallets/row.tsx
brickpop: mmm ok, I wonder if there would be a cleaner way to handle this on the data layer, which doesn't get into the way of Human Readable strings (anything we change in the future can break this, translations, etc)

packages/web-app/src/context/transactions.tsx
```diff
@@ -42,6 +55,129 @@ const TransactionsProvider: React.FC<Props> = ({children}) => {
     [transaction]
   );
 
+  const getMinDuration = (
```
brickpop: Couldn't this be a pure function defined in an `util.ts` file or similar?
We're redeclaring this function every time the component is rendered, and there is no state involved

packages/web-app/src/context/transactions.tsx
brickpop: Well, my point is that adding console.log's is not a user feature, but a developer helper. If more and more console.logs are added, it will stop being useful to us because there will be hundreds of messages around

packages/web-app/src/hooks/useCache.tsx
```diff
@@ -0,0 +1,41 @@
...
+const UseCacheContext = createContext<ICacheContext>({} as ICacheContext);
+
+export const useCache = () => {
+  const ctx = useContext(UseCacheContext);
+  if (ctx === null) {
+    throw new Error(
+      'useCache() can only be used on the descendants of <UseCacheProvider />'
+    );
+  }
+  return ctx;
```
brickpop:
```suggestion
  const cache = useContext(UseCacheContext);
  if (cache === null) {
    throw new Error(
      'useCache() can only be used on the descendants of <UseCacheProvider />'
    );
  }
  return cache;
```

Silly minor edit

brickpop:
```suggestion
  const value = useContext(UseCacheContext);
  if (value === null) {
    throw new Error(
      'useCache() can only be used on the descendants of <UseCacheProvider />'
    );
  }
  return value;
```

Silly minor edit

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -0,0 +1,92 @@
...
+    if (!isAddress(address)) {
+      throw Error('invalid address when trying to get dao');
+    }
+    const cacheKey = `dao-${address}`;
+    const cacheValue = getCache(cacheKey);
+    // If dao is in cache return it
+    if (cacheValue) {
+      return cacheValue;
+    }
+    // client.getDao(address).then((value) => setCache(cacheKey, value)).catch((e)=> throw Error(e))
```
brickpop: Will it be useful in the future? (the commented line)

packages/web-app/src/hooks/useClient.tsx
```diff
@@ -0,0 +1,66 @@
...
+const UseClientContext = createContext<ClientContext>({} as ClientContext);
+
+export const useClient = () => {
+  const ctx = useContext(UseClientContext);
+  if (ctx === null) {
+    throw new Error(
+      'useClient() can only be used on the descendants of <UseClientProvider />'
+    );
+  }
+  return ctx;
```
brickpop:
```suggestion
  const client = useContext(UseClientContext);
  if (client === null) {
    throw new Error(
      'useClient() can only be used on the descendants of <UseClientProvider />'
    );
  }
  return client;
```

Silly minor edit

packages/web-app/src/components/whitelistWallets/row.tsx
josemarinas: Yes, maybe we should change this with a `t('labels.myWallet')` for translations to be helpful, but then I'll have to review the code at some points because there are some points where this condition `address==='My Wallet'` is checked

packages/web-app/src/components/addWallets/index.tsx
josemarinas: We care just the first time, after that the field is disabled and if you add the dependency it will update the value to the same each time the number of addresses change. Besides, I'm not sure that it will watch for changes in inner properties, but if it does, you won't be able to update the amount property.

packages/web-app/src/context/transactions.tsx
josemarinas: Yes, this must be moved to the `utils`, did it there and forgot to move it

packages/web-app/src/context/transactions.tsx
josemarinas: Yes, for now for me is relevant, because the integration with the contracts is still giving some errors at some points, but when this is totally debugged and working this must be removed, same with the `console.log` showing the address when you create a dao

packages/web-app/src/hooks/useCachedDao.tsx
josemarinas: I'm not sure, this is part of a `get by address` function that I left there as a placeholder when first implemented the`useCachedDao` hook. There is also an `update` placeholder, I can get rid of both of them if you want to clean the hook

packages/web-app/src/context/transactions.tsx
brickpop: Then, let's do it there and import it here :)

packages/web-app/src/hooks/useCachedDao.tsx
brickpop: I'd remove it, if this is so trivial to work out when needed.

BTW, you cannot `throw Error(...)`, you should either `throw new Error("text here")` or else `thow e`. But then, you don't need the `catch` block:
```suggestion
    // client.getDao(address).then((value) => setCache(cacheKey, value)).catch((e)=> throw e)
```

brickpop (approved):

sepehr2github (approved):
Thanks, Jose. Great Job. There are still some tasks left. We will take care of those on other Prs.

PR #310, opened 2022-04-07
Feature: Implement Web3modal

josemarinas:
## Description

- Change `useWallet` for `useSigner` hook
- Add new disconnect modal
- Add configuration in vite to support the web3 modal

Task: [DAO-789]()

## Type of change

## Checklist:


packages/web-app/src/app.tsx
```diff
@@ -62,9 +62,10 @@ function App() {
           </Routes>
         </Suspense>
       </Layout>
-      <WalletMenu />
+      {/* <WalletMenu /> */}
```
sepehr2github: Remove This commented menu please

packages/web-app/vite.config.ts
```diff
@@ -22,10 +25,26 @@ export default defineConfig(({mode}) => {
   return {
     base: "",
     plugins: [
+      // ↓ Needed for development mode
```
sepehr2github, line 27: please use prettier to fix format issues here

packages/web-app/src/containers/navbar/walletMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+  return (
+    <ModalBottomSheetSwitcher
+      onClose={() => close('wallet')}
+      isOpen={isWalletOpen}
+    >
+      <ModalHeader>
+        <AvatarAddressContainer>
+          <Avatar src={ensAvatarUrl || address || ''} size="small" />
+          <AddressContainer>
+            {ensName && <Title>{ensName}</Title>}
```
sepehr2github, line 44: I couldn't see my ens name on the wallet Modal. It was OK on the wallet button but here Nothing showed for me! if you don't have any ensName it please let me know. I can test for you 👍

packages/web-app/src/containers/navbar/index.tsx
```diff
@@ -52,22 +52,31 @@ const getNetworkStatus = (id: number) => {
...
 
   const status = useMemo(() => {
-    return isConnected() ? getNetworkStatus(chainId!) : 'default';
+    return isConnected ? getNetworkStatus(chainId!) : 'default';
   }, [chainId, isConnected]);
 
   const handleWalletButtonClick = () => {
-    isConnected() ? open() : connect('injected');
+    if (isConnected) {
+      console.log(isWalletOpen);
```
sepehr2github: I'm not sure! but this console is unnecessary, please remove it.

packages/web-app/package.json
```diff
@@ -37,9 +37,11 @@
     "react-hook-form": "^7.22.2",
     "react-i18next": "^11.12.0",
     "react-router-dom": "^6.0.2",
+    "rollup-plugin-polyfill-node": "^0.9.0",
```
sepehr2github, line 40: May I ask you to give me some information regarding adding this component?

sepehr2github, line 40: May I ask you to give me some information regarding this component?

sepehr2github (commented):
Thanks a lot, Jose! Nice kick-off 👍 . You did a great job, Just left a few comments for you, my friend!

packages/web-app/src/hooks/useWallet.tsx
```diff
@@ -0,0 +1,84 @@
+import {useSigner} from 'use-signer';
+import {JsonRpcSigner, Web3Provider} from '@ethersproject/providers';
+import {useEffect, useState} from 'react';
+import {BigNumber} from 'ethers';
+import {Network} from '@ethersproject/networks';
+
+export interface IUseWalletProps {
```
WissenIstNacht: Could you reference the original interface here instead of copying all the fields into a new type? Something like

```
interface UseWalletProps extends SignerValue {
  //whatever is in IUseWallet but not in SignerValue. 
}
```

Obviously the logic extends to any other fields that may be borrowed from other interfaces/types :)

packages/web-app/src/hooks/useWallet.tsx
```diff
@@ -0,0 +1,84 @@
...
+    if (address && provider) {
+      provider?.getBalance(address).then((balance: BigNumber) => {
+        setBalance(balance);
+      });
+    }
+  }, [provider, address]);
+
+  // Update ENS name and avatar
+  useEffect(() => {
+    if (provider && address) {
```
WissenIstNacht, line 35: Were you aware that there already exist a hook for this? If so, why did you chose to re-implement it here? If not, could you try to merge the code?

packages/web-app/src/hooks/useWallet.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const {chainId, methods, signer, provider, address, status} = useSigner();
+  const [balance, setBalance] = useState<BigNumber | null>(null);
+  const [ensName, setEnsName] = useState<string>('');
+  const [ensAvatarUrl, setEnsAvatarUrl] = useState<string>('');
+  const [networkName, setNetworkName] = useState<string>('');
+  const [isConnected, setIsConnected] = useState<boolean>(false);
+
+  // Update balance
+  useEffect(() => {
+    if (address && provider) {
```
WissenIstNacht, line 26: Were you aware that there already exist a hook for this? If so, why did you chose to re-implement it here? If not, could you try to merge the code?

packages/web-app/src/hooks/useWalletTokens.tsx
```diff
@@ -116,7 +115,7 @@ export function useWalletTokens(): HookData<TokenBalance[]> {
         tokenList.map(address => {
           if (isETH(address)) {
             return [
-              balance,
+              balance ? balance.toString() : '',
```
WissenIstNacht, line 118: Due to your removing of the type when destructuring `useWallet`, balance can now be a `BigNumber`. Have you made sure this does not break things in the UI (in other words does `toString` of a `BigNumber` result in the same string value as before?)

packages/web-app/src/containers/navbar/walletMenu.tsx
WissenIstNacht: I think this is because the `useEnsName` is still tied to `useWallet` (I mean the package here, not the new hook :yum:) and wasn't updated to work with the new wallet mechanism. I left a comment in the `useWallet` (the hook, this time) to address this.

packages/web-app/src/containers/navbar/walletMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+} from '@aragon/ui-components';
+import {useGlobalModalContext} from 'context/globalModals';
+import styled from 'styled-components';
+import {useSigner} from 'use-signer';
+import {useEnsAvatar, useEnsName} from 'hooks/useEnsData';
+import ModalBottomSheetSwitcher from 'components/modalBottomSheetSwitcher';
+import {shortenAddress} from '@aragon/ui-components/src/utils/addresses';
+import {handleClipboardActions} from 'utils/library';
+import useScreen from 'hooks/useScreen';
+export const WalletMenu = () => {
```
WissenIstNacht, line 19: Could you leave a line of space here and there? :D

(At least around functions would be nice)

packages/web-app/src/containers/navbar/desktop.tsx
```diff
@@ -46,8 +45,8 @@ const DesktopNav: React.FC<DesktopNavProp> = props => {
   const navigate = useNavigate();
   const selectedDao = useReactiveVar(selectedDAO);
   const [showCrumbMenu, setShowCrumbMenu] = useState(false);
-  const {isConnected, account, ensName, ensAvatarUrl}: useWalletProps =
-    useWallet();
+  const {status, address, ensName, ensAvatarUrl} = useWallet();
+  const isConnected = status === 'connected';
```
WissenIstNacht: Isn't the `isConnect` value already computed in `useWallet`?

WissenIstNacht (changes requested):
Hey Jose, thanks a ton for all the work.

Most of the of comments I left where on minor issues that need to be changed. However, there are two more important comments around already existing hooks (ens and balance). It would be great if we could settle on just one piece of code :)

packages/web-app/src/app.tsx
brickpop: +1

packages/web-app/src/containers/navbar/index.tsx
brickpop: +1

packages/web-app/src/containers/navbar/index.tsx
```diff
@@ -52,22 +52,31 @@ const getNetworkStatus = (id: number) => {
...
+      open('wallet');
+      return;
+    }
+    methods.selectWallet().catch((err: Error) => {
+      // TODO: maybe add an error message when
+      // the error is different from closing the window
+      console.error(err);
```
brickpop, line 77: Can you please add a Jira task for handling this error on the UI side?

packages/web-app/src/containers/navbar/walletMenu.tsx
brickpop, line 19:
```suggestion

export const WalletMenu = () => {
```

packages/web-app/src/containers/navbar/walletMenu.tsx
brickpop: Have a look here @josemarinas: 
- `provider.lookupAddress(address) => ensDomain (possibly empty)`
- `provider.resolveName(ensDomain) => address`


https://docs.ethers.io/v5/api/providers/provider/#Provider-ResolveName

packages/web-app/src/containers/navbar/walletMenu.tsx
```diff
@@ -0,0 +1,109 @@
...
+          />
+        )}
+      </ModalHeader>
+      <ModalBody>
+        <StyledButtonText
+          size="large"
+          mode="ghost"
+          iconLeft={<IconSwitch />}
+          label="View Transactions"
+          onClick={() => alert('not implemented')}
```
brickpop, line 71: Can we create a task or identity the task that will implement this?

packages/web-app/src/index.tsx
```diff
@@ -10,22 +10,36 @@ import {GlobalModalsProvider} from 'context/globalModals';
...
 import {ProvidersProvider} from 'context/providers';
+import {UseSignerProvider} from 'use-signer';
+import {IProviderOptions} from 'web3modal';
+import WalletConnectProvider from '@walletconnect/web3-provider';
+
+const providerOptions: IProviderOptions = {
+  walletconnect: {
+    package: WalletConnectProvider,
+    options: {
+      infuraId: 'mainnet.eth.aragon.network',
```
brickpop: Is this out API key? Shouldn't this be fetched from `process.env` or from the settings somehow?

brickpop (commented):
Thank you for your contribution Jose ✨

packages/web-app/vite.config.ts
josemarinas: When I run `yarn lint` doesn't seem to have a problem over here

packages/web-app/package.json
josemarinas: The web3modal uses some libraries from nodejs, so this is just a polyfill to make `global` and other variables available. Without this the web3modal won't work

packages/web-app/src/hooks/useWallet.tsx
josemarinas: Yes, the thing is that `SignerValue` is not exported in the package, so I can't use it. @brickpop can you update the use signer and export the `SignerValue` type?

packages/web-app/src/hooks/useWalletTokens.tsx
josemarinas: Yes, I tested with my wallet in the finance page, the only one where `useWalletTokens` is used, and it works

packages/web-app/src/hooks/useWallet.tsx
josemarinas: Yes, but since they are not used in other parts of the code, for me, it makes sense having them in the new `useWallet`, but if you prefer to have them separated we can leave them separated, but I'll have to adapt them because they use the `useProviders` hook that uses the old `useWallet`

packages/web-app/src/containers/navbar/walletMenu.tsx
josemarinas: That's what I'm doing inside the new `useWallet` hook, so the ensName should appear, but I don't have an ENS domain, so I can't test it properly

packages/web-app/src/hooks/useWallet.tsx
WissenIstNacht: I think I remembered something wrong here. I don't think there's a hook for the balance. Sorry 😅

packages/web-app/src/hooks/useWallet.tsx
WissenIstNacht: No problem. If you think it makes more sense to keep them together, go for it. 😊 

The only thing I would ask, then, is to delete the file with the separated hooks (`useEnsData.tsx`) and make sure nothing in the code base references them anymore.

WissenIstNacht (changes requested):
Hey, thanks a lot for addressing the issues. The only thing missing is the formatting on the the vite config file pointed out by @sepehr2github.

If you're having issues with this, let me know.

// Edit: I just noticed your reply further up. You are right, es-lint does not pick up on this. The issue is that the respective linting command is scoped to `src`, which doesn't include config files.

packages/web-app/src/hooks/useWallet.tsx
josemarinas: Sure, it is updated in this new commit

packages/web-app/src/containers/navbar/index.tsx
josemarinas: https://aragonassociation.atlassian.net/browse/APP-249

Here it goes, feel free to update it

packages/web-app/src/containers/navbar/walletMenu.tsx
josemarinas: Yes, first we will have to come into an agreement of how these transactions are going to be handled. In the sense of if a transaction is pending and the integration of this with the client and SDK.

josemarinas:
> Hey, thanks a lot for addressing the issues. The only thing missing is the formatting on the the vite config file pointed out by @sepehr2github.
> 
> If you're having issues with this, let me know.
> 
> // Edit: I just noticed your reply further up. You are right, es-lint does not pick up on this. The issue is that the respective linting command is scoped to `src`, which doesn't include config files.

Ok, that makes sense, I made a new commit fixing the linting of the `vite.config.js`

packages/web-app/src/hooks/useWallet.tsx
josemarinas: Jordi updated the `use-signer` paackage so now we can inport `SignerValue`

WissenIstNacht (approved):
Thanks a lot for all the changes. LGTM!

packages/web-app/src/hooks/useWallet.tsx
```diff
@@ -0,0 +1,73 @@
...
+      provider?.getAvatar(address).then((avatarUrl: string | null) => {
+        avatarUrl ? setEnsAvatarUrl(avatarUrl) : setEnsAvatarUrl('');
+      });
+    }
+  }, [address, provider]);
+
+  // update isConnected
+  useEffect(() => {
+    setIsConnected(status === 'connected');
+  }, [status]);
```
brickpop, line 48: Couldn't this be as simple as having `const isConnected = status === 'connected';` ?
We're creating a state that is 100% computed, but this is very minor to be any stopper.

brickpop (approved):

PR #313, opened 2022-04-07
Feature: APP-80 Network URL

WissenIstNacht:
## Description

Please include a summary of the change and be sure you follow the contributions rules we do provide [here](./CONTRIBUTIONS.md)

Task: [80](https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/41?modal=detail&selectedIssue=APP-80)

## Type of change

## Checklist:


packages/web-app/src/containers/navbar/index.tsx
```diff
@@ -45,26 +45,30 @@ const processes: StringIndexed = {
...
   const processName = useMemo(() => {
     const results = matchRoutes(processPaths, pathname);
     if (results) return results[0].route.path;
   }, [pathname]);
 
+  // NOTE: Since the wallet is no longer the determining factor for the app's
+  // network, this logic needs to be reconsidered. Currently, the app can no
+  // longer be on an "unsupported network" (the user would be redirected to a
+  // "not found" page instead). So the status currently really just shows
+  // whether the app operates on a testnet or on a mainnet ("default").
```
Fabricevladimir, line 64: So meaning the network indicator has nothing to do with the wallet and everything with the dao?

packages/web-app/src/containers/selectChainForm/index.tsx
```diff
@@ -119,26 +108,30 @@ const SelectChainForm: React.FC = () => {
...
-            key={id}
+            key={selectedNetwork}
             name="blockchain"
             rules={{required: true}}
             control={control}
             defaultValue={{
-              id: (account && chainId) || 1,
-              label: (account && networkName) || 'Ethereum',
-              network: networkType,
+              id: CHAIN_METADATA['ethereum'].id,
```
Fabricevladimir, line 118: Why default to ethereum instead of the wallet?

packages/web-app/src/utils/paths.ts
```diff
@@ -1,12 +1,45 @@
...
+
+/**
+ * Replaces the network parameter in certain URLs. If the path passed to this
+ * function has no parameter, the function will return that path unchanged.
+ *
+ * @param path parametrized path that needs its paramter replaced
+ * @param network network paramater that will replace ":network" in the path
+ * @returns concrete path with network.
+ */
+export function replaceNetworkParam(path: string, network: SupportedNetworks) {
```
Fabricevladimir, line 32: This, is annoying to have to do honestly, I wonder if we can find a way to circumvent it. Once we have dao name, we might be able to do route nesting which wont need us to manually add the network

Fabricevladimir (commented):
Good work, and thanks for the cleanup along the way! Left a couple of comments

packages/web-app/src/utils/constants/api.ts
```diff
@@ -0,0 +1,32 @@
+type SubgraphNetworkUrl = {
+  [key: string]: string | undefined;
+};
+
+export const SUBGRAPH_API_URL: SubgraphNetworkUrl = {
+  main: undefined,
+  // rinkeby:
+  //   'https://api.thegraph.com/subgraphs/name/aragon/aragon-zaragoza-rinkeby',
+  rinkeby:
+    'https://api.thegraph.com/subgraphs/name/rekard0/rekard0-zaragoza-rinkeby',
```
sepehr2github: This subgraph URL isn't valid anymore. we should use aragon graph instead of serkawt

sepehr2github (commented):
Thanks, Vali, Awesome job. It's a really important PR for WebApp. overall It's good but there are some minors things:

there are two things that I'm worried about. 1) integration with web3modal 2) graphs URLs need to be changed to Aragon ones

packages/web-app/src/containers/navbar/index.tsx
WissenIstNacht: Yes.

Just to avoid confusion: I said `usupported` aragon app is not deployed to network x (e.g., Ropsten is unsupported). This is not the same as `wrong network` which would be when the app is on Mainnet, but your wallet is on Rinkeby. The former can no longer happen, the latter can (but is handled differently).

packages/web-app/src/containers/selectChainForm/index.tsx
WissenIstNacht: Great question. I thought about it but then decided to leave things that involve the wallet untouched as much as possible until Jose's PR is finalized. I should have left a comment for someone to tackle at a later point, though. I'll update that.

packages/web-app/src/utils/paths.ts
WissenIstNacht: I think the router lib actually provides something along those lines. As always, I only found out after I had already implemented it. So in order not to waste time I left it in. But if it's a quick fix I'll replace it with the one from the library 👍🏼

packages/web-app/src/utils/constants/api.ts
WissenIstNacht: Oh, nice 🥳

Is that already in the dev branch? If so I will merge it in. Otherwise, I will leave this up for someone who is working on this to change if that's ok :)

packages/web-app/src/containers/navbar/index.tsx
Fabricevladimir: Do a quick check with Selim on that will ya? Because when working on the navbar the network indicator came up and it seemed as if it was geared towards the wallet and not the app.

packages/web-app/src/utils/constants/api.ts
Fabricevladimir: Can confirm already in dev!

WissenIstNacht: Yup, thanks, have merged the changes into my PR.

packages/web-app/src/containers/navbar/index.tsx
WissenIstNacht: For future reference: The wallet is indeed no longer the driver for network information. Instead this task is completely handled by the app (which in turn is informed by the URL params). The consequence of this is that the app can no longer be on an unsupported network. 

The network indicator has therefore been redesigned to show whether the wallet is on the same network as the app. If it is not, a warning banner is shown. If it is, nothing is shown. Also, the banner that distinguishes between main and test nets is gone.

Fabricevladimir (approved):

sepehr2github (approved):
Thanks, For resolving comments 👍

PR #318, opened 2022-04-18
Feature: wallet base Dao - addresses modal - APP-4 

sepehr2github:
## Description

- addresses modal added to go live page for wallet base dao
- edit the voting table in component UI to fix single column view issue
- fix reported issues on setup community/governance page

Task1: [APP-4](https://aragonassociation.atlassian.net/browse/APP-4) 
Task2: [APP-309](https://aragonassociation.atlassian.net/browse/APP-309) 
Task3: [APP-310](https://aragonassociation.atlassian.net/browse/APP-310) 

## Type of change

## Checklist:


packages/web-app/src/containers/configureCommunity/index.tsx
```diff
@@ -186,6 +191,7 @@ const ConfigureCommunity: React.FC = () => {
             control={control}
             defaultValue="1"
             rules={{
+              required: t('errors.emptyDistributionDays') as string,
```
Fabricevladimir: Sonarcloud complains about the as string cast 😢

packages/web-app/src/components/whitelistWallets/index.tsx
```diff
@@ -26,30 +26,27 @@ export const WhitelistWallets = () => {
...
-  useEffect(() => {
-    if (account) {
-      update(0, {address: account});
-    }
-  });
-
   // reset all
   const handleDeleteWallets = () => {
-    replace([{address: account}, {address: ''}]);
+    replace([{address: account}]);
```
josemarinas: In the designs, when this table is displayed for the first time has 1 empty field, If you delete this empty address It won't match de design anymore

packages/web-app/src/containers/communityAddressesModal/index.tsx
```diff
@@ -58,7 +69,7 @@ const CommunityAddressesModal: React.FC = () => {
         {filteredAddressList.length !== 0 ? (
           <VotersTable
```
Fabricevladimir, line 70: Can you add a width to this one if possible? When there is only my wallet on the list the table shrinks

packages/web-app/src/components/whitelistWallets/row.tsx
```diff
@@ -74,8 +74,6 @@ export const Row = ({index}: WhitelistWalletsRowProps) => {
           <Dropdown
```
Fabricevladimir, line 78: I noticed that even with the disabled my wallet row, the dropdown can be triggered and the options are still available. With that we can remove the my wallet option all together. I'm not sure why the button isn't taking the disabled state.

Because of that, we can reset or delete my wallet and still fool the form into thinking it's valid. There is also an issue of the error state staying even when we switch from wallet based to token based dao, but if this dropdown or the options in there can be disabled that issue can't be reproduced.

packages/web-app/src/pages/createDAO.tsx
```diff
@@ -50,7 +50,7 @@ const defaultValues = {
     {address: 'My Wallet', amount: '0'},
   ],
   membership: 'token',
-  whitelistWallets: [{address: 'My Wallet'}, {address: ''}],
+  whitelistWallets: [{address: 'My Wallet'}],
```
josemarinas, line 53: Same, that in the previous comment, this is necessary to match the figma design

Fabricevladimir (changes requested):
Thanks for this one! Left a few comments

Fabricevladimir:
One last thing, could you remove the coming soon text from the Authorized Wallets listItem?

josemarinas (commented):
Thank you, these bugs were pretty annoying, I left you a couple of comments

packages/web-app/src/containers/configureCommunity/index.tsx
josemarinas: Yes, but if you don't do this cast then typescript will complain about not doing the cast :smiling_face_with_tear:

packages/web-app/src/containers/configureCommunity/index.tsx
```diff
@@ -27,16 +28,20 @@ const ConfigureCommunity: React.FC = () => {
     return value <= 100 && value >= 0 ? true : t('errors.percentage');
   };
 
-  const minimumParticipationPercent = useMemo(
-    () =>
+  const minimumParticipationPercent = useMemo(() => {
```
WissenIstNacht, line 31: Wait, what...? 😅

A few things look very suspicious here. Why is the code doing `100*100`? Why is it rounding something and then dividing again? Why is it ceiling something inside a round?

Is there a formula written down for this somewhere?

packages/web-app/src/containers/configureCommunity/index.tsx
WissenIstNacht: Sonarcloud is right. The type cast ist not necessary (search for "required: t(" in this project and you'll find examples without string cast).

I'm pretty sure I've had this problem as well. I can't remember how I solved it, though. :/

WissenIstNacht (commented):
Thanks a lot for the fixes!

I don't have much to add. It would be cool if you could get back to me on the math though :)

packages/web-app/src/containers/configureCommunity/index.tsx
sepehr2github: @josemarinas You are the best guy to explain this

josemarinas: Yeah, I'm sure there is a more elegant way of doing this but basically what you want is the rounded  percentage of votes that you will need for example, if the wallet length is 3 and the minimum participation the expected result of this is `66.66%`

So first you compute the number of wallets that are needed to approve the proposals. Using the previous example, the expected result will be `2`, and the code that gives this result is:

```js
          Math.ceil(
            ((minimumParticipation || defaultMinimumParticipation) *
              whitelistWallets.length) /
              100
          )) /
```

Now with this result you want to obtain the `66.67%` previously mentioned. If we call the result obtained in the previous formula `minWalletsApproval` the percentage represented should be computed like this

```js
minWalletsApproval*100/whitelistWallets.length
```
This will give you 66.66666666666... but you are only interested in 2 decimals, so you multiply do this to obtain an integer with the precision you want:
```js
Math.ceil((minWalletsApproval*100/whitelistWallets)*100)
```
And you will obtain `6667` if you divide this by 100 you will obtain `66.67`

As I said, I'm sure there is a simpler and more elegant way of doing this

Fabricevladimir (approved):
Ready to go!

josemarinas (approved):
Seems nice to me, I'll have to review the code on the validation and default value in the tables before the create DAO branch is merged. Because it will expect the value to be the actual wallet address and not `My Wallet`

PR #329, opened 2022-04-28
Fix: Create DAO transaction pipeline

Fabricevladimir:
## Description

- refactors and cleans create-dao context
- points to temporary subgraph for QA
- fixes wallets not being displayed on whitelist membership creation
- prevents user from spamming transactions and closing publish DAO transaction modal while the transaction is loading

Task: [APP-372](https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/40?modal=detail&selectedIssue=APP-372)

## Type of change

## Checklist:


packages/web-app/package.json
```diff
@@ -17,8 +17,8 @@
   },
   "dependencies": {
     "@apollo/client": "^3.5.8",
-    "@aragon/sdk-client": "0.0.3-alpha",
-    "@aragon/sdk-common": "0.0.1-alpha",
+    "@aragon/sdk-client": "^0.0.5-alpha",
```
marcvelmer: I would block the version number instead of using the `^` for versioning. I know this shouldn't cause any problems as it is but... TBH I would just for now block it to the current version, at least till we can deliver the MVP

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,211 @@
+import {
+  ICreateDaoERC20Voting,
+  ICreateDaoWhitelistVoting,
+} from '@aragon/sdk-client';
+
+import {
+  DaoConfig,
+  VotingConfig,
+} from '@aragon/sdk-client/dist/internal/interfaces/dao';
```
marcvelmer: I don't think it's a good idea to import from `dist`, because anything can be changed here (and will be). I would rather only import those things the SDK is expecting to export, which are those you can import via `@aragon/sdk-client`. So, this should be removed for sure. The question here is... I'm not exporting these both interfaces because I don't think they are necessary. The only important interface here for the UI should be `ICreateDaoERC20Voting` and `ICreateDaoWhitelistVoting` which already have the information about the typing needed for the `DaoConfig` and the `VotingConfig`. Is this really needed? I would suggest no.

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  const {createErc20, createWhitelist} = useDao();
+
+  /*************************************************
+   *                   Handlers                    *
+   *************************************************/
+  const handlePublishDao = () => {
+    switch (membership) {
+      case 'token':
+        setTransaction(getERC20VotingDaoSettings());
```
marcvelmer: `setTransaction`? I see here naming problems. That's not a transaction, right? It's just the data for creating the DAO. Would be better something like `setCreationDaoData`?

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,211 @@
...
+        break;
+      case 'wallet':
+        setTransaction(getWhiteListVotingDaoSettings());
```
marcvelmer: Same as above

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,211 @@
...
+    }, [getDaoConfig, getValues, getVotingConfig]);
+
+  // run dao creation transaction
+  const createDao = useCallback(async () => {
+    setTransactionState(TransactionState.LOADING);
+
+    try {
+      const address =
+        membership === 'token'
+          ? await createErc20(transaction as ICreateDaoERC20Voting)
```
marcvelmer: Same as above

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,211 @@
...
+          : await createWhitelist(transaction as ICreateDaoWhitelistVoting);
```
marcvelmer: Same as above

packages/web-app/package.json
```diff
@@ -17,8 +17,8 @@
...
+    "@aragon/sdk-common": "^0.0.1-alpha",
```
marcvelmer: Can you try if everything works without `@aragon/sdk-common`? Where is being used by the UI? This dependency is needed, but only by the `@aragon/sdk-client` package, which should be included transparently for you as a dependency from a dependency, but should not be explicitly here.

Fabricevladimir: It should work without it then. I wasn't exactly sure why we had it installed, but a quick search proved it to be unused. Nice catch, thanks!

marcvelmer (approved):

packages/web-app/src/pages/createDAO.tsx
```diff
@@ -50,22 +51,27 @@ const defaultValues = {
   tokenTotalSupply: 0,
   links: [{label: '', href: ''}],
   wallets: [
-    {address: 'DAO Treasury', amount: '0'},
-    {address: 'My Wallet', amount: '0'},
+    {address: i18n.t('labels.daoTreasury'), amount: '0'},
```
brickpop: Will this keep working if the user changes the language in the future? Wouldn't it leave an outdated default string?

packages/web-app/src/utils/constants/api.ts
```diff
@@ -5,13 +5,16 @@ type SubgraphNetworkUrl = Record<SupportedNetworks, string | undefined>;
 export const SUBGRAPH_API_URL: SubgraphNetworkUrl = {
   ethereum: undefined,
   rinkeby:
-    'https://api.thegraph.com/subgraphs/name/aragon/aragon-zaragoza-rinkeby',
+    // 'https://api.thegraph.com/subgraphs/name/aragon/aragon-zaragoza-rinkeby',
```
brickpop: I would add a parameterized (process.env) getter, a very clear `TODO` comment and a Task to remove this later on

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,207 @@
...
+   *                   Helpers                     *
+   *************************************************/
+  // get dao configurations
+  const getDaoConfig = useCallback(() => {
+    const {daoName} = getValues();
+    return {
+      name: daoName,
+      metadata: constants.AddressZero,
+    };
+  }, [getValues]);
```
brickpop, line 99: Shouldn't this be dependent on the actual data, and not on a function "pointer"? This function could be redeclared at each single react iteration. The end goal is to cache the data itself I guess?

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,207 @@
...
+    return {
+      minParticipation: parseInt(minimumParticipation) || 0,
+      minSupport: parseInt(minimumApproval) || 0,
+      minDuration: getSecondsFromDHM(
+        parseInt(durationDays),
+        parseInt(durationHours),
+        parseInt(durationMinutes)
+      ),
+    };
+  }, [getValues]);
```
brickpop, line 120: Same as above

packages/web-app/src/context/createDao.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+      // mint configuration
+      mintConfig: values.wallets
+        .filter(wallet => isAddress(wallet.address))
+        .map(wallet => ({
+          address: wallet.address,
+          balance: BigInt(wallet.amount),
+        })),
+    };
+  }, [getDaoConfig, getValues, getVotingConfig]);
```
brickpop, line 148: Same as above

brickpop (commented):
Hi Fabrice, thank you for your PR! I'm leaving comments on the context part, not touching on the rest

packages/web-app/src/pages/createDAO.tsx
Fabricevladimir: This one, I'm not sure about, but, I was made aware of an issue with wallets anyway so it'll be gone once I've pushed the fix

packages/web-app/src/context/createDao.tsx
Fabricevladimir: `getValues` doesn't change between renders (see documentation [here](https://react-hook-form.com/api/useform/getvalues)). Since the dependency (getValues) doesn't change, the function itself won't get redeclared. 

The intention itself isn't actually to cache the data per se. The idea behind is to set it up such that if different packages come with different properties for dao config and voting config (below), we can simply extend this method with a conditional. That said, we can easily switch useCallback for useMemo to have the actual data instead of a function pointer as you've said. Thoughts?

packages/web-app/src/context/createDao.tsx
Fabricevladimir: Another issue perhaps with straight setting the values from `getValues` is that the CreateDao context is wrapped around the entire form so to speak, meaning the values `getValues` would change as the user fills out the form, but we wouldn't be able to track that. We can maybe do it with `watch` (see docs [here](https://react-hook-form.com/api/useform/watch)) that introduces re-renders as the state of the watched fields change.

packages/web-app/src/pages/createDAO.tsx
brickpop: mm ok, but there where is the fix? Can't it be part of this PR? 
The issue here is that there is a first execution using the default language, and then... if different settings are put in place by the app, this default value will never be recomputed

packages/web-app/src/context/createDao.tsx
brickpop: If what we care about is the data, then I'd lean towards useMemo but I lack context to understand how the form context is built in the first place. Having the precomputed value as opposed to a sync getter that could be returned right away

packages/web-app/src/context/createDao.tsx
brickpop: The most logical way to frame this (at least in my limited view) would be by having all the raw parameters as the dependencies array, and then using useMemo to do the parsing and creating the struct that can be passed out

packages/web-app/src/pages/createDAO.tsx
Fabricevladimir: I'm working on it, it'll be pushed up as soon as it's done :)

packages/web-app/src/context/createDao.tsx
Fabricevladimir: No, I think I get what you mean. What I'm saying though is that `getValues` and `watch` are the two ways provided by the `react-hook-form` library to access the form data (short of actually passing the data directly to the context, which has the issue of pushing up unnecessary data). `watch` gives us "observable" parameters at the cost of re-renders and since this context is wrapped around the entire flow, every time one piece of that data changes, it'll trigger a re-render. `getValues` seems the better option because we only, well, get the values 😄 once we actually need them. The callback wrapped helper methods are only to prevent the methods themselves from being recreated on each render unless the dependencies changes, which they never do because `getValues` itself doesn't get redeclared.

sepehr2github (approved):
Thank you, Fabrice. Besides what Marc & jordi said. I only found one issue which isn't a issue because user has to connect his/her wallet before starting the process!

once you took care of Jordi concerns I feel It's good to go

https://drive.google.com/file/d/13bvJbausExj6Dids8yY3navCZ0wssPXR/view?usp=sharing

PR #343, opened 2022-05-06
Feature: Add DAO card to explorer

josemarinas:
## Description

Add DAO card to the explorer page

Task: [APP-295](https://aragonassociation.atlassian.net/browse/APP-295)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/components/daoCard/index.tsx
+++ b/packages/web-app/src/components/daoCard/index.tsx
@@ -1,0 +1,95 @@
+import React from "react"
+import styled from "styled-components"
+import { Avatar, IconBlock, IconPerson } from '@aragon/ui-components'
+
+export interface IDaoCardProps {
+  name: string
+  logo: string
+  description: string
+  chainId: number
+  daoType: DaoType
+}
+
+type DaoType = 'type-1' | 'type-2'
+
+
+
+export const DaoCard = (props: IDaoCardProps) => {
+  return (
+    <Container
+      data-testid="daoCard"
+    >
+      <DaoWrapper>
+        <DaoDataWrapper>
+          <HeaderContainer>
+            <Avatar
+              mode="circle"
+              size="large"
+              src="https://banner2.cleanpng.com/20180325/sxw/kisspng-computer-icons-avatar-avatar-5ab7529a8e4e14.9936310115219636745829.jpg"
+            />
+            <Title>
+              DAO Name
+            </Title>
+          </HeaderContainer>
+          <Description>
+            this is a description
+          </Description>
+        </DaoDataWrapper>
+
+        <DaoDetailsWrapper>
+          <IconWrapper>
+            <IconBlock />
+            <IconLabel>
+              Ethereum
+            </IconLabel>
+          </IconWrapper>
+          <IconWrapper>
+            <IconPerson />
+            <IconLabel>
+              DAOName
+            </IconLabel>
+          </IconWrapper>
+        </DaoDetailsWrapper>
+      </DaoWrapper>
+    </Container>
+  )
+}
+
+const Container = styled.button.attrs({
+  className:
+    `p-2 desktop:p-3 w-full flex flex-col box-border
+    focus:outline-none focus:ring-2 focus:ring-primary-500
+    active:border-ui-200
+    hover:border-primary-500 hover:border-3 hover:border-solid
+    bg-white rounded-xl`
+})``;
+
+const HeaderContainer = styled.div.attrs({
+  className: `flex flex-row space-x-2 items-center`
+})``;
+
+const Title = styled.p.attrs({
+  className: `font-bold text-ui-800 text-base desktop:text-xl`
+})``;
+const Description = styled.p.attrs({
+  className: `font-medium text-ui-600 text-sm desktop:text-base flex`
+})``;
+
+const DaoWrapper = styled.div.attrs({
+  className: `flex flex-col space-y-3`
+})``;
+
+const DaoDetailsWrapper = styled.div.attrs({
+  className: `flex flex-row space-x-3`
+})``;
+const IconLabel = styled.p.attrs({
+  className: ``
+})``;
+const IconWrapper = styled.div.attrs({
+  className: `flex flex-row space-x-1 items-center`
+})``;
+
+const DaoDataWrapper = styled.div.attrs({
+  className: `flex flex-col space-y-1.5`
+})``;
+
--- a/packages/web-app/src/components/daoCard/index.tsx
+++ b/packages/web-app/src/components/daoCard/index.tsx
@@ -101,7 +101,7 @@ const IconWrapper = styled.div.attrs({
 })``;
 
 const DaoDataWrapper = styled.div.attrs({
-  className: 'flex flex-col grow space-y-1.5',
+  className: 'flex flex-col grow space-y-1.5 flex-1',
 })``;
 
 const StyledIconBlock = styled(IconBlock).attrs({
--- a/packages/web-app/src/components/daoCard/index.tsx
+++ b/packages/web-app/src/components/daoCard/index.tsx
@@ -3,8 +3,7 @@ import styled from 'styled-components';
 import {AvatarDao, IconBlock, IconCommunity} from '@aragon/ui-components';
 import {CHAIN_METADATA} from 'utils/constants';
 import useScreen from 'hooks/useScreen';
-import { useTranslation } from 'react-i18next';
-import { useNetwork } from 'context/network';
+import {useTranslation} from 'react-i18next';
 
 export interface IDaoCardProps {
   name: string;
@@ -17,7 +16,7 @@ export interface IDaoCardProps {
 export type DaoType = 'wallet-based' | 'token-based';
 
 const useGetDaoType = (daoType?: DaoType) => {
-  const {t} = useTranslation()
+  const {t} = useTranslation();
   switch (daoType) {
     case 'token-based':
       return t('explore.explorer.tokenBased');
@@ -34,7 +33,7 @@ type DescriptionProps = {
 export const DaoCard = (props: IDaoCardProps) => {
   const [networkName, setNetworkName] = useState('');
   const {isDesktop} = useScreen();
-  const daoType = useGetDaoType(props.daoType)
+  const daoType = useGetDaoType(props.daoType);
 
   useEffect(() => {
     let networks: keyof typeof CHAIN_METADATA;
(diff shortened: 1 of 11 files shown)


packages/web-app/src/components/daoCard/index.tsx
```diff
@@ -0,0 +1,113 @@
+import React, {useEffect, useState} from 'react';
+import styled from 'styled-components';
+import {Avatar, IconBlock, IconCommunity} from '@aragon/ui-components';
+import {CHAIN_METADATA} from 'utils/constants';
+import {t} from 'i18next';
+
+export interface IDaoCardProps {
+  name?: string;
+  logo: string;
```
Fabricevladimir: Can you make logo optional just in case the DAO has no logo?

packages/web-app/src/components/daoCard/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  description?: string;
+  chainId?: number;
+  daoType?: DaoType;
+}
+
+export type DaoType = 'wallet-based' | 'token-based';
+
+const getDaoType = (daoType?: DaoType) => {
```
Fabricevladimir: Were you able to take a look at what the query returns for a DAO "type"?

packages/web-app/src/components/daoCard/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+      return t('explore.explorer.tokenBased');
+    case 'wallet-based':
+      return t('explore.explorer.walletBased');
+  }
+};
+
+export const DaoCard = (props: IDaoCardProps) => {
+  const [networkName, setNetworkName] = useState('');
+
+  useEffect(() => {
```
Fabricevladimir, line 38: Nothing to change here, but would be nice if this logic could be extracted to the parent component and leave this one as a mere presentation component. But I guess this is going to depend on what the query returns. Thoughts? 🤔

packages/web-app/src/components/daoCard/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+        return;
+      }
+    }
+  }, [props.chainId]);
+
+  return (
+    <Container data-testid="daoCard">
+      <DaoDataWrapper>
+        <HeaderContainer>
+          <Avatar mode="circle" size="large" src={props.logo} />
```
Fabricevladimir: There is a AvatarDao, but it's in need of a refactor. I have done so in my upcoming PR (CardProposal). It also has a fallback for when no logo is present. I think I understand why that `logo` is not optional. Once, the refactor has been merged, I propose we switch this over to using the AvatarDao instead.

packages/web-app/src/containers/daoExplorer/daoExplorer.tsx
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  const handleFliterChange = (filterValue: string) => {
+    if (filterValue === 'my-daos') {
+      setDaos([placeholderDaos[0]]);
+      setFilter(filterValue);
+      return;
+    }
+    setFilter(filterValue);
+    setDaos([...placeholderDaos]);
```
Fabricevladimir, line 57: I'm assuming we'll actually filter once we're fetching from the subgraph, right?

packages/web-app/src/components/daoCard/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  className: 'flex flex-row space-x-2 items-center',
+})``;
+
+const Title = styled.p.attrs({
+  className: 'font-bold text-ui-800 text-base desktop:text-xl',
+})``;
+
+const Description = styled.p.attrs({
+  className: `
+  font-medium text-ui-600 text-sm desktop:text-base flex line-clamp-2 text-left
```
Fabricevladimir: On Mobile, line clamp should be 3

packages/web-app/src/containers/daoExplorer/daoExplorer.tsx
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  return (
+    <Container>
+      <MainContainer>
+        <HeaderWrapper>
+          <Title>{t('explore.explorer.title')}</Title>
+          <ButtonGroup
```
Fabricevladimir: Button group should hug contents and not be full width on mobile.

packages/web-app/src/pages/explore.tsx
```diff
@@ -79,6 +84,11 @@ const Container = styled.div.attrs({
   className: 'mx-auto',
 })``;
 
+const DaoWrapper = styled.div.attrs({
```
Fabricevladimir: Can you take a look at the spacing on mobile as well?

Fabricevladimir (changes requested):
Good work! Left a couple of comments and suggestions

packages/web-app/src/components/daoCard/index.tsx
josemarinas: No, this just me guessing stuff, I tried to see the metadata on subgraph, but it returns a hash and I could not find it on IPFS, so these are the two types of DAO and this should be updated once subgraph returns the metadata

packages/web-app/src/components/daoCard/index.tsx
josemarinas: Yes, I assumed it will return the chainID. I believe the computation of the chainName is fine there, yes it introduces a bit of logic into the component but if you move it to the parent component you will have to watch the list of DAOs and iterate over all the list each time the user clicks see more

packages/web-app/src/components/daoCard/index.tsx
josemarinas: Sure, if you want we can merge your PR first, and then I'll rebase my code and add the `AvatarDao` component

packages/web-app/src/containers/daoExplorer/daoExplorer.tsx
josemarinas: Yes, for now is just to have the state and show different things depending on which filter you have selected, for example the `my-daos` filter does not show the `Show more` button

packages/web-app/src/components/daoCard/index.tsx
josemarinas: Noted, thank you :)

packages/web-app/src/containers/daoExplorer/daoExplorer.tsx
josemarinas: Yes, I tried to fix that, but I couldn't find a way for it not growing, I spent a couple of hours and I decided to address it later

packages/web-app/src/pages/explore.tsx
josemarinas: Sure :)

packages/web-app/src/components/daoCard/index.tsx
josemarinas: Yes, I should make the name mandatory and the logo optional, you are right

packages/web-app/src/components/daoCard/index.tsx
Fabricevladimir: Fair enough.

packages/web-app/src/components/daoCard/index.tsx
Fabricevladimir: Good point.

packages/web-app/src/components/daoCard/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  logo: string;
+  description?: string;
+  chainId?: number;
+  daoType?: DaoType;
+}
+
+export type DaoType = 'wallet-based' | 'token-based';
+
+const getDaoType = (daoType?: DaoType) => {
+  switch (daoType) {
```
WissenIstNacht, line 18: What happens on `undefined`?

packages/web-app/src/components/daoCard/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+    case 'token-based':
+      return t('explore.explorer.tokenBased');
```
WissenIstNacht, line 22: Another thing that bugs me here:

Usually, the `t` function is obtained through the `useTranslation` hook. But in this function you use and *imported* `t` function. Does that do the same thing? If you're not 100% certain I would suggest going the safe way and add

````
  const {t} = useTranslation();
````

right at the start of the function.

File: packages/web-app/src/components/daoCard/index.tsx (hunk 9 above)
Comment (WissenIstNacht) [inline packages/web-app/src/components/daoCard/index.tsx:20]: What happens on `undefined`?

**Review by WissenIstNacht — CHANGES_REQUESTED** [review]
Thanks, looks good :)

File: packages/web-app/src/components/daoCard/index.tsx (hunk 10 above)
Comment (josemarinas) [reply packages/web-app/src/components/daoCard/index.tsx:22]: Yes, but this is an external function to the render function and hooks can only be used inside the render function. I can include this function inside the render function `DaoCard` and then I can use the `useTranslation` but since this is something computed externally to the render function, for me, it makes sense to have it outside the render.

File: packages/web-app/src/components/daoCard/index.tsx (hunk 9 above)
Comment (josemarinas) [reply packages/web-app/src/components/daoCard/index.tsx:20]: The only optional value should be the logo because is not enforced in the dao creation, the other ones should be always present, so no default case needed

File: packages/web-app/src/components/daoCard/index.tsx (hunk 10 above)
Comment (Fabricevladimir) [reply packages/web-app/src/components/daoCard/index.tsx:22]: Do you know if this pulls the configs from our instance of i18n that we are exporting from the `i18n.config.ts` in the root of the project? Unless you're absolutely certain, I would feel a bit more confident if we used the `t` function from the export ( `import i18n from '../../i18n.config.ts'`) like so `i18n.t('translationKey')` 😅

File: packages/web-app/src/containers/carousel/index.tsx [hunk 11]
```diff
@@ -7,10 +7,36 @@ import 'react-responsive-carousel/lib/styles/carousel.min.css';
 import CTACard from 'components/ctaCard';
 import {CTACards} from 'components/ctaCard/data';
 import useScreen from 'hooks/useScreen';
+import {useWallet} from 'hooks/useWallet';
 
 const Carousel: React.FC = () => {
   const {isDesktop} = useScreen();
   const navigate = useNavigate();
+  const {methods, isConnected} = useWallet();
```
Fabricevladimir, line 15: Good catch!

Fabricevladimir (approved):

packages/web-app/src/components/daoCard/index.tsx
josemarinas: I'm not sure, so I'll import it like that, @WissenIstNacht are you ok with this?

WissenIstNacht (approved):

PR #347, opened 2022-05-10
Feature/app 401 redirect on non existing dao

WissenIstNacht:
## Description

Reroutes users to "Not Found" page if the dao parameter does not belong to an existing DAO.
Task: [ID]()

## Type of change

## Checklist:


RakeshUP (approved):

packages/web-app/src/app.tsx
```diff
@@ -77,7 +78,7 @@ function App() {
               <Route path="settings" element={<SettingsPage />} />
             </Route>
           </Route>
-          <Route path={paths.NotFound} element={<NotFoundPage />} />
+          <Route path={NotFound} element={<NotFoundPage />} />
```
Fabricevladimir, line 83: Why is this one special enough to have it's own constant?

packages/web-app/src/hooks/useDao.tsx
```diff
@@ -0,0 +1,24 @@
+import {useQuery} from '@apollo/client';
+import {DAO_BY_ADDRESS} from 'queries/dao';
+import {useEffect} from 'react';
+import {useNavigate, useParams} from 'react-router-dom';
+import {NotFound} from 'utils/paths';
+
+export function useDaoParam() {
+  const {dao} = useParams();
+  // NOTE At this point, daoParam will always be defined.
+  const {data, error, loading} = useQuery(DAO_BY_ADDRESS, {
```
Fabricevladimir: Can the two hooks not be combined? Like data is null if no dao exists. Just a thought. 

Also, am I correct to assume that the data will be coming from the SDK in the future and not queried directly? If so, can you add a comment indicating so?

packages/web-app/src/hooks/useDao.tsx
```diff
@@ -0,0 +1,24 @@
...
+    variables: {id: dao ? dao : ''},
```
Fabricevladimir: Previous comment stated that the dao parameter will already have been defined by then, why pass an empty string?

packages/web-app/src/hooks/useDao.tsx
```diff
@@ -0,0 +1,24 @@
+import {useQuery} from '@apollo/client';
+import {DAO_BY_ADDRESS} from 'queries/dao';
+import {useEffect} from 'react';
+import {useNavigate, useParams} from 'react-router-dom';
+import {NotFound} from 'utils/paths';
+
+export function useDaoParam() {
```
Fabricevladimir: Typo in naming? Did you intend to name the hook `useDao` like the file name suggests?

packages/web-app/src/hooks/useDao.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const {dao} = useParams();
+  // NOTE At this point, daoParam will always be defined.
+  const {data, error, loading} = useQuery(DAO_BY_ADDRESS, {
+    variables: {id: dao ? dao : ''},
+  });
+  const navigate = useNavigate();
+
+  useEffect(() => {
```
Fabricevladimir: For this, an interesting solution might be to move the redirecting to NotFound and loading logic into the DaoLayout component. And then the dao id can be gotten safely from every child route of DaoLayout using react-router's useParam or alternatively a wrapper. Thoughts?

Fabricevladimir (approved):
Thanks, good work. Left a few comments.

packages/web-app/src/app.tsx
WissenIstNacht: Nothing. It's the only one that didn't really need changing...

Btw: The constants have become a bit less useful since we're no longer using absolute paths here. I'm still in favor of keeping the file though. Firstly, because some are still used in certain places. Secondly, because they serve as some sort of documentation.

packages/web-app/src/hooks/useDao.tsx
WissenIstNacht: I changed the hook from `useDaoParam` at some point, because `useDao` was too generic. I'll adapt the file name, though. 👍🏼

packages/web-app/src/hooks/useDao.tsx
WissenIstNacht: Well, this kind of is the hook that combines them...? Or do you mean something different. (It takes the the dao param from the URL, then validates it by querying subgraph).

Regarding where data comes from: I thought we'd mainly stick to the web-app. But there may have been updates I'm not aware of 😅

packages/web-app/src/app.tsx
Fabricevladimir: I mean, if they're being used _anywhere_ then the file is worth keeping.

packages/web-app/src/hooks/useDao.tsx
WissenIstNacht: Pleasing the gods of Typescript. (The param should never be undefined at runtime, but at compile time TS can't know that because `dao` is `string | undefined`.

Fabricevladimir: Ah, makes sense.

packages/web-app/src/hooks/useDao.tsx
WissenIstNacht: I thought about doing something similar (In particular moving the validation logic into the `DaoLayout` and then passing the validated param as a prop to the pages). I didn't follow through because of time.

packages/web-app/src/hooks/useDao.tsx
Fabricevladimir: Hmm 🤔 Disregard the first comment. I realize that you might need to check if a dao exists without actually needing anything from it.

There was a proposal to retrieve all data via the SDK, but I believe this is to be implemented later.

packages/web-app/src/hooks/useDao.tsx
Fabricevladimir: Fair enough. I mean don't even need to pass the valid dao param as a prop considering it can be easily gotten using the `useParam` and `useDaoParam` hooks. DaoLayout would just serve as a gatekeeper making sure nothing goes to the children with an invalid dao parameter. Something for later 👍🏾

PR #359, opened 2022-05-18
Fix: Token amount max and min

Fabricevladimir:
## Description

- Shows error when negative value or value greater than max token amount is given when setting up ERC20 Voting DAO

Task: [APP-501](https://aragonassociation.atlassian.net/browse/APP-501)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/components/addWallets/row.tsx
+++ b/packages/web-app/src/components/addWallets/row.tsx
@@ -10,6 +10,7 @@ import {
   ValueInput,
 } from '@aragon/ui-components';
 import styled from 'styled-components';
+import {BigNumber} from 'ethers';
 import React, {useState} from 'react';
 import {useTranslation} from 'react-i18next';
 import {Controller, useFormContext} from 'react-hook-form';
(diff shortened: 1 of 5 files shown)


packages/web-app/src/components/addWallets/row.tsx
```diff
@@ -32,6 +33,12 @@ export type WalletField = {
   amount: string;
 };
 
+// largest decimal that can be represented in 224 bits
+// before adding the 18 decimals
+const MAX_TOKEN_AMOUNT = BigNumber.from(
```
WissenIstNacht: Isn't this constant given by Ethers or something? Also, could you move to the constants?

WissenIstNacht (approved):
Thanks for the fix Fabrice!

packages/web-app/src/components/addWallets/row.tsx
Fabricevladimir: Unfortunately it isn't no

sepehr2github (approved):
Thanks, Fabrice! It seems Good to me.

PR #361, opened 2022-05-19
Fix: Create DAO network selection

Fabricevladimir:
## Description

- Sets default network when accessing process to be wallet network
- Changes copy on transaction modal successful state

Task: [APP-489](https://aragonassociation.atlassian.net/browse/APP-489), [APP-491](https://aragonassociation.atlassian.net/browse/APP-491)

## Type of change

## Checklist:


packages/web-app/src/utils/constants/chains.ts
```diff
@@ -25,6 +25,21 @@ export function isSupportedNetwork(
...
 
+/**
+ * Get the network name with given chain id
+ * @param chainId Chain id
+ * @returns the name of the supported network or undefined if network is unsupported
+ */
+export function getSupportedNetworkByChainId(
+  chainId: number
+): SupportedNetworks | undefined {
+  if (isSupportedChainId(chainId)) {
```
WissenIstNacht, line 36: I think I would accept only `SupportedChainId` as args and leave this  check up to the caller. True, this leads code dulicaiton. On the other hand, they have to do some check anyway (e.g. to deal with undefined as result).

Not a big deal, though. And thanks for the useful function. :)

WissenIstNacht (approved):
Thanks, everything seems to be working fine :)

packages/web-app/src/utils/constants/chains.ts
Fabricevladimir: I previously had the function accepting only `SupportedChainId` but I realized that not only is code duplicated for the check, but also, any non literal value passed in would need to be typecasted to `SupportedChainId` since the value of `chainId` coming from the wallet is unknown at compile time and is of type `number` in the first place.

WissenIstNacht: Just FYI: The check on line 36 automatically casts (or rather, narrows) the type of a variable with type `number` to `SupportedChainId`. I.e., the variable `chainId` will be considered to be of type `SupportedChainId` within the entire if clause.

RakeshUP (approved):

packages/web-app/src/utils/constants/chains.ts
Fabricevladimir: What I meant was that you have to do the following everytime if `chainId` is of type `SupportedChainId`: 
```
const { chainId } = useWallet();

getSupportedNetworkByChainId(chainId as SupportedChainId);
```

PR #363, opened 2022-05-20
Feature: ListItemHeader UI component

Fabricevladimir:
## Description

- adds ListItemHeader component with corresponding tests and story

Task: [APP-509](https://aragonassociation.atlassian.net/browse/APP-509)

## Type of change

## Checklist:

### DIFF
--- a/packages/ui-components/src/components/listItem/header.tsx
+++ b/packages/ui-components/src/components/listItem/header.tsx
@@ -74,7 +74,7 @@ const ContentWrapper = styled.div.attrs(({horizontal}: VariableAlignment) => ({
 }))<VariableAlignment>``;
 
 const Value = styled.p.attrs({
-  className: 'text-2xl text-ui-800 text-bold truncate',
+  className: 'text-2xl text-ui-800 font-bold truncate',
 })``;
 
 const Label = styled.p.attrs({className: 'text-ui-500 truncate'})``;
--- a/packages/ui-components/src/components/listItem/header.tsx
+++ b/packages/ui-components/src/components/listItem/header.tsx
@@ -1,4 +1,4 @@
-import React, {useMemo} from 'react';
+import React from 'react';
 import styled from 'styled-components';
 import {ButtonText} from '../button';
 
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/listItem/header.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <ContentWrapper orientation={orientation}>
+          <Value>{props.value}</Value>
+          <Label>{props.label}</Label>
+        </ContentWrapper>
+      </VerticalContainer>
+    );
+  }
+
+  return (
+    <HorizontalContainer data-testid="listItem-header">
```
WissenIstNacht: That's a bit awkward... Specifying horizontal on mobile just displays nothing. And it's probably a bit of a pain to fix this without having access to an `useScreen` hook. But at least a comment pointing this out would be helpful.

packages/ui-components/src/components/listItem/header.tsx
```diff
@@ -0,0 +1,86 @@
...
+    className: `${
+      orientation === 'vertical'
+        ? 'space-y-0.25'
+        : 'flex flex-1 items-baseline space-x-2'
+    }`,
+  })
+)<ContentWrapperProps>``;
+
+const Value = styled.p.attrs({
+  className: 'text-2xl text-ui-800 text-bold truncate',
```
WissenIstNacht: I think it's `font-bold` :yum:

WissenIstNacht (approved):
Looking good, thanks :)

packages/ui-components/src/components/listItem/header.tsx
Fabricevladimir: Yeah, not ideal. Currently the component is driven by the horizontal/vertical but it would be better to have it be more mobile vs desktop. Actually, there is no reason not to include `useScreen` in the ui-components, as long as the screen sizes are coming from the config.

packages/ui-components/src/components/listItem/header.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /** Value to display */
+  value: string;
+  onClick: () => void;
+};
+
+export const ListItemHeader: React.FC<ListItemHeaderProps> = ({
+  orientation = 'vertical',
+  ...props
+}) => {
+  const horizontal = useMemo(() => orientation === 'horizontal', [orientation]);
```
WissenIstNacht: No need to change, just FYI: For a cheap computation like this the memo is likely more costly than not having it.

Ready to merge from my side 👍🏼 

Also, well done solving the issue with the header not showing depending on the orientation 💪🏼

PR #367, opened 2022-05-23
Feature: Upgrade node version

josemarinas:
## Description

Upgrade node version from 14 to 16.

This is needed because the `js-ipfs` library used in the SDK requires node 16.

Task: [ID]()

## Type of change

## Checklist:


WissenIstNacht:
Does the `yarn.lock` really not change?! 😮

josemarinas:
> Does the `yarn.lock` really not change?! open_mouth

Nope, no dependencies changed, just added a requirement in the node engine version and updated the version used to build and deploy the app, so this does not affect the dependencies

packages/web-app/package.json
```diff
@@ -86,5 +86,8 @@
     "vite": "^2.4.2",
     "vite-tsconfig-paths": "^3.3.17",
     "whatwg-fetch": "^3.6.2"
+  },
```
mathewmeconry, line 89: not sure if this is necessary

mathewmeconry (approved):
LGTM. Small comment where I am not sure

packages/web-app/package.json
josemarinas: Not strictly necessary, but if you don't put it, and you try to do a `yarn dev` for example, it won't complain. So I left it there for in the case you are testing the app with node 14 then `npm` or `yarn` will complain because it is not the correct one, even when you have it specified in the other `package.json`. I'm not sure if importing a package that requires node 16 will make this unnecessary, but for now I'd leave it there

brickpop (approved):

PR #375, opened 2022-06-06
Feature: Mint token proposal action

RakeshUP:
## Description

WalletRow component needs to be changed/duplicated to add 'actions' prefix to all form inputs and also change the way totalSupply is calculated(needs an initialSupply for this particular case). This can be done while integrating with sdk 

Task: [75](https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/40?modal=detail&selectedIssue=APP-75)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/containers/actionBuilder/mintTokens/addAddress.tsx
+++ b/packages/web-app/src/containers/actionBuilder/mintTokens/addAddress.tsx
@@ -1,0 +1,97 @@
+import React, {useEffect} from 'react';
+import styled from 'styled-components';
+import {ButtonText} from '@aragon/ui-components';
+import {useTranslation} from 'react-i18next';
+import {useFormContext, useFieldArray, useWatch} from 'react-hook-form';
+
+import Row from 'components/addWallets/row';
+import Header from 'components/addWallets/header';
+import Footer from 'components/addWallets/footer';
+
+type AddAddressProps = {
+  index?: number;
+  setActionsCounter?: (index: number) => void;
+};
+
+const AddAddress: React.FC<AddAddressProps> = ({
+  index = 0,
+  setActionsCounter,
+}) => {
+  const {t} = useTranslation();
+
+  const {control, setValue, trigger} = useFormContext();
+  const wallets = useWatch({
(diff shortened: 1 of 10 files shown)


packages/web-app/src/containers/actionBuilder/mintTokens/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                onClick={() => {
+                  duplicateAction(index);
+                  setOpenMenu(false);
+                }}
+                bgWhite
+              />
+              <ListItemAction
+                title={t('labels.resetAction')}
+                onClick={() => {
+                  resetWithdrawFields();
```
WissenIstNacht, line 62: This doesn't work. Forgot to adapt a copy/paste?

packages/web-app/src/containers/actionBuilder/mintTokens/addAddress.tsx
```diff
@@ -0,0 +1,97 @@
...
+          label={t('labels.whitelistWallets.addAddress') as string}
+          mode="ghost"
+          size="large"
+          onClick={handleAddWallet}
+        />
+        <ButtonText
+          label={t('labels.whitelistWallets.uploadCSV') as string}
+          mode="secondary"
+          size="large"
+          onClick={handleAddWallet}
```
WissenIstNacht, line 80: The ticket's acceptance criteria still mentions CSV as a feature for this. I'm fine with not having CSV in the MVP, but in that case I think we should remove the button. Or at the very least make it disabled.

Could you ask for clarification on this in Jira?

WissenIstNacht (changes requested):
Thanks for the PR Rakesh.

Aside from the comments regarding your code, there's two more things I noticed:
1. The percentages for the tokens minted to an address are computed using the floor function. I think the simple round method (altough not perfect either), would be nicer.
2. If you add and fill out a Mint Token action, and then add a second mint token action, the first one will be duplicated into the second. Is this expected behaviour?

Don't worry about these points for this task/PR. Just wanted to point it our for the future.

PR #379, opened 2022-06-08
fix mainnet dao id validation

WissenIstNacht:
## Description

Makes app react to changes in the URL's network parameter (i.e., it revalidates the dao parameter with respect to the new network if the network is changed in the URL). To test that this works:
- select an existing DAO
- visit any of its pages
- change the network in the URL
- This should redirect you to not-found, as the dao with the same id does most likely not exist on another testnet and certainly doesn't exist on mainnet.

Task: [494](https://aragonassociation.atlassian.net/browse/APP-494)

## Type of change

### DIFF
--- a/packages/web-app/src/hooks/useDaoParam.tsx
+++ b/packages/web-app/src/hooks/useDaoParam.tsx
@@ -27,6 +27,13 @@ export function useDaoParam() {
   const navigate = useNavigate();
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


Fabricevladimir (approved):
Thanks

RakeshUP (approved):

sepehr2github (approved):

PR #383, opened 2022-06-09
Fix: Gas Estimation Native token

Fabricevladimir:
## Description

- Fixes the native token shown when creating DAO
- Removes test DAO so that data is being loaded from currently accessed DAO

Task: [APP-344](https://aragonassociation.atlassian.net/browse/APP-344)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/services/prices.ts
+++ b/packages/web-app/src/services/prices.ts
@@ -100,7 +100,7 @@ async function fetchTokenData(
 
   // build url based on whether token is native token
   const url = isNativeToken
-    ? `/coins/${getNativeTokenId(platformId, network)}`
+    ? `/coins/${getNativeTokenId(network)}`
     : `/coins/${platformId}/contract/${address}`;
 
   const {data, error} = await client.query({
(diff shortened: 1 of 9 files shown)


RakeshUP (approved):

packages/web-app/src/services/prices.ts
```diff
@@ -154,4 +158,23 @@ async function fetchTokenPrice(
...
+ * to be priced.
+ * @param platformId platform id
+ * @param network network name
+ * @returns native token id
+ */
+function getNativeTokenId(
+  platformId: string | null,
+  network: SupportedNetworks
+): string {
+  if (platformId === ASSET_PLATFORMS.polygon || network === 'mumbai') {
```
WissenIstNacht: When would the first condition not be met, but the second would? Why would this be dependent on the `platformID` and not simply on the network?

WissenIstNacht (approved):
Hey man, looks good!

The code around the `getNativeTokenId` looks a bit weird to me, but it's really a minor thing. Don't sweat it.

packages/web-app/src/services/prices.ts
Fabricevladimir: The first would not be met when the network is Mumbai. None of the testnets have a coingecko platformId because they don't support pricing and information for testnet tokens. That said, it would have been a lot clearer I used network to check for both instead of platformId. Good catch, thanks.

PR #384, opened 2022-06-09
Feature: APP-547 Fix token list on token menu

sepehr2github:
## Description

- Token fetching issue fixed on the token menu
-Custom contract checked on deposit flow and It's working fine

Task: [APP-547](https://aragonassociation.atlassian.net/browse/APP-547)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/context/providers.tsx
+++ b/packages/web-app/src/context/providers.tsx
@@ -32,7 +32,6 @@ type ProviderProviderProps = {
  */
 export function ProvidersProvider({children}: ProviderProviderProps) {
   const {chainId, provider} = useWallet();
-  const [web3Provider, setWeb3Provider] = useState(provider || null);
 
   const [infuraProvider, setInfuraProvider] = useState(
     new InfuraProvider(NW_ARB, INFURA_PROJECT_ID_ARB)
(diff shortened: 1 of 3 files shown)


packages/web-app/src/context/providers.tsx
```diff
@@ -1,6 +1,6 @@
 import {InfuraProvider, Web3Provider} from '@ethersproject/providers';
 import React, {createContext, useContext, useEffect, useState} from 'react';
-import {useWallet} from 'use-wallet';
+import {useWallet} from 'hooks/useWallet';
```
Fabricevladimir, line 3: Aha!

Fabricevladimir (approved):
Thanks. One comment though is that we probably shouldn't return tokens with balance 0 since the user cannot deposit them in the first place.

packages/web-app/src/context/providers.tsx
```diff
@@ -31,10 +31,8 @@ type ProviderProviderProps = {
  * therefore be null if no wallet is connected.
  */
 export function ProvidersProvider({children}: ProviderProviderProps) {
-  const {chainId, ethereum} = useWallet();
-  const [web3Provider, setWeb3Provider] = useState(
-    ethereum ? new Web3Provider(ethereum) : null
-  );
+  const {chainId, provider} = useWallet();
+  const [web3Provider, setWeb3Provider] = useState(provider || null);
```
RakeshUP: do we need to store this in a separate state variable? can't we just use the provider returned by useWallet and send it to the child components?

RakeshUP (approved):

sepehr2github:
> Thanks. One comment though is that we probably shouldn't return tokens with balance 0 since the user cannot deposit them in the first place.

That's true, but regardless when the user sees the insufficient balance, he/she will understand, anyway, good idea we can add later

PR #399, opened 2022-06-17
Addresses issues from APP-591

githubdoramon:
## Description

This PR addresses the following
- Fetches correct balances from user wallet (we were fetching each token contract balanceOf instead of user wallet)
- Sorts tokens on user wallet when we show them on deposit modal, always putting native token at first place
- Display the correct Native token, instead of hardcoding to Eth
- only adds the native token once to the list regardless of how many re-renders
- Handles errors if for any reason user has a weird token on her wallet that would break the app on trying to fetch the balance

Task: [APP-591](https://aragonassociation.atlassian.net/browse/APP-591)

## Type of change

## Checklist:

- [X] I have selected the correct base branch.
- [X] I have performed a self-review of my own code.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have made corresponding changes to the documentation.
- [X] My changes generate no new warnings.
- [X] Any dependent changes have been merged and published in downstream modules.
- [X] I ran all tests with success and extended them if necessary.
- [X] I have updated the ``CHANGELOG.md`` file in the root folder.
- [X] I have tested my code on the test network.

### DIFF
--- a/packages/web-app/src/hooks/useWalletTokens.tsx
+++ b/packages/web-app/src/hooks/useWalletTokens.tsx
@@ -6,8 +6,10 @@ import {useState, useEffect} from 'react';
 import {erc20TokenABI} from 'abis/erc20TokenABI';
 import {useWallet} from 'hooks/useWallet';
 import {useProviders} from 'context/providers';
-import {isETH, fetchBalance, getTokenInfo} from 'utils/tokens';
+import {isNativeToken, fetchBalance, getTokenInfo} from 'utils/tokens';
 import {TokenBalance, HookData} from 'utils/types';
+import {useNetwork} from 'context/network';
+import {CHAIN_METADATA} from 'utils/constants';
 
 // TODO The two hooks in this file are very similar and should probably be
 // merged into one. The reason I'm not doing it now is that I'm not sure if
@@ -80,6 +82,9 @@ export function useUserTokenAddresses(): HookData<string[]> {
 export function useWalletTokens(): HookData<TokenBalance[]> {
   const {address, balance} = useWallet();
   const {infura: provider} = useProviders();
+  const {network} = useNetwork();
+  const nativeCurrency = CHAIN_METADATA[network].nativeCurrency;
+
   const {
     data: tokenList,
     isLoading: tokenListLoading,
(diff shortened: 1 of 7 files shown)


RakeshUP (approved):

sepehr2github (approved):

packages/web-app/src/hooks/useWalletTokens.tsx
```diff
@@ -99,47 +104,67 @@ export function useWalletTokens(): HookData<TokenBalance[]> {
         return;
       }
 
-      if (Number(balance) !== -1 && Number(balance) !== 0)
-        tokenList.push(constants.AddressZero);
+      if (
+        Number(balance) !== -1 &&
```
WissenIstNacht: Might not ever be an issue in practice, but `balance?.eq(-1)` is safer.

packages/web-app/src/utils/tokens.ts
```diff
@@ -147,27 +149,35 @@ export const fetchBalance = async (
...
+    if (shouldFormat) {
+      const {decimals} = await getTokenInfo(
+        tokenAddress,
+        provider,
+        nativeCurrency
+      );
+      return formatUnits(balance, decimals);
+    }
+    return balance;
+  } catch (err) {
```
WissenIstNacht: Thanks for adding a try/catch block. But now instead of handling the error we're just swallowing it. I.e., if something goes wrong here we just return `0` and nobody will know why.

Also,  promises come with a "failure state". Unless we explicitly transform the error into a custom error that we can work with further up the call stack, it might not even be worth to do this here.

packages/web-app/src/utils/constants/chains.ts
```diff
@@ -52,11 +58,7 @@ export type ChainData = {
   explorer: string;
   logo: string;
   rpc: string[];
-  nativeCurrency: {
-    name: string;
-    symbol: string;
-    decimals: number;
-  };
+  nativeCurrency: NativeTokenData;
```
WissenIstNacht, line 61: Look-up type could have been used

WissenIstNacht (commented):
Thanks for the fixes. Left a couple of minor comments.

WissenIstNacht:
Closed in favor of #401

PR #403, opened 2022-06-20
Feature: Withdraw action in proposal creation

RakeshUP:
## Description

The only real change is this [diff](https://github.com/aragon/apps/pull/403/files#diff-6242221d708df91262b8eddf22ae11de873944247b91a75db99d1813cf1e2247R120-R137). The rest of the changes are due to me restructuring the code to be able to share it between the new proposal and new withdraw page

And the caveats remain:
- With Whitelist Voting DAOs, create proposal may fail(if you are not part of the whitelist - gas estimation will show error. We need to prevent users from starting the process itself maybe)
- Metadata isn't handled(needs the IPFS client)
- VotingTerminal data and Setup Voting data are hardcoded. We need to read that from subgraph

Task: [APP-586](https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/40?modal=detail&selectedIssue=APP-586)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/context/createProposal.tsx
+++ b/packages/web-app/src/context/createProposal.tsx
@@ -1,0 +1,194 @@
+import {ICreateProposal} from '@aragon/sdk-client';
+
+import React, {useState, useCallback, useMemo} from 'react';
+import {constants} from 'ethers';
+import {parseUnits} from 'ethers/lib/utils';
+import {generatePath, useNavigate} from 'react-router-dom';
+import {useFormContext} from 'react-hook-form';
+import {useTranslation} from 'react-i18next';
+import {useQuery} from '@apollo/client';
+
+import PublishModal from 'containers/transactionModals/publishModal';
+import {TransactionState} from 'utils/constants';
+import {Governance} from 'utils/paths';
+import {useClient} from 'hooks/useClient';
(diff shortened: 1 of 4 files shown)


packages/web-app/src/context/createProposal.tsx
```diff
@@ -0,0 +1,194 @@
...
+    return whitelistClient.dao.whitelist.createProposal(
+      votingAddress,
+      proposalCreationParams
+    );
+  };
+
+  const encodeActions = () => {
+    const actions = getValues().actions;
+    return actions.map((action: Record<string, string>) => {
+      if (action.name === 'withdraw_assets') {
```
WissenIstNacht, line 123: I assume this is eventually going to become something like a switch statement that encodes each type of action, right?

WissenIstNacht (approved):
Thanks man, LGTM

Fabricevladimir (approved):
Sweet! Good work

packages/web-app/src/context/createProposal.tsx
RakeshUP: Yes. This will be the common method to encode actions and we need to use a switch case and maybe define the action names using an enum

PR #413, opened 2022-06-29
Fix/app 643 font weights

WissenIstNacht:
## Description

Fixes issues with the default font-weight. Font weight is now default over any HTML element. Also fixes a couple of instances where font weight was medium instead of bold.

Task: [643](https://aragonassociation.atlassian.net/browse/APP-643)

## Type of change


RakeshUP (approved):

sepehr2github (approved):
Thanks for the changes!

PR #419, opened 2022-07-06
Feat: Keep wallet connected on page load

RakeshUP:
## Description

The wallet selector popup appears only after you click logout from the app and try to login again. Otherwise, it will always try to automatically connect to the last connected wallet(asks for password if you are locked out of the wallet, if the wallet is unlocked, then it's connected silently)

Task: [APP-487](https://aragonassociation.atlassian.net/browse/APP-487)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/app.tsx
+++ b/packages/web-app/src/app.tsx
@@ -49,9 +49,12 @@ function App() {
   const {methods} = useWallet();
 
   useEffect(() => {
+    // This check would prevent the wallet selection modal from opening up if the user hasn't logged in previously.
+    // But if the injected wallet like Metamask is locked and the user has logged in before using that wallet, there will be a prompt for password.
     if (localStorage.getItem('WEB3_CONNECT_CACHED_PROVIDER')) {
       methods.selectWallet();
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   useEffect(() => {
--- a/packages/web-app/src/app.tsx
+++ b/packages/web-app/src/app.tsx
@@ -21,6 +21,7 @@ import ExploreNav from 'containers/navbar/exploreNav';
 import Footer from 'containers/exploreFooter';
 import NetworkErrorMenu from 'containers/networkErrorMenu';
 import TransferMenu from 'containers/transferMenu';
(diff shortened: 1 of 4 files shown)


packages/web-app/src/app.tsx
```diff
@@ -45,6 +46,16 @@ function App() {
   // TODO this needs to be inside a Routes component. Will be moved there with
   // further refactoring of layout (see further below).
   const {pathname} = useLocation();
+  const {methods} = useWallet();
+
+  useEffect(() => {
+    // This check would prevent the wallet selection modal from opening up if the user hasn't logged in previously.
+    // But if the injected wallet like Metamask is locked and the user has logged in before using that wallet, there will be a prompt for password.
+    if (localStorage.getItem('WEB3_CONNECT_CACHED_PROVIDER')) {
```
sepehr2github, line 54: It seems good! Should we use indexed db rather than local storage, if we can build some policies regarding client caching that would be awesome! @carlosgj94 How do you think?

sepehr2github (approved):

packages/web-app/src/app.tsx
RakeshUP: the web3modal package is the one writing to localStorage. We are just reading it from there to see if there was a logged-in wallet

WissenIstNacht:
Thanks Rakesh, the auto log-in seems to work fine.

Keep in mind that the code now no longer allows you to switch between wallets. If you log-in to metamask once, you wont be able to log in to wallet connect after that (unless you remove the data stored by the browser I guess). This is probably not inteded behaviour. Is there a simple way to reset the memory if a user manually logs out?

RakeshUP:
@WissenIstNacht could you take a look now, I made a change to remove the cachedProvider in web-app. I made changes to the package vocdoni/use-signer to achieve this and seemed to be working when testing, will be checking that again. Meanwhile made a [hacky fix](https://github.com/aragon/apps/pull/419/files#diff-59cdda007e0408f39b9939908bb481a7bf5e1e1e2a1af65b9459676314856808R34) on web-app, will be removed if I am able to solve this in `vocdoni/use-signer`

WissenIstNacht (approved):
Seems to work just fine for now, thanks!

PR #420, opened 2022-07-06
Fix: APP-511 - list item address

WissenIstNacht:
## Description

Adds a label field to the component and adapts the webapp.

Note: Mapping ENS domains to their avatar source will be done in a later step.

Task: [511](https://aragonassociation.atlassian.net/browse/APP-511?atlOrigin=eyJpIjoiNDlmMjk0MzdmMWVlNDc4MzhhYWVmNjdkYzk1YWVmYzAiLCJwIjoiaiJ9)

### DIFF
--- a/packages/ui-components/src/components/listItem/address.tsx
+++ b/packages/ui-components/src/components/listItem/address.tsx
@@ -1,8 +1,9 @@
 import React, {ButtonHTMLAttributes, FC} from 'react';
 import styled from 'styled-components';
+
 import {shortenAddress} from '../../utils/addresses';
 import {AvatarWallet} from '../avatar';
-import {IconLinkExternal} from '../icons';
+import {IconLinkExternal, IconPerson} from '../icons';
 
 type TokenInfo = {
   amount: number;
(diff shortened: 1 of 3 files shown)


RakeshUP (approved):

packages/ui-components/src/components/listItem/address.tsx
```diff
@@ -45,6 +48,13 @@ export const ListItemAddress: FC<ListItemAddressProps> = ({
   );
 };
 
+type AvatarProps = Pick<ListItemAddressProps, 'src'>;
+
+const Avatar: FC<AvatarProps> = ({src}) => {
+  if (!src) return <IconPerson className="w-2.5 h-2.5" />;
+  return <AvatarWallet src={src || ''} />;
```
carlosgj94: Is this `logical or` needed here if we are already checking it in the line before?

sepehr2github (approved):
Apart from left comments, seems good to go! Thanks Vali

packages/ui-components/src/components/listItem/address.tsx
WissenIstNacht: It should work without the empty string fallback, but tsc was not having it for some reason...

Works now though. PR is updated accordingly 👍🏼

PR #436, opened 2022-07-20
Fix/app 699 mobile dialog

WissenIstNacht:
## Description

Fix minor layout changes on the mobile dialogue:

- Backdrop sometimes not filling vertical viewport
- Bottom sheet being too wide because of a text overflow in `DaoHeader` 

Note: As commented in the ticket, two tasks/bugs still need to be tackled. One is preventing a buggy animation when scrolling through a page with an open bottom sheet (not an issue with desktop modals, only mobile bottom sheets). The other is preventing the background from scrolling when a modal/bottom sheet is open. The latter would probably solve the former. I will ask for a new bug ticket to be created focusing just on those issues.

Task: [699](https://aragonassociation.atlassian.net/browse/APP-699?atlOrigin=eyJpIjoiZDMxM2Q0NDczZGEyNGI5Mjg4NjA2YzY4OWMzZDEzMjUiLCJwIjoiaiJ9)

## Type of change

### DIFF
--- a/packages/ui-components/src/components/backdrop/backdrop.tsx
+++ b/packages/ui-components/src/components/backdrop/backdrop.tsx
@@ -57,6 +57,7 @@ const StyledBackdrop = styled.div.attrs(({visible}: StyledBackdropProps) => {
     transition: 'visibility 0.2s, opacity 0.2s linear',
     backdropFilter: 'blur(24px)',
     cursor: 'pointer',
+    marginTop: 0,
   };
 
   return {className, style};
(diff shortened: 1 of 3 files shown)


RakeshUP (approved):
LGTM. 

The underlying background scroll could be prevented by adding something like `document.body.style.overflow = "hidden"` on the openCallback and removing it on close callback I guess..

sepehr2github (approved):
Thanks, Vali! Good to go 👍

PR #454, opened 2022-08-03
Feature: APP-799 Add withdraw executer widget

sepehr2github:
## Description

- Executor widget added to review proposal for withdraw action

Task: [APP-799](https://aragonassociation.atlassian.net/browse/APP-799)

## Type of change

## Checklist:


RakeshUP (approved):
The new execution widget looks very good! But the app crashes when using the New Withdraw flow on the Finance page. Could you look into that?

sepehr2github:
> The new execution widget looks very good! But the app crashes when using the New Withdraw flow on the Finance page. Could you look into that?

Thanks, for mentioning it, Issue is addressed!

packages/web-app/src/components/reviewExecution/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+export const ReviewExecution: React.FC = () => {
+  const {t} = useTranslation();
+  const {getValues} = useFormContext();
+  const {actions} = getValues();
+  const {actions: actionTypes} = useActionsContext();
+
+  const {data: daoId} = useDaoParam();
+  const {data: dao} = useDaoMetadata(daoId);
+
+  const RenderActions = useCallback(
```
WissenIstNacht: A function that returns components? That's just another component! :)

I think I can understand why you did this (easier not having to pass down variables as props maybe). But I would really define `RenderActions` as a normal component outside of `ReviewExecution`. This also takes care of detecting changes (currently, you're missing `action` in the dependency array of use callback). Also, if you're worried about performance of rerenders, don't be. Practically none of the things on this page will ever trigger a re-render.

packages/web-app/src/components/reviewExecution/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+type RenderActionsProps = {
+  action: Action;
+  type: ActionsTypes;
+};
+
+export const ReviewExecution: React.FC = () => {
+  const {t} = useTranslation();
+  const {getValues} = useFormContext();
+  const {actions} = getValues();
```
WissenIstNacht, line 14: Remember when we discussed that this could become like a single execution widget that could be used both for review pages as well as proposal details pages? From a technical perspective this could be tricky. Because from the proposal details page we won't have access to the actions context, right?

No need worry about this right now. Just something we might to ask ourselves at some point.

WissenIstNacht (changes requested):
Thanks a lot, the widget indeed looks sweet! I left a quick refactor request, then it should be good to go!

Fabricevladimir (changes requested):
Good work! I do echo @WissenIstNacht's sentiments in terms of making the execution widget take in the actions and whatever else it needs as props. That said, we can leave this as is and refactor it when we implement the widget on the proposal details page. 

One other thing I encountered was when creating a withdraw through the main create proposal flow, selecting a token from the list doesn't close the modal for whatever reason.

packages/web-app/src/components/reviewExecution/index.tsx
sepehr2github: Agreed, as Fabrice said my plan was to tackle this one once we want to implement the widgets on proposal details page

WissenIstNacht (approved):
Yup, checks out. Thanks a lot!

packages/web-app/src/components/reviewExecution/actionsFilter.tsx
```diff
@@ -0,0 +1,33 @@
...
+  action,
+  type,
+}: ActionsFilterProps) => {
+  const {data: daoId} = useDaoParam();
+  const {data: dao} = useDaoMetadata(daoId);
+
+  switch (type) {
+    case 'withdraw_assets':
+      return (
+        <>
```
Fabricevladimir: Any particular reason for the fragment tags?

Fabricevladimir (approved):
Good to go once you've resolved the conflicts!

PR #457, opened 2022-08-08
Feat: Review DAO checkbox validation

RakeshUP:
## Description

- Refactor the Review DAO page to use DescriptionList component
- Checkboxes in the review DAO page now have an error state
- Added AlertCard to review DAO page

Task: [598](https://aragonassociation.atlassian.net/browse/APP-598), [390](https://aragonassociation.atlassian.net/browse/APP-390)

## Type of change

## Checklist:

### DIFF
--- a/packages/ui-components/src/components/checkbox/checkboxListItem.tsx
+++ b/packages/ui-components/src/components/checkbox/checkboxListItem.tsx
@@ -60,7 +60,7 @@ type ContainerTypes = {
 };
 
 const Container = styled.div.attrs(({disabled, state}: ContainerTypes) => ({
-  className: `w-full py-1.5 px-2 rounded-xl border-2 focus:outline-none focus-visible:ring-2 focus-visible:ring-primary-500 ${
+  className: `py-1.5 px-2 rounded-xl border-2 focus:outline-none focus-visible:ring-2 focus-visible:ring-primary-500 ${
     disabled
       ? 'bg-ui-100 border-ui-300'
       : `bg-ui-0 group hover:border-primary-500 cursor-pointer ${
@@ -71,7 +71,7 @@ const Container = styled.div.attrs(({disabled, state}: ContainerTypes) => ({
 }))<ContainerTypes>``;
 
 const HStack = styled.div.attrs(({disabled, state}: ContainerTypes) => ({
-  className: `flex justify-between items-center group-hover:text-primary-500 ${
+  className: `flex justify-between items-center group-hover:text-primary-500 space-x-1.5 ${
     disabled
       ? 'text-ui-600'
       : state !== 'default'
--- a/packages/ui-components/src/components/checkbox/checkboxListItem.tsx
+++ b/packages/ui-components/src/components/checkbox/checkboxListItem.tsx
@@ -27,6 +27,7 @@ export type CheckboxListItemProps = {
   multiSelect?: boolean;
   disabled?: boolean;
   state?: 'default' | 'active' | 'multi';
+  errorBorder?: boolean;
   onClick?: React.MouseEventHandler;
 };
 
@@ -36,6 +37,7 @@ export const CheckboxListItem: React.FC<CheckboxListItemProps> = ({
   multiSelect = false,
   disabled = false,
   state = 'default',
+  errorBorder = false,
   onClick,
 }) => {
   return (
@@ -43,6 +45,7 @@ export const CheckboxListItem: React.FC<CheckboxListItemProps> = ({
       data-testid="checkboxListItem"
       state={state}
       disabled={disabled}
+      errorBorder={errorBorder}
       {...(disabled ? {} : {onClick})}
     >
       <HStack disabled={disabled} state={state}>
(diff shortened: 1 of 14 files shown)


packages/ui-components/src/components/checkbox/checkboxListItem.tsx
```diff
@@ -27,6 +27,7 @@ export type CheckboxListItemProps = {
   multiSelect?: boolean;
   disabled?: boolean;
   state?: 'default' | 'active' | 'multi';
+  errorBorder?: boolean;
```
Fabricevladimir: I think this bad boy can just be `error` since it indicates when there's an error. Might even be feasible to have `error` as a fourth state. Anyway we can leave any refactors if necessary for later.

packages/web-app/src/containers/goLive/index.tsx
```diff
@@ -71,7 +82,13 @@ export const GoLiveFooter: React.FC = () => {
   const IsButtonDisabled = () =>
     !Object.values(reviewCheck).every(v => v === true);
```
Fabricevladimir, line 83: Can you modify this so that the button is disabled if at least one of the values is not true? So that you can uncomment line 100?

Fabricevladimir (changes requested):
Good work! Quite a few moving parts there. Left a couple of comments.

packages/ui-components/src/components/checkbox/checkboxListItem.tsx
WissenIstNacht: I think `state` was not the right word here. The `state` should really be `type` or `kind`. But yeah, otherwise agreed. But yeah, let's not open that can of worms now ^^'

packages/web-app/src/components/descriptionList/index.tsx
```diff
@@ -7,6 +12,9 @@ export type DescriptionListProps = {
   title: string;
   onEditClick?: () => void;
   editLabel?: string;
+  checkBoxError?: boolean;
+  checkBoxErrorMessage?: string;
+  checkedState?: CheckboxListItemProps['state'];
```
WissenIstNacht, line 17: Yeah, this is why this prop field really shouldn't be called state, hehe.

The `active` property is indeed a state, the others really aren't...

packages/web-app/src/components/descriptionList/index.tsx
```diff
@@ -32,21 +43,21 @@ export const DescriptionListContainer: React.FC<DescriptionListProps> = ({
...
-          />
-          <div className="flex-shrink-0 tablet:w-3/4">
-            <CheckboxSimple
+      {onChecked && (
+        <div className="ml-auto space-y-1.5 tablet:w-3/4">
+          <div className="tablet:flex">
+            <CheckboxListItem
               label="These values are correct"
               multiSelect
               onClick={onChecked}
```
WissenIstNacht: `onChecked?.()`. You're call, though (pun intended :D)

packages/web-app/src/components/descriptionList/index.tsx
WissenIstNacht: Yeah, this is why this prop field really shouldn't be called `state`, hehe.

The `active` property is indeed a `state`, the others really aren't...

packages/web-app/src/components/descriptionList/index.tsx
```diff
@@ -32,21 +43,21 @@ export const DescriptionListContainer: React.FC<DescriptionListProps> = ({
...
-            mode="ghost"
-            onClick={onEditClick}
-          />
-          <div className="flex-shrink-0 tablet:w-3/4">
-            <CheckboxSimple
+      {onChecked && (
+        <div className="ml-auto space-y-1.5 tablet:w-3/4">
+          <div className="tablet:flex">
+            <CheckboxListItem
               label="These values are correct"
```
WissenIstNacht: i18n

WissenIstNacht (changes requested):
If it's not too much work, could you separate the `type` of the checkboxes from their `state`?

packages/web-app/src/containers/goLive/index.tsx
RakeshUP: When a button is disabled, the onClick is not called; validations are impossible. We have to come up with new variants or introduce a new prop to capture the 'primary-disabled', etc., kinds of buttons

Fabricevladimir: A quick workaround in the meantime could be to wrap the button in a div which will have the disabled click handler. When the button is disabled, the click event will be actually on the parent div.  Something like the following, unless of course it's quicker to change the button styling:

```jsx
<div onClick={showInvalidFields}>
    <button onClick={goToNextPage}/>
</div>
```

WissenIstNacht (approved):
Thanks for addressing the comments, approving the PR.

Just as a sidenote (no need for action on this): Maybe I expressed my self poorly. What I meant for the interface of the checkboxes in the component library is that we have a property called `state` that holds `'active' | 'error' | 'warning' | ...`, _as well as_ a property called `type` that holds `'single' | 'multi' | ...`.

packages/web-app/src/containers/goLive/index.tsx
RakeshUP: Implemented your solution. Thanks @Fabricevladimir

Fabricevladimir (approved):

PR #459, opened 2022-08-08
Feature: APP-602 - New Cookie Modals

Fabricevladimir:
## Description
- Implements reworked privacy policy & cookie setting modals
- Removes `draft` filter from Governance page

Task: [APP-602](https://aragonassociation.atlassian.net/browse/APP-602), [APP-743](https://aragonassociation.atlassian.net/browse/APP-743)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
+++ b/packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
@@ -26,10 +26,7 @@ const CookieSettingsMenu: React.FC<CookieSettingsMenuProps> = props => {
   const [functionalSelected, setFunctionalSelected] = useState<boolean>(true);
 
   return (
-    <Modal
-      show={props.show}
-      className="fixed bottom-0 z-50 w-full bg-ui-50 rounded-t-xl"
-    >
+    <Modal show={props.show}>
       <div>
         <ModalHeader>
           <ButtonIcon
@@ -107,7 +104,11 @@ type Props = {
 };
 
 const Modal = styled.div.attrs(({show}: Props) => ({
-  className: `${show ? 'block' : 'hidden'}` as string | undefined,
+  className: `${show ? 'block' : 'hidden'} fixed z-50 bg-ui-50
+  bottom-0 w-full rounded-t-xl
+  tablet:bottom-3 tablet:left-1/2 tablet:w-56 tablet:rounded-b-xl
+  tablet:transform tablet:-translate-x-1/2
+  desktop:bottom-auto desktop:top-12`,
 }))<Props>`
   box-shadow: 0px 24px 32px rgba(31, 41, 51, 0.04),
     0px 16px 24px rgba(31, 41, 51, 0.04), 0px 4px 8px rgba(31, 41, 51, 0.04),
--- a/packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
+++ b/packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
@@ -8,8 +8,6 @@ import styled from 'styled-components';
 import {useTranslation} from 'react-i18next';
 import React, {useState} from 'react';
 
-import ModalBottomSheetSwitcher from 'components/modalBottomSheetSwitcher';
-
 export type CookiesType = {
   analytics: boolean;
   functional: boolean;
@@ -28,10 +26,9 @@ const CookieSettingsMenu: React.FC<CookieSettingsMenuProps> = props => {
   const [functionalSelected, setFunctionalSelected] = useState<boolean>(true);
 
   return (
-    <ModalBottomSheetSwitcher
-      isOpen={props.show}
-      onClose={props.onClose}
-      onOpenAutoFocus={e => e.preventDefault()}
+    <Modal
+      show={props.show}
+      className="fixed bottom-0 z-50 w-full bg-ui-50 rounded-t-xl"
     >
       <div>
         <ModalHeader>
@@ -84,18 +81,18 @@ const CookieSettingsMenu: React.FC<CookieSettingsMenuProps> = props => {
           </div>
         </BottomSheetContentContainer>
       </div>
-    </ModalBottomSheetSwitcher>
+    </Modal>
   );
 };
 
 export default CookieSettingsMenu;
 
 const Title = styled.div.attrs({
-  className: 'flex-1 font-bold text-center text-ui-800',
+  className: 'flex-1 font-bold text-center ft-text-base text-ui-800',
 })``;
 
 const ModalHeader = styled.div.attrs({
-  className: 'flex items-center p-2 space-x-2 bg-ui-0 rounded-xl sticky top-0',
+  className: 'flex items-center p-2 space-x-2 bg-ui-0 rounded-xl',
 })`
   box-shadow: 0px 4px 8px rgba(31, 41, 51, 0.04),
     0px 0px 2px rgba(31, 41, 51, 0.06), 0px 0px 1px rgba(31, 41, 51, 0.04);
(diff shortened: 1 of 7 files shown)


RakeshUP (approved):
LGTM

packages/web-app/src/containers/privacyPolicy/index.tsx
```diff
@@ -1,69 +1,49 @@
...
   const {isDesktop} = useScreen();
-  const [showCookieSettings, setShowCookieSettings] = useState<boolean>(false);
 
   if (!showPolicy) return null;
 
   return (
     <>
       {isDesktop ? (
-        <div className="fixed bottom-5 w-full">
+        <div className="fixed bottom-2 w-full">
```
WissenIstNacht, line 27: Does it mess up the layout if you merge this into the container...?

packages/web-app/src/context/privacyContext.tsx
```diff
@@ -170,20 +191,29 @@ const PrivacyContextProvider: React.FC = ({children}) => {
...
 
+  /*************************************************
+   *                    Render                     *
+   *************************************************/
   return (
     <PrivacyContext.Provider value={value}>
       {children}
       <PrivacyPolicy
-        showPolicy={!policyAccepted}
+        showPolicy={showPolicyMenu}
```
WissenIstNacht, line 201: I see there's older code in here, so this is probably a bit out of scope. Still gonna mention it: The structuring of all this is a bit weird to me. There's a context here. Yet, there's still props being passed down everywhere? In the menus, couldn't you just access the context's stuff via the corresponding `useContext` you wrapped below?

packages/web-app/src/containers/privacyPolicy/privacyPolicyContent.tsx
```diff
@@ -57,7 +64,7 @@ const PrivacyPolicyContent: React.FC<PrivacyPolicyContentProps> = ({
 export default PrivacyPolicyContent;
 
 const Text = styled.div.attrs({
-  className: 'flex-1 desktop:text-sm text-ui-600',
+  className: 'flex-1 ft-text-sm text-ui-600',
```
WissenIstNacht, line 67: nice catch

packages/web-app/src/containers/privacyPolicy/privacyPolicyContent.tsx
```diff
@@ -27,27 +27,34 @@ const PrivacyPolicyContent: React.FC<PrivacyPolicyContentProps> = ({
...
+          label={t('privacyPolicy.acceptAllCookies')}
+          mode="secondary"
+          bgWhite
           {...(isDesktop
-            ? {mode: 'secondary', size: 'small'}
-            : {mode: 'ghost', size: 'large', className: 'w-full'})}
-          onClick={props.onShowCookieSettings}
+            ? {size: 'small'}
+            : {size: 'large', className: 'w-full'})}
+          onClick={props.onAcceptAll}
```
WissenIstNacht, line 36: Btw, what's your policy on the whole (not) spreading props subject? :D

packages/web-app/src/context/privacyContext.tsx
```diff
@@ -170,20 +191,29 @@ const PrivacyContextProvider: React.FC = ({children}) => {
...
         onRejectAll={rejectAll}
-        onAcceptPolicy={setPrivacyPolicy}
+        onShowCookieSettings={handleShowCookiesSettings}
       />
       <CookiePreferenceMenu
         show={showPreferenceMenu}
         onClose={preferenceMenuCallbacks.onReject}
         onAccept={preferenceMenuCallbacks.onAccept}
       />
+      <CookieSettingsMenu
```
WissenIstNacht, line 211: Where the heck was this called before? 😅

packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
```diff
@@ -28,74 +26,68 @@ const CookieSettingsMenu: React.FC<CookieSettingsMenuProps> = props => {
...
+      <BottomSheetContentContainer>
+        <div className="space-y-1.5">
+          <CheckboxListItem
+            label={t('privacyPolicy.functional')}
+            state={functionalSelected ? 'active' : 'default'}
+            onClick={() => setFunctionalSelected(!functionalSelected)}
+            helptext={t('privacyPolicy.functionalHelpText')}
+            multiSelect
+          />
+          <CheckboxListItem
```
WissenIstNacht: Did you plan to wait for Rakesh's PR to land for this?

packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
```diff
@@ -28,74 +26,68 @@ const CookieSettingsMenu: React.FC<CookieSettingsMenuProps> = props => {
...
+            state={analyticsSelected ? 'active' : 'default'}
+            onClick={() => setAnalyticsSelected(!analyticsSelected)}
+            helptext={t('privacyPolicy.analyticsHelpText')}
+            multiSelect
+          />
+        </div>
+        <div className="flex space-x-2">
+          <ButtonText
+            className="flex-1"
+            label={t('privacyPolicy.acceptSelectedCookies')}
```
WissenIstNacht: I know you have your orders. Just food for thought, though. If you unselect both options, this becomes a bit weird, because the button still states "Accept". I think your code does the right thing and rejects both. But yeah it feels weird to click on "Accept". Maybe "Select" would be more appropriate. But yeah, worse cse stick to the designs on this. Can figure it our later.

WissenIstNacht (approved):
Hey there, looking good!

I left a couple of random comments here and there. Curious to know what you think on some of them, but nothing that requires immediate attention.

packages/web-app/src/containers/privacyPolicy/index.tsx
Fabricevladimir: Not exactly, but I would have to center it differently with the transformations, and since there is no clicking that can be done on the horizontal axis next to this modal, I kept this. For the other modal, the centering is done differently so that the div doesn't intercept click events.

packages/web-app/src/context/privacyContext.tsx
Fabricevladimir: Hahaha, yes this is true. The intention is for the context values to be used anywhere in the app, and since there was minimal props drilling, as the menus are direct children and need other props that are not exported from the context, I went ahead and passed down those props.

packages/web-app/src/containers/privacyPolicy/privacyPolicyContent.tsx
Fabricevladimir: I try to avoid it if possible, but that honestly is purely for the sake of readability. Usually I go that route if there are multiple properties relying on the same condition. So instead of checking the condition twice like so:
```jsx
<Element size={condition ? "small" : "large"} className={condition ? " ": "w-full"} />
```
AFAIK there isn't a performance downside or anything, plus, I'm only spreading *specific* props so nothing unexpected gets passed in.

packages/web-app/src/context/privacyContext.tsx
Fabricevladimir: It was inside the `PrivacyPolicy` and because there was previously a backdrop, when it opened, it would actually hide the `PrivacyPolicy` menu which meant I didn't need to take care of closing the `PrivacyPolicy` menu when opening the `CookieSettingsMenu`. But now that backdrops are gone, I had to pull this out with it's open/close state so the context can manage both properly.

packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
Fabricevladimir: Yeah, it's preferable if his merges so I can rebase and make any necessary modifications

packages/web-app/src/containers/privacyPolicy/cookieSettingsMenu.tsx
Fabricevladimir: This is a good point. The entire mechanic can get a bit weird. I'll bring this up with design people. That said, I think it's meant to be an abbreviation of "Accept these settings, which just so happen to be no permissions given!"

packages/web-app/src/containers/privacyPolicy/privacyPolicyContent.tsx
WissenIstNacht: Absolutely, this is purely a style thing.

PR #464, opened 2022-08-11
App 722 - Limiting Proposal Actions

WissenIstNacht:
## Description

This PR limits the number of certain action in the configure action step. This includes:
- Disabling actions in the `ActionMenu` if the limit is reached
- Removing the option to duplicate certain actions (this was only the case for minting token, AFAIK)

Additionally, this PR contains a bit of renaming and restructuring:
- The action context no longer has a `daoAddress` state (which was not used anyway, btw). Instead, in now receives this as a prop. The reason for this is that it should NEVER be able to have a different value in the context then outside of it.
- Any index that indexes an `action` from the form context is now called `actionIndex`.
- Renamed `actionCounter` because it doesn't count actions. (thanks for the headaches, haha)

Task: [722](https://aragonassociation.atlassian.net/browse/APP-722), [792](https://aragonassociation.atlassian.net/browse/APP-792)

## Type of change

### DIFF
--- a/packages/web-app/src/containers/actionBuilder/mintTokens/addressTokenRow.tsx
+++ b/packages/web-app/src/containers/actionBuilder/mintTokens/addressTokenRow.tsx
@@ -16,9 +16,9 @@ import styled from 'styled-components';
 
 import useScreen from 'hooks/useScreen';
 import {handleClipboardActions} from 'utils/library';
+import {ActionIndex} from 'utils/types';
 
-type IndexProps = {
-  actionIndex: number;
+type IndexProps = ActionIndex & {
   fieldIndex: number;
 };
 
@@ -94,9 +94,9 @@ const TokenField: React.FC<IndexProps> = ({actionIndex, fieldIndex}) => {
   );
 };
 
-const DropdownMenu: React.FC<
-  Omit<AddressAndTokenRowProps, 'newTokenSupply'>
-> = ({fieldIndex, onDelete}) => {
+type DropdownProps = Omit<AddressAndTokenRowProps, 'newTokenSupply'>;
+
+const DropdownMenu: React.FC<DropdownProps> = ({fieldIndex, onDelete}) => {
   const {t} = useTranslation();
 
   return (
--- a/packages/web-app/src/containers/actionBuilder/mintTokens/addressTokenRow.tsx
+++ b/packages/web-app/src/containers/actionBuilder/mintTokens/addressTokenRow.tsx
@@ -94,30 +94,31 @@ const TokenField: React.FC<IndexProps> = ({actionIndex, fieldIndex}) => {
   );
 };
 
-const DropdownMenu: React.FC<Omit<AddressAndTokenRowProps, 'newTokenSupply'>> =
-  ({fieldIndex, onDelete}) => {
-    const {t} = useTranslation();
+const DropdownMenu: React.FC<
+  Omit<AddressAndTokenRowProps, 'newTokenSupply'>
+> = ({fieldIndex, onDelete}) => {
+  const {t} = useTranslation();
 
-    return (
-      <Dropdown
-        align="start"
-        trigger={
-          <ButtonIcon mode="ghost" size="large" icon={<IconMenuVertical />} />
-        }
-        sideOffset={8}
-        listItems={[
-          {
-            component: (
-              <ListItemAction title={t('labels.removeWallet')} bgWhite />
-            ),
-            callback: () => {
-              onDelete(fieldIndex);
-            },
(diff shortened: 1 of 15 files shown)


RakeshUP (approved):

packages/web-app/src/containers/actionBuilder/mintTokens/addressTokenRow.tsx
```diff
@@ -94,30 +94,31 @@ const TokenField: React.FC<IndexProps> = ({actionIndex, fieldIndex}) => {
   );
 };
 
-const DropdownMenu: React.FC<Omit<AddressAndTokenRowProps, 'newTokenSupply'>> =
-  ({fieldIndex, onDelete}) => {
-    const {t} = useTranslation();
+const DropdownMenu: React.FC<
```
Fabricevladimir: This is driving me crazy. If I haven't linted this page 170145725 times already...

packages/web-app/src/containers/actionBuilder/mintTokens/index.tsx
```diff
@@ -32,21 +31,17 @@ type AddressBalance = {
...
+  const {removeAction} = useActionsContext();
   const {setValue} = useFormContext();
 
   const handleReset = () => {
-    setValue(`actions.${index}.inputs.mintTokensToWallets`, []);
+    setValue(`actions.${actionIndex}.inputs.mintTokensToWallets`, []);
   };
 
   const methodActions = [
-    {
```
Fabricevladimir, line 46: There was a task on the backlog for this. Add the ID in the PR description so we don't forget to drag the ticket through the process?

packages/web-app/src/containers/configureWithdraw/index.tsx
```diff
@@ -58,12 +57,12 @@ const ConfigureWithdrawForm: React.FC<ConfigureWithdrawFormProps> = ({
   const [name, from, tokenAddress, isCustomToken, tokenBalance, symbol] =
     useWatch({
       name: [
-        `actions.${index}.name`,
-        `actions.${index}.from`,
-        `actions.${index}.tokenAddress`,
-        `actions.${index}.isCustomToken`,
-        `actions.${index}.tokenBalance`,
-        `actions.${index}.tokenSymbol`,
+        `actions.${actionIndex}.name`,
```
Fabricevladimir, line 60: We done gone and screwed up. None of these fields are inside the action's inputs.

packages/web-app/src/pages/newProposal.tsx
```diff
@@ -53,7 +50,7 @@ const NewProposal: React.FC = () => {
         showTxModal={showTxModal}
         setShowTxModal={setShowTxModal}
       >
-        <ActionsProvider>
+        <ActionsProvider daoId={dao}>
```
Fabricevladimir, line 53: Are we actually using daoId for anything? IF I recall correctly we had it as an eternal metadata prop that we could just get off of here and add to the form when necessary, but it got forgotten I think?

Fabricevladimir (approved):
Looks good!

packages/web-app/src/containers/configureWithdraw/index.tsx
WissenIstNacht: Where are they, then? 😅 Everything seemed fine to me when I went through the flow?

packages/web-app/src/pages/newProposal.tsx
WissenIstNacht: I don't know, tbh. It's currently not being used, so technically we could remove it. I'm all for it, but up to you guys.

packages/web-app/src/containers/actionBuilder/mintTokens/index.tsx
WissenIstNacht: Oh, ok! I kinda modified 722 to fit it in there 😂

But I'll link the other task as well then. Thanks for the catch!

packages/web-app/src/containers/configureWithdraw/index.tsx
Fabricevladimir: I meant the values aren't structured the same as others (should be `actions.${index}.inputs.tokenSymbol`). It's not a big deal; that's probably because Finance Withdraw was implemented before we considered it as an action for the main proposal flow.

packages/web-app/src/pages/newProposal.tsx
Fabricevladimir: I'm for removing it as well. If the form needs dao address, we can use the `useDaoParam` hook.

PR #465, opened 2022-08-16
Feature: APP-869 Vote Now CTA

Fabricevladimir:
## Description
- adds different voting terminal cta based on login and membership status of a user when viewing a proposal

NOTE: For reviewers, if there isn't a currently active proposal, feel free to directly set `status` to `active` in the `useDaoProposal` hook.

Task: [APP-869](https://aragonassociation.atlassian.net/browse/APP-869)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/containers/reviewProposal/index.tsx
+++ b/packages/web-app/src/containers/reviewProposal/index.tsx
@@ -226,7 +226,7 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
           <VotingTerminal
             breakdownTabDisabled
             voteNowDisabled
-            statusLabel={t('votingTerminal.notStartedYet')}
+            statusLabel={t('votingTerminal.status.draft')}
             approval={approval}
             participation={participation}
             startDate={formattedStartDate}
--- a/packages/web-app/src/containers/votingTerminal/index.tsx
+++ b/packages/web-app/src/containers/votingTerminal/index.tsx
@@ -43,6 +43,7 @@ export type VotingTerminalProps = {
   votingInProcess?: boolean;
   onVoteClicked?: React.MouseEventHandler<HTMLButtonElement>;
   onCancelClicked?: React.MouseEventHandler<HTMLButtonElement>;
+  voteButtonLabel?: string;
 };
 
 export const VotingTerminal: React.FC<VotingTerminalProps> = ({
@@ -61,6 +62,7 @@ export const VotingTerminal: React.FC<VotingTerminalProps> = ({
   onVoteClicked,
   votingInProcess,
   onCancelClicked,
+  voteButtonLabel,
 }) => {
   const [query, setQuery] = useState('');
   const [buttonGroupState, setButtonGroupState] = useState('info');
@@ -240,7 +242,7 @@ export const VotingTerminal: React.FC<VotingTerminalProps> = ({
       ) : (
         <VoteContainer>
           <ButtonText
-            label={t('votingTerminal.voteNow')}
+            label={voteButtonLabel || t('votingTerminal.voteUnavailable')}
             size="large"
             onClick={onVoteClicked}
             className="w-full tablet:w-max"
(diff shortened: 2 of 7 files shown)


packages/web-app/src/pages/proposal.tsx
```diff
@@ -23,27 +23,47 @@ import ResourceList from 'components/resourceList';
...
+  const {data: canVote, isLoading: canVoteLoading} = useWalletCanVote(
+    daoId,
+    address!
+  );
+
+  // Note: these two refs being used to hold "memories" of previous "state"
+  // across renders in order to automate the following states:
+  // loggedOut -> login modal => switch network modal -> vote options selection;
+  // ref for holding value of whether user was previously not logged in when vote now was clicked
+  const wasNotLoggedIn = useRef(false);
```
WissenIstNacht: Nice trick :)

Maybe you thought about this yourself, but just as food for thought: Instead of using two boolean refs You might be able to  use only one ref that holds a more complex state.

packages/web-app/src/containers/reviewProposal/index.tsx
```diff
@@ -226,7 +226,7 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
           <VotingTerminal
             breakdownTabDisabled
             voteNowDisabled
-            statusLabel={t('votingTerminal.notStartedYet')}
+            statusLabel={t('votingTerminal.status.draft')}
```
WissenIstNacht, line 229: So we have drafts again now...? Wasn't that just "pending"? I don't get it anymore lol.

Update: Sorry, this essentially while you fill out the proposal? That would make sense.

packages/web-app/src/containers/votingTerminal/index.tsx
```diff
@@ -2,19 +2,26 @@ import React, {useMemo, useState} from 'react';
...
-import {ButtonGroup, Option} from '@aragon/ui-components/src/components/button';
 import {
   VotersTable,
   VoterType,
 } from '@aragon/ui-components/src/components/table';
-import {CheckboxListItem} from '@aragon/ui-components/src/components/checkbox';
 import {useTranslation} from 'react-i18next';
-import {IconClock, StateEmpty} from '@aragon/ui-components';
 
+// TODO: clean up props: some shouldn't be optional
```
WissenIstNacht, line 25: Also, it feels like most of this has already been typed out somewhere.

packages/web-app/src/hooks/useWalletCanVote.tsx
```diff
@@ -0,0 +1,63 @@
...
+import {CHAIN_METADATA} from 'utils/constants';
+import {useSpecificProvider} from 'context/providers';
+import {isWhitelistMember, useDaoMembers} from './useDaoMembers';
+import {HookData} from 'utils/types';
+
+// Hook to determine if connected address is eligible to vote on proposals of the current
+// DAO. Note that for ERC20, voting member is anyone holding the token
+// and not just those who have previously interacted with the DAO, obviously.
+// This of course will need to be modified for proper voter eligibility
+export const useWalletCanVote = (
```
WissenIstNacht, line 15: There was something similar on Sepehr's PR and I commented on there, too. What are your thoughts on doing actions like this for each type DAO specifically, vs over all types at the same time? Or in this concrete case, What do you think about breaking this ability to vote for erc20 daos ("case 1") and wallet based  daos ("case 2")? It seems like a bit of a waste to do both checks every time. In particular, since the parent that calls this is probably already aware of which case it is. Of course this could also be solved by adding an argument as prop and then switching inside the hook.

Or maybe it's actually ok the way it is. Let me know your thoughts. Would be nice to do things similarly across the app.

WissenIstNacht (approved):
Thanks, lgtm.

I've left some comments, but except from replying, none acutally require any action.

packages/web-app/src/containers/reviewProposal/index.tsx
Fabricevladimir: Yes, I know we're bringing back drafts at some point. So while we're filling out the proposal it makes more sense to have it as a draft since it hasn't actually been published yet

RakeshUP (approved):

PR #468, opened 2022-08-18
Fix: New SDK integration & enabled deposit and dao creation flow

sepehr2github:
## Description

Note1: Proposal Creation wasn't on the area of this task
Note2: The creation process is mocked so metamask won't trigger anymore

Task: [APP-681](https://aragonassociation.atlassian.net/browse/APP-681)

## Type of change

## Checklist:

- [] I have updated the ``CHANGELOG.md`` file in the root folder.


packages/web-app/src/context/createDao.tsx
```diff
@@ -26,8 +28,7 @@ import {useWallet} from 'hooks/useWallet';
 import {useGlobalModalContext} from './globalModals';
 import {useClient} from 'hooks/useClient';
 import {usePollGasFee} from 'hooks/usePollGasfee';
-
-type DAOCreationSettings = ICreateDaoERC20Voting | ICreateDaoWhitelistVoting;
+import {IPluginListItem} from '@aragon/sdk-client/dist/internal/interfaces/common';
```
Fabricevladimir: We need to get this exported properly from the SDK

packages/web-app/src/context/createDao.tsx
```diff
@@ -139,97 +116,148 @@ const CreateDaoProvider: React.FC<Props> = ({children}) => {
...
-        gsnForwarder: constants.AddressZero,
+  const {tokenPrice, maxFee, averageFee, stopPolling} = usePollGasFee(
+    estimateCreationFees,
+    shouldPoll
+  );
 
-        whitelistVoters: values.whitelistWallets.map(wallet => wallet.address),
-      };
-    }, [getDaoConfig, getValues, getVotingConfig]);
+  const createDaoIterator = useMemo(() => {
```
Fabricevladimir: Why go with an memoized iterator instead of calling the method in the `createDao` function underneath?

packages/web-app/src/context/createDao.tsx
```diff
@@ -139,97 +116,148 @@ const CreateDaoProvider: React.FC<Props> = ({children}) => {
...
+    return client?.methods.create(getDaoSettings());
```
Fabricevladimir: In `handlePublishDao`, we called `getDaoSettings` and set the result to the `daoCreationData` state object. Shouldn't `daoCreationData` be used here and in `estimateCreationFees` instead of calling `getDaoSettings` again?

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -1,59 +1,57 @@
...
-  createErc20: (dao: ICreateDaoERC20Voting) => Promise<string>;
-  createWhitelist: (dao: ICreateDaoWhitelistVoting) => Promise<string>;
+  createErc20: (pluginAddress: Address) => ClientErc20 | Error;
+  createWhitelist: (pluginAddress: Address) => ClientAddressList | Error;
 }
 
+/**
+ * This hook can use to build ERC20 or whitelist clients
+ * @method createErc20 By passing instance plugin address will create an ERC20Client
+ * @method createWhitelist By passing instance plugin address will create an ERC20Client
```
Fabricevladimir: Update comment to to WhitelistClient instead of ERC20Client. Copy pasta, I presume?

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -1,59 +1,57 @@
...
+ * @returns The corresponding Client
+ */
+
 export const useDao = (): IUseDaoResponse => {
```
Fabricevladimir: Might need to rename this so it's more intuitive & matches the filename like we have per convention. I guess this is a client factory, yes?

Fabricevladimir (commented):
Thanks, man! Left a couple of comments

packages/web-app/src/hooks/useClient.tsx
```diff
@@ -30,34 +27,44 @@ export const useClient = () => {
...
-      const context = new SdkContext({
+      const contextParams: ContextParams = {
         network: 'rinkeby', // TODO: remove temporarily hardcoded network
         signer,
         web3Providers: web3Providers
           ? web3Providers.split(',')
-          : [
-              'https://eth-rinkeby.alchemyapi.io/v2/bgIqe2NxazpzsjfmVmhj3aS3j_HZ9mpr',
-            ],
+          : [import.meta.env.VITE_IPFS_ALCHEMY_API as string],
```
WissenIstNacht: The alchemy import makes sense. Do you remember why Jose added `web3Providers`? Or what exactly it holds? And why we're only using it conditionally?

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -1,59 +1,57 @@
...
 
 interface IUseDaoResponse {
-  createErc20: (dao: ICreateDaoERC20Voting) => Promise<string>;
-  createWhitelist: (dao: ICreateDaoWhitelistVoting) => Promise<string>;
+  createErc20: (pluginAddress: Address) => ClientErc20 | Error;
+  createWhitelist: (pluginAddress: Address) => ClientAddressList | Error;
 }
 
+/**
+ * This hook can use to build ERC20 or whitelist clients
```
WissenIstNacht:
```suggestion
 * This hook can be used to build ERC20 or whitelist clients
```

packages/web-app/src/hooks/useCachedDao.tsx
WissenIstNacht: +1 on this. I would go even further an ask if this even still the right approach to this (I'm sure this might have made sense at some point, but I feel like a lot has changed).

What is the idea of this code? Is it here for convenience? I.e., to make it for the devs that rely on this global Client context to get the exact (sub-) client that they need? 

Also, more general thoughts:
1) why return a function instead of directly returning the `ClientXYZ` instance directly? (UPDATE I think I find out why. But I think it's the wrong way to go. `pluginAddress` should be a prop. Then we would not need to return functions).
2) Why not create two hooks, one for each client? Actually maybe this is not a good idea, because then we'd have to extend this manually for each new type of client. But still, I think returning both every time might not be necessary.

packages/web-app/src/hooks/useCachedDao.tsx
```diff
@@ -1,59 +1,57 @@
...
-  const createErc20 = (dao: ICreateDaoERC20Voting): Promise<string> => {
-    if (!erc20Client) {
-      return Promise.reject(
-        new Error('ERC20 SDK client is not initialized correctly')
-      );
+  const {client, context} = useClient();
+
+  const createErc20 = (pluginAddress: Address): ClientErc20 | Error => {
+    if (!client || !context) {
+      return new Error('ERC20 SDK client is not initialized correctly');
```
WissenIstNacht: I think the error is no longer correct now. Since we're only creating the erc20 client now :p

PS: same for whitelist further down

packages/web-app/src/context/deposit.tsx
```diff
@@ -58,21 +58,21 @@ const DepositProvider = ({children}: {children: ReactNode}) => {
...
-        return client?.estimate.increaseAllowance(
-          depositParams.token as string,
-          depositParams.daoAddress,
-          depositParams.amount
-        );
-    }
-  }, [client, currentStep, depositParams]);
+    // if (client && depositParams) {
+    // if (currentStep === 2 || isNativeToken(depositParams.token as string)) {
+    return client?.estimation.deposit(depositParams as IDepositParams);
```
WissenIstNacht: Do we not need the code around this any more?

WissenIstNacht (changes requested):
Nice! Eventhough this might break a thing or two, I'm kinda excited to introduce this. :) 

I've requested a couple of minro chages. I've also left a couple of questions. Please get back to me on them.

WissenIstNacht:
Also my `yarn.lock` is different then yours. Did you install `dot-env` at some point? If so, make sure to clean your lock file (or let me know and I can push the necessary changes).

packages/web-app/src/context/deposit.tsx
sepehr2github: Yes, We need but the token allowance estimation wasn't available on SDK, so we will fix it once the estimation method was available

packages/web-app/src/hooks/useClient.tsx
sepehr2github: Sorry my bad, Fixed the env configuration

packages/web-app/src/hooks/useCachedDao.tsx
sepehr2github: The method refactored Also as an answer, Yes It's for convenience also It's for getting the exact sub-client

packages/web-app/src/hooks/usePluginClient.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * This hook can be used to build ERC20 or whitelist clients
+ * @method createErc20 By passing instance plugin address will create an ERC20Client
+ * @method createWhitelist By passing instance plugin address will create an WhitelistClient
+ * @returns The corresponding Client
+ */
+
+export const usePluginClient = (
+  type: ClientTypes,
```
WissenIstNacht, line 20: Don't be afraid Sepehr... use the power! :D

No but seriously, we're using Typescript for a reason. Please use actual type instead of Enums, as Enums can not be statically checked. Thanks :)

packages/web-app/src/hooks/useClient.tsx
```diff
@@ -30,34 +27,46 @@ export const useClient = () => {
...
+  const [client, setClient] = useState<Client>();
+  const [context, setContext] = useState<SdkContext>();
 
   useEffect(() => {
     if (signer) {
-      const web3Providers = import.meta.env
-        .VITE_REACT_APP_SDK_WEB3_PROVIDERS as string;
+      const alchemyApiAddress = import.meta.env
+        .VITE_REACT_APP_ALCHEMY_API_KEY as string;
+      const alchemyApiKey = import.meta.env
```
WissenIstNacht: Small comment here: The `.env` file should only contain what can not be written in code. This can basically be two thing:
1) Something that you don't know while you write code (for example, the sha of the commit that will trigger a build of this code. You just can not know this while you write code).
2) Something that you can not share in your source code (i.e., passwords, api keys, etc.)

From this two points we can deduce that this is not the case for the URL:
 1) We already know the URL
 2) There is nothing private about the URL.

WissenIstNacht (approved):
Thanks a lot for the changes Sepehr.

I left two additional minor comments. Would  be nice if you could take care of them. The PR is already approved, because I won't be available tomorrow. Just merge it once you addressed the two comments.

packages/web-app/src/hooks/useClient.tsx
sepehr2github: I agree with this idea but Carlos told me to put all URLs and variables on env, but let me do a double sync will update

WissenIstNacht: Then I apologies for giving you contradictory feedback! 🙏🏼

sepehr2github: No, thats ok. He's reason was that   Just Imagine a server goes down, you would need to make change, put up a PR, get two approvals, pass CI/CD, then wait for deployment and finally check. That's at least 30 mins you are down, or with something broken. That doesn't make sense, when you could just change it in a variable, but for now we will only add api keys to env files then add urls later
@WissenIstNacht

Fabricevladimir (approved):
Thanks Sepehr

PR #469, opened 2022-08-18
Chore: Analytics for Create DAO

RakeshUP:
## Description

Adds rudderstack analytics to createDAO flow

Task: [APP-826](https://aragonassociation.atlassian.net/browse/APP-826)

## Type of change

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ yarn-error.log*
 .env.development.local
 .env.test.local
 .env.production.local
+.env
 
 # deployments
 deployments
--- a/packages/web-app/src/components/fullScreenStepper/step.tsx
+++ b/packages/web-app/src/components/fullScreenStepper/step.tsx
@@ -12,7 +12,7 @@ export type StepProps = {
   nextButtonLabel?: string;
   isNextButtonDisabled?: boolean;
   onBackButtonClicked?: () => void;
-  onNextButtonClicked?: () => void;
+  onNextButtonClicked?: (next: () => void) => void;
   // This method can use to trigger validations once user clicked on disabled next button
   onNextButtonDisabledClicked?: () => void;
 };
--- a/packages/web-app/src/hooks/useWallet.tsx
+++ b/packages/web-app/src/hooks/useWallet.tsx
@@ -15,6 +15,7 @@ export interface IUseWallet extends SignerValue {
    * the network name returned by the useNetwork context).
    */
   isOnWrongNetwork: boolean;
+  network: string;
 }
 
 export const useWallet = (): IUseWallet => {
@@ -66,5 +67,6 @@ export const useWallet = (): IUseWallet => {
     isConnected,
     isOnWrongNetwork,
     methods,
+    network,
   };
 };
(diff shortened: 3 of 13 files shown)


packages/web-app/src/pages/createDAO.tsx
```diff
@@ -225,20 +226,52 @@ const CreateDAO: React.FC = () => {
                 />
               </>
             }
+            onNextButtonClicked={next => {
```
sepehr2github: I don't want to be sensitive, But that would be great if you can define a new method with next inside it, so the only thing you need to do is to pass the config and avoid duplicating the next or event name

.gitignore
```diff
@@ -29,6 +29,7 @@ yarn-error.log*
 .env.development.local
 .env.test.local
 .env.production.local
+.env
```
sepehr2github, line 32: where did you use the environment variables?

sepehr2github (approved):
Hey, Rakash. The Pr was good to me but I left a few comments and questions

.gitignore
RakeshUP: It was already being used in the `analytics.ts` file. I just wanted to add to gitignore

packages/web-app/src/hooks/useWallet.tsx
```diff
@@ -66,5 +67,6 @@ export const useWallet = (): IUseWallet => {
     isConnected,
     isOnWrongNetwork,
     methods,
+    network,
```
WissenIstNacht, line 70: Hey just a word of caution on this: The `network` you return here is the one defined by the app. Is it possible you would actually want the wallet's network? In that case you'd probably want to transform `chainId` into a network and return that. Also, could you call this `walletNetwork` just to avoid people confusing the two?

packages/web-app/src/components/fullScreenStepper/step.tsx
```diff
@@ -12,7 +12,7 @@ export type StepProps = {
   nextButtonLabel?: string;
   isNextButtonDisabled?: boolean;
   onBackButtonClicked?: () => void;
-  onNextButtonClicked?: () => void;
+  onNextButtonClicked?: (next: () => void) => void;
```
WissenIstNacht, line 15: Oh man, that's really going next order haha 👍🏻

Don't bother with too much with this, I'm just thinking aloud. But in the actual stepper, couldn't we just assume that `next()` called and call it in addition to `onNextButtonClicked`, instead of nesting it into it? Or would that lead to trouble when doing it on the last step?

WissenIstNacht (approved):
Thanks a lot. Check the wallet network thingy, but aside from that it's good to go!

packages/web-app/src/components/fullScreenStepper/step.tsx
RakeshUP: Yes, I thought so too. Might create problems in the last step.

packages/web-app/src/hooks/useWallet.tsx
RakeshUP: The `network` is okay I guess. We use this `network` in the URL and I figured why not use it in analytics too?

PR #475, opened 2022-08-25
Feature: APP-918 - Execution Terminal Status

Fabricevladimir:
## Description
- Sets proper status for execution terminal based on the proposal status, wallet connection and voter eligibility status
- Refactors Execution Terminal to match new designs
- Adds empty state for Execution Terminal

**Execution Terminal is broken for main Withdraw flow as well**
Also, no action is being passed to Execution Terminal on proposal details page on purpose because we are waiting on SDK method for action decoding.

Task: [APP-918](https://aragonassociation.atlassian.net/browse/APP-918)

## Type of change

## Checklist:


packages/web-app/src/locales/en/common.json
```diff
@@ -167,8 +167,7 @@
     "newTokens": "New Tokens",
     "newHolders": "New Holders",
     "totalTokens": "Total Tokens",
-    "totalHolders": "Total Holders",
-    "seeCommunity": "See Community"
```
WissenIstNacht, line 171: Removing duplicates, nice.

packages/web-app/src/locales/en/common.json
```diff
@@ -218,7 +217,15 @@
     "noProposals": "No proposals found!",
     "executionCard": {
       "title": "Execution",
-      "description": "These smart actions are executed when the proposal reaches sufficient support. Find out which actions are executed."
+      "description": "These smart actions are executed when the proposal reaches sufficient support.",
+      "addAction": "Add Action",
```
WissenIstNacht, line 221: Adding duplicates, nice 😁

packages/web-app/src/pages/proposal.tsx
```diff
@@ -145,110 +154,126 @@ const Proposal: React.FC = () => {
...
-            };
-          }
-          break;
+  const [
+    voteNowDisabled,
+    statusLabel,
+    alertMessage,
+    executionStatus,
+    handleVoteClicked,
+  ] = useMemo(() => {
```
WissenIstNacht, line 163: Just a note on... let me call them "large" memos. In other words, memo where you compute a couple of values and then return them them together as an object:

I understand that it makes it somewhat cleaner to have everything in on place. However, keep in mind that you now perform a huge computation (whether this actually and expensive computation is an other discussion but not relevant here) in one go. This means that whenever _any_ of the dependencies change, the whole block get re-executed. And obviously the likelyhood of this happening actually increases with every dependeny you (are forced) to add. This kind of defeats the purpose. If you break them up you might have situation where one changes and triggers a couple of others to change. But at least you might have some that were not affected.

//Edit: Just to be clear, sometimes things are what they are, and you have to. I've not analysed the situation in full here. Maybe it's necessary. I only wanted to put the thought out there :D

WissenIstNacht (approved):
Thanks :)

RakeshUP (approved):

packages/web-app/src/locales/en/common.json
Fabricevladimir: It should be just a modification I think lol

packages/web-app/src/pages/proposal.tsx
Fabricevladimir: Fair observation. I would like to point out thought that in this case, nothing should change save for the login state/wrong network status. Plus, the meat of the memo are actually small `cases` of the `switch`.

PR #482, opened 2022-09-05
Feature: app-822 Enable Edit Status on Accordion

sepehr2github:
## Description

- Enable Edit Status on Accordion
- add reset field integration

Task: [APP-822](https://aragonassociation.atlassian.net/browse/APP-822)

## Type of change

## Checklist:


RakeshUP (approved):

packages/web-app/src/pages/editSettings.tsx
```diff
@@ -46,42 +46,111 @@ const EditSettings: React.FC = () => {
...
-    // TODO: Need to add community settings later
+    // TODO: Need to add community settings later, Also the Alerts share will be added later
     setValue(
       'membership',
       daoDetails?.plugins[0].id === 'erc20voting.dao.eth' ? 'token' : 'wallet'
     );
+  };
+
+  useEffect(() => {
+    if (
```
Fabricevladimir: Two things here:

- I think the name should be the dao name coming from the metadata instead of the dao ens because the core is what assigns the ens domain once a name is given. If we allow the user to change the dao ens domain itself, then we might run into all kinds of issues with the domain not matching up to what we expect. I would suggest reaching out to Selim & someone on Core about this for proper clarification.
- Because the DAOResources aren't being compared, changes in them aren't picked up.

packages/web-app/src/pages/editSettings.tsx
```diff
@@ -46,42 +46,111 @@ const EditSettings: React.FC = () => {
...
-    setValue,
-  ]);
+  };
 
   useEffect(() => {
-    setValue('minimumApproval', Math.round(daoSettings.minTurnout * 100));
+    setCurrentMetadata();
+    setCurrentGovernance();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [daoDetails, daoSettings]);
```
Fabricevladimir: I think if you add only `setCurrentMetadata` and `setCurrentGovernance` and wrap those in callbacks, this might get eslint to stop complaining and still run the effect when you desire it since when `daoDetails` and `daoSettings` change, the method setting them will change which will then trigger the useEffect.

Fabricevladimir (approved):

PR #483, opened 2022-09-05
Fix: App-842 - Token Summary Floating Point Imprecision

WissenIstNacht:
## Description

This PR:
- Introduces the `Big.js` library
- Adapts the token-minting flow to use `Big.js`

As can detailed in the bug ticket, this PR solves issues arising from the imprecision of floating point arithmetic.

Note:
The logic responsible for the number of new token holders broke in a past PR. Also, the 
"allocation" label is not properly aligned. Both to be fixed in dedicated bug tickets.

ts.

Task: [842](https://aragonassociation.atlassian.net/browse/APP-842)

## Type of change

### DIFF
--- a/packages/web-app/package.json
+++ b/packages/web-app/package.json
@@ -59,6 +59,7 @@
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.0.0",
     "@typechain/ethers-v5": "^9.0.0",
+    "@types/big.js": "^6.1.5",
     "@types/jest": "^26.0.24",
     "@types/node": "^16.3.2",
     "@types/react": "^17.0.14",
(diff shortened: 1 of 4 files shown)


packages/web-app/package.json
```diff
@@ -31,6 +31,7 @@
     "apollo": "^2.33.10",
     "apollo-link-rest": "^0.9.0-rc.1",
     "apollo3-cache-persist": "^0.14.0",
+    "big.js": "^6.2.1",
```
Fabricevladimir, line 34: What's the advantage of this over ~either hard limiting the number of decimals to a fixed value or~ using the `BigNumber` from `ethers`?

Fabricevladimir:
Also, linter is failing for some reason

RakeshUP (approved):

packages/web-app/package.json
WissenIstNacht: It does floating-point arithmetic :)

Fabricevladimir (approved):

PR #488, opened 2022-09-12
Feature: App 891 - Member List from SDK

WissenIstNacht:
## Description

This PR introduces the SDK as provider for data related to DAO members.

Note:
  - There's still basic but functional searching and pagination on the Community page. At some point, this will be moved to the SDK altogether.
  - The token for any ERC-20 DAO is _mocked_, as are all any balances shown for their members.

Task: [891](https://aragonassociation.atlassian.net/browse/APP-891)

## Type of change

### DIFF
--- a/packages/web-app/src/hooks/useDaoMembers.tsx
+++ b/packages/web-app/src/hooks/useDaoMembers.tsx
@@ -66,7 +66,9 @@ export const useDaoMembers = (
 
         const members =
           pluginType === 'erc20voting.dao.eth'
-            ? rawMembers.map(m => {
+            ? // TODO as soon as the SDK exposes Token information, fetch balances
+              // from contract.
+              rawMembers.map(m => {
                 return {
                   address: m,
                   balance: Math.floor(Math.random() * 500 + 1),
@@ -79,6 +81,7 @@ export const useDaoMembers = (
               });
         members.sort(sortMembers);
         setData(members);
+        setTotalMemberCount(members.length);
         setError(undefined);
       } catch (err) {
         console.error(err);
@@ -90,14 +93,6 @@ export const useDaoMembers = (
 
     fetchMembers();
   }, [client?.methods, pluginAddress, pluginType]);
-
-  // when search term is passed in, only set total on the first time,
-  // i.e. when totalMemberCount is null
-  useEffect(() => {
-    if (data && totalMemberCount === null) {
-      setTotalMemberCount(data.length);
-    }
-  }, [data, totalMemberCount]);
 
   return {
     data: {
--- a/packages/web-app/src/hooks/useDaoMembers.tsx
+++ b/packages/web-app/src/hooks/useDaoMembers.tsx
@@ -3,25 +3,30 @@ import {useEffect, useState} from 'react';
 import {HookData} from 'utils/types';
 import {PluginTypes, usePluginClient} from './usePluginClient';
 
-export type DaoWhitelist = {
-  id: string;
+export type WalletMember = {
+  address: string;
 };
 
-export type MemberBalance = {
-  address: string;
+export type BalanceMember = WalletMember & {
   balance: number;
 };
 
 export type DaoMembers = {
-  members: DaoWhitelist[] | MemberBalance[];
-  filteredMembers: DaoWhitelist[] | MemberBalance[];
+  members: WalletMember[] | BalanceMember[];
+  filteredMembers: WalletMember[] | BalanceMember[];
 };
 
 // this type guard will need to evolve when there are more types
-export function isWhitelistMember(
-  member: MemberBalance | DaoWhitelist
-): member is DaoWhitelist {
-  return 'id' in member;
+export function isWalletListMember(
+  member: BalanceMember | WalletMember
+): member is WalletMember {
+  return !('address' in member);
+}
+
+export function isBalanceMember(
+  member: BalanceMember | WalletMember
+): member is BalanceMember {
+  return 'balance' in member;
 }
 
 /**
@@ -29,8 +34,9 @@ export function isWhitelistMember(
  * for a search term to be passed in to filter the members list. NOTE: the
  * totalMembers included in the response is the total number of members in the
  * DAO, and not the number of members returned when filtering by search term.
+ *
  * @param pluginAddress plugin from which members will be retrieved
- * @param type plugin type
+ * @param pluginType plugin type
  * @param searchTerm Optional member search term  (e.g. '0x...')
  * @returns A list of DAO members and the total number of members in the DAO
  */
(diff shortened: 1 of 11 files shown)


packages/web-app/src/hooks/useDaoMembers.tsx
```diff
@@ -1,115 +1,130 @@
...
-  } = useQuery(DAO_MEMBERS, {
-    client: client[network],
-    fetchPolicy: 'no-cache',
-    variables: {
-      id,
-      ...(searchTerm ? {filter: {id: searchTerm}} : {}),
-    },
-  });
+  const client = usePluginClient(pluginAddress, pluginType);
+  console.log('stop annoying me', searchTerm);
```
RakeshUP: 😂

RakeshUP (approved):

packages/web-app/src/hooks/useDaoMembers.tsx
```diff
@@ -1,115 +1,129 @@
...
-// members list as well for token based DAOs.
 export type DaoWhitelist = {
   id: string;
 };
 
-export type DaoTokenBased = {
+export type MemberBalance = {
   address: string;
   balance: number;
 };
```
Fabricevladimir: A rename might be a good idea for both, especially `DaoWhitelist`. Would also be nice to have either `id` or `address` and not both. The type guard then would have to graduate to checking if `balance` is in there instead of `id`

packages/web-app/src/hooks/useDaoMembers.tsx
```diff
@@ -1,115 +1,129 @@
...
  * Hook to fetch DAO members. Fetches token if DAO is token based, and allows
- * for a search term to be passed in to filter the members list.
- * NOTE: the totalMembers included in the response is the total number of members in the DAO,
- * and not the number of members returned when filtering by search term.
- * @param id DAO id
+ * for a search term to be passed in to filter the members list. NOTE: the
+ * totalMembers included in the response is the total number of members in the
+ * DAO, and not the number of members returned when filtering by search term.
+ * @param pluginAddress plugin from which members will be retrieved
+ * @param type plugin type
```
Fabricevladimir: Update to match param name

packages/web-app/src/hooks/useDaoMembers.tsx
```diff
@@ -1,115 +1,129 @@
...
  * @param searchTerm Optional member search term  (e.g. '0x...')
  * @returns A list of DAO members and the total number of members in the DAO
  */
 export const useDaoMembers = (
```
Fabricevladimir, line 43: I assume the filtering is being done temporarily until it's available on SDK?

packages/web-app/src/hooks/useDaoMembers.tsx
```diff
@@ -1,115 +1,129 @@
...
+          pluginType === 'erc20voting.dao.eth'
+            ? // TODO as soon as the SDK exposes Token information, fetch balances
+              // from contract.
+              rawMembers.map(m => {
+                return {
+                  address: m,
+                  balance: Math.floor(Math.random() * 500 + 1),
+                } as MemberBalance;
+              })
+            : rawMembers.map(m => {
```
Fabricevladimir, line 84: This is one of those times where it sort of makes sense for SDK to return array of strings but for us it becomes an awkward mapping.

packages/web-app/src/hooks/useDaoMembers.tsx
```diff
@@ -1,115 +1,129 @@
...
+    if (!searchTerm) {
+      setFilteredData([]);
+    } else {
+      const filtered =
+        pluginType === 'erc20voting.dao.eth'
+          ? (data as MemberBalance[]).filter(d =>
+              d.address.includes(searchTerm)
+            )
+          : (data as DaoWhitelist[]).filter(d => d.id.includes(searchTerm));
+      setFilteredData(filtered);
```
Fabricevladimir, line 115: What's the reasoning behind returning both sets of data and not just the filtered when a search term is provided?

packages/web-app/src/hooks/usePluginClient.tsx
```diff
@@ -4,46 +4,57 @@ import {
...
+ * @method createWhitelist By passing instance plugin address will create an
+ * WhitelistClient
  * @returns The corresponding Client
  */
-
 export const usePluginClient = (
-  type: PluginTypes,
-  pluginAddress: Address
+  pluginAddress: Address,
+  pluginType?: PluginTypes
```
Fabricevladimir, line 26: Why make `pluginType` optional when that leads directly to to an error state in the default case? Is that so we don't have to do the jig where we pass in the `pluginType || ''` when we're waiting for the dao details to load?

Fabricevladimir (commented):
Thanks, Vali ; nice work. 😃 Left a couple of comments. Also, don't forget to update SDK version

packages/web-app/src/hooks/useDaoMembers.tsx
WissenIstNacht: I thought this was named that way for a reason and didn't touch it. Glad to rename it though 👍🏼

packages/web-app/src/hooks/useDaoMembers.tsx
WissenIstNacht: You don't need to do funny stuff to keep track of the different amounts of members.

packages/web-app/src/hooks/usePluginClient.tsx
WissenIstNacht: > when that leads directly to to an error state in the default case?

Didn't I add a guard against this?

> Is that so we don't have to do the jig where we pass in the `pluginType || ''` when [...]

Well, yeah. Particularly because `pluginType || ''` wouldn't work. `PluginType` wouldn't accept an empty string.

packages/web-app/src/hooks/useDaoMembers.tsx
WissenIstNacht: Yes, this is what the first bullet point in the description was trying to say 🙈

packages/web-app/src/hooks/useDaoMembers.tsx
Fabricevladimir: My assumption was the SDK would give you the total number of members even as it returns the filtered search result. Is that not the case? That way total number of members would always be there and the number of filtered would be `data.length`

WissenIstNacht: I have absolutely no idea 😂 

If it matters a lot to you I can try to figure it out and/or revert back to the way it was done. It's just that I was fighting with the code because after the changes it didn't work they way it was done before. 😅

Fabricevladimir (approved):
Gtg

PR #489, opened 2022-09-13
Feature: APP-966 - Proposal Details - Voting Terminal

Fabricevladimir:
## Description
- Adds proposal settings and results to `VotingTerminal` on Proposal Details page
- Adds voting logic (voting eligibility is **mocked**)
- Sets `voting power` and `token amount` columns to optional in `VotersTable`

Task: [APP-966]()

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/containers/votingTerminal/index.tsx
+++ b/packages/web-app/src/containers/votingTerminal/index.tsx
@@ -1,26 +1,30 @@
-import React, {useMemo, useState} from 'react';
-import styled from 'styled-components';
-
-// TODO: Change to use proper imports
-import {IconClock, IconInfo, StateEmpty} from '@aragon/ui-components';
+import {ProposalStatus} from '@aragon/sdk-client/dist/internal/interfaces/common';
 import {
-  AlertInline,
   AlertCard,
-} from '@aragon/ui-components/src/components/alerts';
-import {
+  AlertInline,
   ButtonGroup,
   ButtonText,
+  CheckboxListItem,
+  IconClock,
+  IconInfo,
+  IconRadioCancel,
+  LinearProgress,
   Option,
-} from '@aragon/ui-components/src/components/button';
-import {CheckboxListItem} from '@aragon/ui-components/src/components/checkbox';
-import {SearchInput} from '@aragon/ui-components/src/components/input';
-import {LinearProgress} from '@aragon/ui-components/src/components/progress';
-import {
+  SearchInput,
+  StateEmpty,
   VotersTable,
   VoterType,
-} from '@aragon/ui-components/src/components/table';
-import {useTranslation} from 'react-i18next';
+} from '@aragon/ui-components';
 import {shortenAddress} from '@aragon/ui-components/src/utils/addresses';
+import React, {useMemo, useState} from 'react';
+import {useTranslation} from 'react-i18next';
+import styled from 'styled-components';
+
+export type ProposalVoteResults = {
+  yes: {value: string | number; percentage: number};
+  no: {value: string | number; percentage: number};
+  abstain: {value: string | number; percentage: number};
+};
 
 // TODO: clean up props: some shouldn't be optional
 export type VotingTerminalProps = {
@@ -32,18 +36,14 @@ export type VotingTerminalProps = {
   participation?: string;
   approval?: string;
   voters?: Array<VoterType>;
-  status?: string;
-  statusLabel?: string;
-  strategy: string;
+  status?: ProposalStatus;
+  statusLabel: string;
+  strategy?: string;
   token?: {
     symbol: string;
     name: string;
   };
-  results?: {
-    yes: {value: string; percentage: string};
-    no: {value: string; percentage: string};
-    abstain: {value: string; percentage: string};
-  };
+  results?: ProposalVoteResults;
   votingInProcess?: boolean;
   onVoteClicked?: React.MouseEventHandler<HTMLButtonElement>;
   onCancelClicked?: React.MouseEventHandler<HTMLButtonElement>;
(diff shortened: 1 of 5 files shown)


RakeshUP (approved):

packages/web-app/src/pages/proposal.tsx
```diff
@@ -82,6 +108,140 @@ const Proposal: React.FC = () => {
...
+          ? t('votingTerminal.status.voteSubmitted')
+          : t('votingTerminal.voteNow');
+        break;
+    }
+    return [voteStatus, voteButtonLabel];
+  }, [proposal?.endDate, proposal?.startDate, proposal?.status, t, voted]);
+
+  // vote button state and handler
+  const {voteNowDisabled, onClick} = useMemo(() => {
+    let voteNowDisabled = true;
```
WissenIstNacht: No need for function-wide temps. You could simply return from the various if/else blocks.

packages/web-app/src/pages/proposal.tsx
```diff
@@ -95,7 +255,7 @@ const Proposal: React.FC = () => {
         {!isDesktop && (
           <Breadcrumb
             onClick={(path: string) =>
-              navigate(generatePath(path, {network, daoId}))
+              navigate(generatePath(path, {network, daoId: 'daoId'}))
```
WissenIstNacht: Why `daoId`?

packages/web-app/src/utils/proposals.ts
```diff
@@ -0,0 +1,243 @@
...
+  totalWeight: bigint
+): {voters: Array<VoterType>; summary: string} {
+  let weights = BigNumber.from(0);
+  let votingPower;
+  let tokenAmount;
+
+  // map to voters structure
+  const voters = votes.map(vote => {
+    // increment number of voted tokens
+    weights = weights.add(BigNumber.from(vote.weight));
```
WissenIstNacht, line 87: Is this not the same as total weights? Also, it's not used...?

packages/web-app/src/utils/proposals.ts
```diff
@@ -0,0 +1,243 @@
...
+      option: MappedVotes[vote.vote],
+      votingPower,
+      tokenAmount,
+    };
+  });
+
+  // calculate participation summary
+  const formattedTotalWeight = abbreviateTokenAmount(
+    parseFloat(
+      Number(formatUnits(totalWeight, token.decimals)).toFixed(2)
```
WissenIstNacht, line 115: Technically `abbreviateTokenAmount` also does rounding.

Either way, the number cast and then parse to string is not necessary, as `toFixed` already returns a string :)

// This happens in a couple of other places too, btw

packages/web-app/src/utils/proposals.ts
```diff
@@ -0,0 +1,243 @@
...
+  );
+
+  const formattedParticipation = abbreviateTokenAmount(
+    parseFloat(
+      Number(formatUnits(weights, token.decimals)).toFixed(2)
+    ).toString()
+  );
+
+  const participationPercentage = parseFloat(
+    Big(Number(weights)).mul(100).div(Number(totalWeight)).toFixed(2)
```
WissenIstNacht, line 126: I thought `div` would expect a `Big`? I.e., `Big(Number(totalWeight))`? (I might very well be wrong, though)

packages/web-app/src/containers/votingTerminal/index.tsx
```diff
@@ -82,15 +82,15 @@ export const VotingTerminal: React.FC<VotingTerminalProps> = ({
...
-  const statusIcon = useMemo(
-    () =>
-      status === 'pending' || status === 'active' ? (
-        <IconClock className="text-info-500" />
-      ) : (
-        <IconInfo className="text-info-500" />
-      ),
-    [status]
-  );
+  const statusIcon = useMemo(() => {
```
WissenIstNacht: Typically, functions that return components are themselves components :) And if you're really worried about performance, you can always wrap that component with memo.

WissenIstNacht (commented):
Cheers man, lots of good things in here.

packages/web-app/src/pages/proposal.tsx
Fabricevladimir: Fair point 👍🏾

packages/web-app/src/pages/proposal.tsx
Fabricevladimir: Oof, residue from disabling subgraph check. Good catch

packages/web-app/src/utils/proposals.ts
Fabricevladimir: It is not. This is perhaps a naming thing. `totalWeight` is how many tokens total can vote based on when the proposal was created, whereas `weights` is the total of all tokens _actually voted_.

Also, it is being used for `formattedParticipation` and `participationPercentage` on lines 121 & 126 respectively

packages/web-app/src/utils/proposals.ts
Fabricevladimir: Ah, yes there is a reason for this. I might extract it into some sort of function so it's centralized. We want 2 digits after the decimal point, but only when there are significant digits and not zero ie (1.45 -> good, 1 -> good, 1.00 -> bad).

The workaround itself: 
1. `toFixed` returns a string with two digits after the decimal including the zeros ('5.00')
2. `parseFloat` drops the digits after the decimal if they are only zeros leaving us with 5

It's not an elegant solution honestly and perhaps can be improved on (even if only by extracting it into some sort of function), but for now it works hehehe

packages/web-app/src/utils/proposals.ts
Fabricevladimir: `div` actually expects a `BigSource` which itself is `number | string | Big`.

packages/web-app/src/utils/proposals.ts
WissenIstNacht: Oh, I see. Come to think of it, this could actually be added to the abbreviation function itself.

WissenIstNacht (approved):
👌🏼

PR #490, opened 2022-09-14
Feature: App-965 Proposal Actions - Execution Widget

sepehr2github:
## Description

- Execution widget added to the proposal details page

Task: [APP-965](https://aragonassociation.atlassian.net/browse/APP-965)

## Type of change

## Checklist:


packages/web-app/src/hooks/useDaoProposal.tsx
```diff
@@ -27,15 +32,37 @@ export const useDaoProposal = (
...
   const client = usePluginClient(type, pluginAddress);
 
   useEffect(() => {
     async function getDaoProposal() {
       try {
         setIsLoading(true);
-
         const proposal = await client?.methods.getProposal(proposalId);
+        if (proposal && encodedData) proposal.actions[0] = encodedData;
         if (proposal) setData(proposal);
```
Fabricevladimir, line 63: I'm guessing you're not merging these two because adding the encoded data is temporary?

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
```diff
@@ -0,0 +1,54 @@
+import {ApolloClient} from '@apollo/client';
+import {Client} from '@aragon/sdk-client';
+import {constants} from 'ethers';
+import {fetchTokenData} from 'services/prices';
+import {SupportedNetworks} from 'utils/constants';
+
+import {ActionWithdraw} from 'utils/types';
+
+/**
+ * Get dao metadata
```
Fabricevladimir: Please update the comment. Also, could you move and rename the file since this isn't a hook but a utility method?

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
```diff
@@ -0,0 +1,54 @@
...
+  client: Client | undefined,
+  apolloClient: ApolloClient<object>,
+  network: SupportedNetworks
+): Promise<ActionWithdraw | undefined> {
+  if (!client || !data) {
+    console.error('SDK client is not initialized correctly');
+    return;
+  }
+
+  const decoded = await client?.decoding.withdrawAction(data);
```
Fabricevladimir: Optional chaining shouldn't be necessary for `client.decoding...` since on line 24 you're returning if there is no client.  Same thing for line 37. 

Also, `withdrawAction` isn't an asynchronous method so you don't need the await.

packages/web-app/src/components/executionWidget/index.tsx
```diff
@@ -21,7 +21,7 @@ export type ExecutionStatus =
 
 type ExecutionWidgetProps = {
   txhash?: string;
-  actions?: Array<Action>;
+  actions?: Array<Action | undefined>;
```
Fabricevladimir, line 24: What's the reasoning in allowing an array of undefined especially considering actions are optional already?

packages/web-app/src/components/executionWidget/index.tsx
```diff
@@ -57,9 +57,12 @@ export const ExecutionWidget: React.FC<ExecutionWidgetProps> = ({
       ) : (
         <>
           <Content>
-            {actions.map((action, index: number) => (
-              <ActionsFilter {...{action}} key={index} type={action.name} />
-            ))}
+            {actions?.map((action, index: number) => {
+              if (action)
```
Fabricevladimir, line 61: See above comment. If there are no actions, this shouldn't run at all thereby eliminating the if(action) check.

packages/web-app/src/hooks/useDaoProposal.tsx
```diff
@@ -46,7 +73,14 @@ export const useDaoProposal = (
     }
 
     getDaoProposal();
-  }, [client?.methods, proposalId]);
+  }, [
+    encodedData,
+    apolloClient,
+    client?.methods,
+    globalClient,
+    network,
```
Fabricevladimir: Why are `apolloClient`, `globalClient` and `network` dependencies for this?

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
```diff
@@ -0,0 +1,54 @@
...
+ * Note: Please rename to useDaoMetadata once the other hook as been deprecated
+ * @param data Uint8Array action data
+ * @param client SDK client, Fetched using useClient
+ * @param apolloClient Apollo client, Fetched using useApolloClient
+ * @param network network of the dao
+ * @returns Return Decoded Withdraw action
+ */
+export async function decodeWithdrawToAction(
+  data: Uint8Array | undefined,
+  client: Client | undefined,
```
Fabricevladimir: I'm not sure it makes sense to have the data be undefined if we're going to complain about it immediately afterwards. I understand client being undefined is so we don't have to do something like below correct?
```
decodeWithdrawToAction(data, client!)
// or
if(client) decodeWithdrawToAction(data, client)
```

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  const decoded = await client?.decoding.withdrawAction(data);
+
+  if (!decoded) {
+    console.error('Unable to decode withdraw action');
+    return;
+  }
+
+  const response = await fetchTokenData(
```
Fabricevladimir: Considering there might not be many actions this is fine for now, but we could get a performance boost if we batch fetchTokenData calls into a single `Promise.all` if there are multiple withdraws. It would also be good separation of concerns as well.

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
```diff
@@ -0,0 +1,54 @@
...
+    apolloClient,
+    network
+  );
+
+  return {
+    amount: Number(decoded.amount),
+    name: 'withdraw_assets',
+    to: decoded.recipientAddress,
+    tokenAddress: response?.address || (decoded?.tokenAddress as string),
+    tokenBalance: 0, // unnecessary
```
Fabricevladimir: We should cleanup the withdraw action at some point.

packages/web-app/src/hooks/useDaoProposal.tsx
```diff
@@ -27,15 +32,37 @@ export const useDaoProposal = (
...
   const [isLoading, setIsLoading] = useState(false);
+  const [encodedData, setEncodedData] = useState<DaoAction | undefined>();
+  const {client: globalClient} = useClient();
+  const {network} = useNetwork();
+
+  const apolloClient = useApolloClient();
+  const daoAddress = '0x1234567890123456789012345678901234567890';
+
+  useEffect(() => {
+    const getClient = async () => {
```
Fabricevladimir: Please update method name.

packages/web-app/src/pages/proposal.tsx
```diff
@@ -39,6 +44,10 @@ const Proposal: React.FC = () => {
   const navigate = useNavigate();
   const {isDesktop} = useScreen();
   const {breadcrumbs, tag} = useMappedBreadcrumbs();
+  const apolloClient = useApolloClient();
+  const [decodedActions, setDecodedActions] =
+    useState<(ActionWithdraw | undefined)[]>();
```
Fabricevladimir, line 66: Why make this an array of undefined as opposed to a list of ActionWithdraws?

packages/web-app/src/pages/proposal.tsx
```diff
@@ -76,6 +85,41 @@ const Proposal: React.FC = () => {
     }
   }, [editor, proposal]);
 
+  useEffect(() => {
+    async function getProposalAction() {
+      const actionPromises = proposal?.actions.map(action => {
+        return decodeWithdrawToAction(
```
Fabricevladimir: Shouldn't there be a check to see whether the action is `withdraw_assets`?

packages/web-app/src/pages/proposal.tsx
```diff
@@ -76,6 +85,41 @@ const Proposal: React.FC = () => {
...
+          action.data,
+          client,
+          apolloClient,
+          network
+        );
+      });
+
+      Promise.all(actionPromises || []).then(value => {
```
Fabricevladimir: This is a bit weird. I would suggest perhaps checking if there are actions in the first place, up top before you get the `actionPromises`.

packages/web-app/src/pages/proposal.tsx
```diff
@@ -152,7 +196,7 @@ const Proposal: React.FC = () => {
           {/* <VotingTerminal /> */}
 
           {/* TODO: Fill out voting terminal props*/}
```
Fabricevladimir: This one is on me hehehe. Can you remove that comment please

Fabricevladimir (changes requested):
Thanks Sepehr. Left a few comments. Unfortunately, the token & information doesn't show up for me when I open a proposal. 😢

packages/web-app/src/hooks/useDaoProposal.tsx
sepehr2github: Yup, We need to remove it in near future

packages/web-app/src/components/executionWidget/index.tsx
sepehr2github: because the promise all promise function has value:(ActionWithdraw | undefined)[] type which means arrays of undefined

packages/web-app/src/hooks/useDaoProposal.tsx
sepehr2github: Forgot to remove, I had no warning on VS which is wired!

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
sepehr2github: Yes, For sure will add necessary todos

packages/web-app/src/components/executionWidget/index.tsx
sepehr2github: I know but if you remove the if you'll see that Vs start complaining :/

packages/web-app/src/pages/proposal.tsx
sepehr2github: As I told you before, It's because of the value type on promise.all

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
sepehr2github: Please remember that this function will use generally, and if we use this later somewhere and for any reason, the data source couldn't work, the application shouldn't crash!

packages/web-app/src/hooks/useDecodeWithdrawToAction.tsx
sepehr2github: Let's wait till we add more actions so We will refactor this, because we need to filter the withdraw actions first

packages/web-app/src/pages/proposal.tsx
```diff
@@ -76,6 +86,55 @@ const Proposal: React.FC = () => {
...
+      if (proposal) {
+        const actionPromises = proposal.actions.map(action =>
+          findDecoder(action)
+        );
+        Promise.all(actionPromises).then(value => {
+          setDecodedActions(value);
+        });
+      }
+    }
+    getProposalAction();
```
WissenIstNacht: If you're using `then` callbacks you don't need to wrap everything in an async function :)
```suggestion
      if (proposal) {
        const actionPromises = proposal.actions.map(action =>
          findDecoder(action)
        );
        Promise.all(actionPromises).then(value => {
          setDecodedActions(value);
        });
      }
```

packages/web-app/src/pages/proposal.tsx
```diff
@@ -76,6 +86,55 @@ const Proposal: React.FC = () => {
     }
   }, [editor, proposal]);
 
+  const findDecoder = useCallback(
```
WissenIstNacht: Memoizing things and then using it in effects can lead to tricky bugs. Since you're only using `findDecoder` in the `useEffect`right below it, could you simply move the function into the `useEffect`?

packages/web-app/src/pages/proposal.tsx
```diff
@@ -76,6 +86,55 @@ const Proposal: React.FC = () => {
...
+            action.data,
+            client,
+            apolloClient,
+            network
+          );
+        default:
+          return {} as ActionWithdraw;
+      }
+    },
+    [apolloClient, client, network]
```
WissenIstNacht: If you're return a promise in the other case it's gonna make things easier if you return a Promise here too. You can do this by wrapping the returned object with the `reolve` function

```
          return Promise.resolve({} as ActionWithdraw);
```

// EDIT sorry, I think I selected the wrong line. The line of code above should oviously replace the one on line 101, not 104 :)

packages/web-app/src/pages/proposal.tsx
```diff
@@ -39,6 +45,10 @@ const Proposal: React.FC = () => {
   const navigate = useNavigate();
   const {isDesktop} = useScreen();
   const {breadcrumbs, tag} = useMappedBreadcrumbs();
+  const apolloClient = useApolloClient();
+  const [decodedActions, setDecodedActions] =
+    useState<(ActionWithdraw | void)[]>();
```
WissenIstNacht: `void` is not the right type here, as it represents the return type of functions that don't return anything. Did you mean `undefined` or `null`? (Note that while you _can_ assign `undefined` to `void` they are _NOT THE SAME_. Therefore, `void` is not a replacement for `undefined`)

packages/web-app/src/components/executionWidget/index.tsx
```diff
@@ -57,9 +57,12 @@ export const ExecutionWidget: React.FC<ExecutionWidgetProps> = ({
       ) : (
         <>
           <Content>
-            {actions.map((action, index: number) => (
-              <ActionsFilter {...{action}} key={index} type={action.name} />
-            ))}
+            {actions?.map((action, index: number) => {
```
WissenIstNacht: For me TS says that `actions` is always an array. I.e., never `null` or `undefined`. Which makes sense sinse we're assigning `[]` by default on line 30. So the question mark is not necessary :)

```suggestion
            {actions.map((action, index: number) => {
```

packages/web-app/src/components/executionWidget/index.tsx
```diff
@@ -21,7 +21,7 @@ export type ExecutionStatus =
 
 type ExecutionWidgetProps = {
   txhash?: string;
-  actions?: Array<Action>;
+  actions?: Array<Action | void>;
```
WissenIstNacht: Same comment as in the other file :)

packages/web-app/src/components/executionWidget/index.tsx
Fabricevladimir: I can kind of see why this is returning the promise is returning undefined. I would suggest filtering the array before passing it in. But the greater question in here then becomes what exactly do we do with those actions that couldn't be decoded for whatever reason? That's something we need to think about.

packages/web-app/src/hooks/useDaoProposal.tsx
Fabricevladimir: I think linter will catch when it's missing dependencies only.

WissenIstNacht (changes requested):
Hey thanks Sepehr :)

Left some suggestion.

packages/web-app/src/pages/proposal.tsx
```diff
@@ -85,23 +86,37 @@ const Proposal: React.FC = () => {
...
-        if (value) setDecodedActions(value);
-      });
+        Promise.all(actionPromises).then(value => {
+          setDecodedActions(value);
+        });
+      }
     }
     getProposalAction();
-  }, [apolloClient, client, network, proposal?.actions]);
+  }, [apolloClient, client, findDecoder, network, proposal, proposal?.actions]);
```
Fabricevladimir: Can you take a look at the dependency array?  I think `findDecoder` and `proposal` are perhaps the only ones necessary. You could go one step further and instead of checking for proposal, you could check for `proposal?.actions` in the if statement. But that might be pure overkill because I can't think of any time where the proposal would change and the actions not change 😄

Fabricevladimir (commented):
I can see the action! Left one small comment

WissenIstNacht (approved):
Good to go from my side, thanks!

Fabricevladimir (approved):

PR #493, opened 2022-09-16
Feat: Store Deposits locally to display loading state

RakeshUP:
## Description

Task: [APP-565](https://aragonassociation.atlassian.net/browse/APP-565)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/hooks/useDaoTransfers.tsx
+++ b/packages/web-app/src/hooks/useDaoTransfers.tsx
@@ -29,7 +29,7 @@ export const useDaoTransfers = (
         const transfers = await client?.methods.getTransfers(daoAddressOrEns);
         if (transfers) {
           if (pendingDepositsTxs.length > 0) {
-            for (let i = 0; i < pendingDepositsTxs.length; i++) {
+            for (let i = 0; i < pendingDepositsTxs.length; ) {
               const tx = pendingDepositsTxs[i];
 
               for (let j = 0; j < transfers.deposits.length; j++) {
@@ -37,6 +37,9 @@ export const useDaoTransfers = (
                 if (deposit.transactionId === tx.transactionId) {
                   pendingDepositsTxs.splice(i, 1);
                   break;
+                }
+                if (j === transfers.deposits.length - 1) {
+                  i++;
                 }
               }
             }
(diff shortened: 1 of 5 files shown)


packages/web-app/src/hooks/useDaoTransfers.tsx
```diff
@@ -16,14 +19,39 @@ export const useDaoTransfers = (
...
   useEffect(() => {
     async function getTransfers() {
       try {
         setIsLoading(true);
 
         const transfers = await client?.methods.getTransfers(daoAddressOrEns);
-        if (transfers) setData(transfers);
+        if (transfers) {
+          if (pendingDepositsTxs.length > 0) {
+            for (let i = 0; i < pendingDepositsTxs.length; i++) {
```
WissenIstNacht: Ahm, couldn't it be that you're jumping elements here, since you're modifying `pendingDepositsTxs` inside the loop? Imaginge you splice away the first item in the first iteration (i.e., when `i` is 0). Due to the splicing the second element will now have index 0. But in the next iteration, `i` will be incremented to 1. Meaning that this second item will effectively be jumped over.

WissenIstNacht (commented):
Hey Rakesh, thanks for the work.

I think I found an issue in the transfer hooks. But I might be wrong. Let me know what you think.

In any case, the pending indicator shows up just nice. However, it doesn't seem to go away. I've created a deposit and waited for a solid half hour. Reloading the page still shows the deposit as pending.

Fabricevladimir (approved):
Thanks! Looks good apart from what Vali mentioned about he slicing

packages/web-app/src/hooks/useDaoTransfers.tsx
RakeshUP: Ohh yes, I had totally missed this one. When I was writing this for the first time I knew the loop handling should be checked as I was using splice, but forgot to do so later. Thanks for finding this out!

Regarding the pending indicator, when we start getting actual data from the SDK, this splice logic will remove the pending tx's that are indexed by thegraph and returned by the SDK. When the user refreshes the page, and if the tx is returned by the SDK, the pending status would have changed. For now, the pending tx won't go away and stays there till localStorage is cleared..

WissenIstNacht (approved):
Cheers, I think this solves the issue.

PR #507, opened 2022-09-28
Feature: Token gating modals for proposal creation

sepehr2github:
## Description

- token gating modal added
- wallet gating modal added

Task: [APP-1039](https://aragonassociation.atlassian.net/browse/APP-1039) -  [APP-988](https://aragonassociation.atlassian.net/browse/APP-988)

## Type of change

## Checklist:


RakeshUP (approved):

packages/web-app/src/app.tsx
```diff
@@ -145,6 +151,35 @@ function App() {
   );
 }
 
+const ProtectedRoute: React.FC = () => {
```
Fabricevladimir: Can you move this to its own file?

packages/web-app/src/app.tsx
```diff
@@ -145,6 +151,35 @@ function App() {
...
+  } = useDaoMembers(
+    daoDetails?.plugins[0].instanceAddress || '',
+    daoDetails?.plugins[0].id as PluginTypes,
+    address as string
+  );
+
+  if (paramIsLoading || detailsAreLoading || MembershipIsLoading)
+    return <Loading />;
+
+  if (filteredMembers.length === 0 && daoDetails)
```
Fabricevladimir: I think it's worth asking for login, but right now we're checking for membership regardless. For example, if user  wallet isn't connected, right now, it assumes you're not a member since the query will return no filtered members. This is deceptive because the user doesn't really know why he's getting booted out of the flow even though he has the proper tokens. What we should do instead, I suggest, is direct user to login first and then check for membership.

packages/web-app/src/containers/gatingMenu/tokenGating.tsx
```diff
@@ -0,0 +1,56 @@
...
+  WarningContainer,
+  WarningTitle,
+} from 'containers/networkErrorMenu';
+import {useDaoToken} from 'hooks/useDaoToken';
+import {useDaoParam} from 'hooks/useDaoParam';
+
+export const TokenGating = () => {
+  const {close, isRequiredTokenOpen} = useGlobalModalContext();
+  const {t} = useTranslation();
+  const {data: dao} = useDaoParam();
```
Fabricevladimir: We need to find a way to stop hitting `useDaoParam` on so many components. 🙈

packages/web-app/src/containers/gatingMenu/tokenGating.tsx
```diff
@@ -0,0 +1,56 @@
...
+      isOpen={isRequiredTokenOpen}
+      onClose={() => close('requiredToken')}
+    >
+      <ModalBody>
+        <StyledImage src={WalletIcon} />
+        <WarningContainer>
+          <WarningTitle>{t('alert.gatingUsers.tokenTitle')}</WarningTitle>
+          <WarningDescription>
+            {t('alert.gatingUsers.tokenDescription', {
+              tokenName: !daoTokenLoading ? daoToken?.name : '',
```
Fabricevladimir: Unfortunately, this doesn't show a token name for me 😞

packages/web-app/src/pages/newProposal.tsx
```diff
@@ -36,6 +37,8 @@ const NewProposal: React.FC = () => {
           setShowTxModal={setShowTxModal}
         >
           <ProposalStepper enableTxModal={enableTxModal} />
+          <TokenGating />
+          <WalletGating />
```
Fabricevladimir: I think these two should be in the `ProtectedRoute` component instead so that everyone can have access to the singular instance, especially since they are needed in multiple flows.

packages/web-app/src/app.tsx
```diff
@@ -105,10 +109,12 @@ function App() {
...
               <Route path="governance" element={<GovernancePage />} />
-              <Route
-                path="governance/new-proposal"
-                element={<NewProposalPage />}
-              />
+              <Route element={<ProtectedRoute />}>
+                <Route
+                  path="governance/new-proposal"
+                  element={<NewProposalPage />}
+                />
```
Fabricevladimir, line 112: Can you move the other protected proposal routes as well: `EditSettings`,  `ProposeNewSettings`, `MintTokensProposal`, `ManageMembersProposal` and probably `NewWithdraw` as well? Either that or leave a comment indicating that since it will get a tad bit annoying if we have to be gated while developing 😆

packages/web-app/src/app.tsx
```diff
@@ -145,6 +151,35 @@ function App() {
...
+    return <Loading />;
+
+  if (filteredMembers.length === 0 && daoDetails)
+    open(
+      daoDetails?.plugins[0].id === 'erc20voting.dao.eth'
+        ? 'requiredToken'
+        : 'requiredWallet'
+    );
+
+  return <Outlet />;
```
Fabricevladimir: What are your thoughts on using the `Outlet` as opposed to returning a child component? Just curious 🙈

Fabricevladimir (commented):
Thanks Sepehr! Good work. Left a couple of comments.

packages/web-app/src/app.tsx
sepehr2github: For performance not sure but I feel using the react-router approach would be much better, Also I decided this way to keep consistency.

WissenIstNacht: Just FYI here: passing `children` simply doesn't work.When I was working on this file a couple of months ago, I remember having the exact same question. So I tried it out and it didn't work. Things might have changed in the meantime. But even if they did, I don't think it's worth rewriting the file. :p

packages/web-app/src/containers/gatingMenu/tokenGating.tsx
WissenIstNacht: Sorry to spam your PR Sepehr, but these are interesting comments :p

I mean you can't actually avoid this parameter, as it's necessary for all sorts of things. But there's something that's bugging me too, here. Namely that the hook actually queries in each component it's called from. I think the best way to solve this would be to create a context around all routes that are DAO-related. In the provider, perform the "dao exists check" and then simply expose the `daoParam` to consumers.

A similar structure might also be useful for things like `client`s and `daoDetails`, btw. Maybe they could even be merged into one?

sepehr2github: Agreed! @WissenIstNacht How about creating a bug task to tackle this

WissenIstNacht: @sepehr2github Why bug? Is something broken? 😅 

I don't think this is an important thing to tackle. The code works as is expected and performs well (so far). As long as this is true I wouldn't make this a priority. But maybe we can add this to the list of todos and tackle it as a simple improvement task at some point?

packages/web-app/src/containers/gatingMenu/tokenGating.tsx
```diff
@@ -0,0 +1,64 @@
...
+    daoDetails?.plugins[0].instanceAddress as string
+  );
+  const navigate = useNavigate();
+
+  console.log('view', daoToken);
+
+  return (
+    <ModalBottomSheetSwitcher
+      isOpen={isRequiredTokenOpen}
+      onClose={() => close('requiredToken')}
```
WissenIstNacht: If you click somewhere outside the modal it simply closes. And then you're on the creation flow and you can simply continue. In other words, you should make sure that you can not close the modal except by pressing the "Ok, understood" button.

(Sorry if this is no the right spot in the code for this comment :p)

packages/web-app/src/containers/gatingMenu/tokenGating.tsx
```diff
@@ -0,0 +1,64 @@
...
+            {t('alert.gatingUsers.tokenDescription', {
+              tokenName:
+                !daoTokenLoading && !detailsAreLoading ? daoToken?.name : '',
+            })}
+          </WarningDescription>
+        </WarningContainer>
+        <ButtonText
+          label={t('alert.gatingUsers.buttonLabel')}
+          onClick={() => {
+            navigate(-1);
```
WissenIstNacht: What happens if you visit the the page on a fresh tab? Should we send people to the DAO dashboard?

packages/web-app/src/containers/gatingMenu/tokenGating.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const {t} = useTranslation();
+  const {data: dao} = useDaoParam();
+  const {data: daoDetails, isLoading: detailsAreLoading} = useDaoDetails(dao!);
+
+  const {data: daoToken, isLoading: daoTokenLoading} = useDaoToken(
+    daoDetails?.plugins[0].instanceAddress as string
+  );
+  const navigate = useNavigate();
+
+  console.log('view', daoToken);
```
WissenIstNacht: :D

packages/web-app/src/components/protectedRoute/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+    daoDetails?.plugins[0].id as PluginTypes,
+    address as string
+  );
+
+  if (paramIsLoading || detailsAreLoading || MembershipIsLoading)
+    return <Loading />;
+
+  if (filteredMembers.length === 0 && daoDetails && isConnected) {
+    open(
+      daoDetails?.plugins[0].id === 'erc20voting.dao.eth'
```
WissenIstNacht: Optional chaining (i.e., the question mark after `daoDetails`) is not needed when you check that `daoDetails` exists in the if statement :)

packages/web-app/src/containers/gatingMenu/walletGating.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
```
WissenIstNacht: (same comments as for token gating)

WissenIstNacht (changes requested):
Hey Sepehr, nice job! Lots of cool things in here.

I'm jumping in on Fabrice's behalf since he's not here. I've found a few tricky things that need some more work. I've left some comments in the code where I could. But there's a few things I didn't know where to comment, so I'm writing them here:

- Visiting the url `.../governance/new-proposal` directly in a fresh app crashes the application for me.
- The gating doesn't check whether a user is logged-in. I.e., if I log out and click on proposal creation I'm not stopped.
- The withdrawal flow is not gated.

packages/web-app/src/containers/gatingMenu/tokenGating.tsx
sepehr2github: Thanks, For catching this great point, after a quick sync with Marco we decided to return the user to Governance page all the time

Fabricevladimir (approved):
Thank you Sepehr, approving 😺
As for the wrong network, if we are to check for it as well, I trust Vali will take care of the review for it, or you can re-request my review again!

packages/web-app/src/containers/gatingMenu/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+  );
+  const {t} = useTranslation();
+
+  return (
+    <WarningContainer>
+      <WarningTitle>{t('alert.gatingUsers.tokenTitle')}</WarningTitle>
+      <WarningDescription>
+        {t('alert.gatingUsers.tokenDescription', {
+          tokenName:
+            !daoTokenLoading && !detailsAreLoading ? daoToken?.name : '',
```
WissenIstNacht, line 36: Small detail here: I think if `daoToken` is `undefined` the text here will be a string with the value "undefined".

packages/web-app/src/hooks/useDaoParam.tsx
```diff
@@ -1,49 +1,30 @@
...
-  } = useQuery(DAO_BY_ADDRESS, {
-    variables: {id: dao ? dao : ''},
-    client: client[network],
-    fetchPolicy: 'cache-and-network',
-  });
+  const {error, isLoading} = useDaoDetails(dao as string);
   const navigate = useNavigate();
 
   useEffect(() => {
-    // undefined mainnet clients cause problem here. This bypasses the dao param
```
WissenIstNacht, line 35: Scary change. Was something pressing you to do it that you had to do it in this PR?

packages/web-app/src/components/protectedRoute/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }, [address, close, isConnected, isOnWrongNetwork, open, status]);
+
+  if (
+    filteredMembers.length === 0 &&
+    daoDetails &&
+    isConnected &&
+    !isOnWrongNetwork
+  ) {
+    open('gating');
```
WissenIstNacht, line 46: Technically this is a side-effect (the function changes something outside its scope) and should technically be in a useEffect :)

WissenIstNacht (approved):
Sorry this took so long, Sepehr. Thanks for all the changes. I left some minor comments. If you want to take a look at them, great. Otherwise the PR is good to go!

packages/web-app/src/hooks/useDaoParam.tsx
sepehr2github: Yes, I made this change because each time you directly open proposal pages It takes time to initiate the client but this condition will move the user to not found page by mistake

PR #510, opened 2022-10-04
App-1001  Encode Metadata and Governance Settings

WissenIstNacht:
## Description

This PR integrates adapts the code from the creation flow to the settings flow. It allows to encode both changes to a DAOs metadata and governance settings.

Task: [1001](https://aragonassociation.atlassian.net/browse/APP-1001)

## Type of change

### DIFF
--- a/packages/web-app/src/pages/proposeSettings.tsx
+++ b/packages/web-app/src/pages/proposeSettings.tsx
@@ -3,12 +3,15 @@ import {
   InstalledPluginListItem,
   ProposalCreationSteps,
 } from '@aragon/sdk-client';
+import {IMetadata} from '@aragon/sdk-client/dist/internal/interfaces/client';
+import {DaoAction} from '@aragon/sdk-client/dist/internal/interfaces/common';
+import {IPluginSettings} from '@aragon/sdk-client/dist/internal/interfaces/plugins';
 import {withTransaction} from '@elastic/apm-rum-react';
+import Big from 'big.js';
 import React, {useCallback, useEffect, useState} from 'react';
 import {useTranslation} from 'react-i18next';
 import {generatePath, useNavigate} from 'react-router-dom';
 
-import {DaoAction} from '@aragon/sdk-client/dist/internal/interfaces/common';
 import {FullScreenStepper, Step} from 'components/fullScreenStepper';
 import {Loading} from 'components/temporary';
 import CompareSettings from 'containers/compareSettings';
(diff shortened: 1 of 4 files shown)


packages/web-app/src/pages/proposeSettings.tsx
```diff
@@ -45,14 +74,271 @@ const ProposeSettings: React.FC = () => {
...
+
+      const durationInSeconds = getSecondsFromDHM(
+        durationDays,
+        durationHours,
+        durationMinutes
+      );
+      const settingsParams: IPluginSettings = {
+        minDuration: durationInSeconds,
+        minSupport: Big(support).div(100).toNumber(),
+        minTurnout: Big(minimumParticipation).div(100).toNumber(),
```
Fabricevladimir, line 165: `minimumParticipation` is returning `NaN` causing this to throw

packages/web-app/src/pages/proposeSettings.tsx
```diff
@@ -45,14 +74,271 @@ const ProposeSettings: React.FC = () => {
...
+      };
+      actions.push(
+        Promise.resolve(
+          pluginClient.encoding.updatePluginSettingsAction(dao, settingsParams)
+        )
+      );
+      return Promise.resolve([] as DaoAction[]);
```
Fabricevladimir: Forgot to return the encoded actions down here

Fabricevladimir (commented):
Thanks Vali! This flow has hidden issues 😅 Left a couple comments specifically regarding the encoding

RakeshUP (approved):

packages/web-app/src/pages/proposeSettings.tsx
WissenIstNacht: Yep, this subtask has a sibling subtask where I'm tackling this. 

I'm doing this over there because the code here should be ok as is. The idea is that at this point, the `minimumParticipation` should never be an invalid input. The flows are supposed to be built in such a way that we're checking user inputs at every step. And we only let them through if the inputs are correct. So I'm going to make sure this is the case in the other subtask. Does that make sense? Let me know if you think I'm missing something here.

WissenIstNacht:
@Fabricevladimir 

> This flow has hidden issues

Yes, I've noticed :joy: As I've pointed out in a response to your comment, there is a second subtask that tries to resolve all of them. I've singled this out because it seemed like a good way to partition work so the PR wouldn't be too huge :)

Fabricevladimir (approved):
Good to go!

PR #515, opened 2022-10-10
Feature: App-1045

sepehr2github:
## Description

- add token minting card to proposal details
- decoding the mini token

Notes:
1) The token minting summery on proposal creation not working well, will address this issue on another PR
2) I forgot the add Vele's requested changes to my last pr (https://github.com/aragon/apps/pull/507#discussion_r993215957 & https://github.com/aragon/apps/pull/507#discussion_r993196576) So I did the changes here 

Task: [APP-1034](https://aragonassociation.atlassian.net/browse/APP-1034)

## Type of change

## Checklist:


Fabricevladimir:
> Notes:
> 
> 1. The token minting summery on proposal creation not working well, will address this issue on another PR

@sepehr2github I believe this issue is happening because the DAO token _isn't_ actually available on Goerli. The token is only on Rinkeby

packages/web-app/src/hooks/useDaoProposal.tsx
```diff
@@ -88,7 +90,24 @@ export const useDaoProposal = (
...
       ]);
-    } else return Promise.all([encodedWithdrawAction]);
+    } else {
+      const encodedTokenMintingActions = members.map((member: string) =>
+        Promise.resolve(
+          (pluginClient as ClientErc20).encoding.mintTokenAction(
+            daoTokenAddress,
+            {
+              address: member,
+              amount: BigInt(10),
```
Fabricevladimir: This needs to be multiplied by 10^daoToken decimal

packages/web-app/src/pages/proposal.tsx
```diff
@@ -158,6 +173,13 @@ const Proposal: React.FC = () => {
           }
         });
 
+      const decodedMintToken = decodeMintTokensToAction(
```
Fabricevladimir: When decoding, the token amount we get back needs to be divided by 10^daotoken decimals as well.

packages/web-app/src/utils/library.ts
```diff
@@ -126,6 +128,41 @@ export async function decodeWithdrawToAction(
...
+  );
+
+  const newTokens = decoded.reduce((accumulator, value) => {
+    return accumulator + Number(value.amount);
+  }, 0);
+
+  return Promise.resolve({
+    name: 'mint_tokens',
+    inputs: {
+      mintTokensToWallets: decoded,
```
Fabricevladimir: Each one of the amounts in the wallet need to be formatted using the doatoken decimals

packages/web-app/src/utils/library.ts
```diff
@@ -126,6 +128,41 @@ export async function decodeWithdrawToAction(
...
+    },
+    summary: {
+      newTokens: BigInt(newTokens),
```
Fabricevladimir: Same thing here.

packages/web-app/src/utils/library.ts
```diff
@@ -126,6 +128,41 @@ export async function decodeWithdrawToAction(
...
+    console.error('SDK client is not initialized correctly');
+    return;
+  }
+
+  const decoded: IMintTokenParams[] = data.map(action =>
+    client.decoding.mintTokenAction(action)
+  );
+
+  const newTokens = decoded.reduce((accumulator, value) => {
+    return accumulator + Number(value.amount);
```
Fabricevladimir: To prevent precision lost, you might need to add those as Bignumbers or Bigs. Not sure how much of a concern that is though to be honest lol

packages/web-app/src/pages/proposal.tsx
```diff
@@ -158,6 +173,13 @@ const Proposal: React.FC = () => {
...
+        mintTokenActions.actions,
+        pluginClient as ClientErc20
+      );
+
+      actionPromises.splice(mintTokenActions.index, 0, decodedMintToken);
+
```
WissenIstNacht: I'm not sure I understand the changes you made to this `useEffect`. Let's see: If one of the proposal's actions is a minting action, you push that to a separate array and store the index. Any other action gets immediately mapped to the decoded version. And then at the end you splice in that separate array into the array of decoded actions. Right?

But why? Why not simply decode the mint actions immediately, the same way you decode any other action? Why take it out first and add it back later?

packages/web-app/src/components/executionWidget/actions/mintTokenCard.tsx
```diff
@@ -80,7 +111,8 @@ export const MintTokenCard: React.FC<{
             <Label>{t('labels.totalTokens')}</Label>
             {tokenSupply ? (
               <p>
-                {(tokenSupply + newTokens).toString()} {daoTokenSymbol}
+                {Big(tokenSupply).toNumber() + Number(newTokens)}{' '}
```
WissenIstNacht: tokenSupply is already of type `Big`

packages/web-app/src/utils/types.ts
```diff
@@ -249,14 +250,14 @@ export type ActionRemoveAddress = {
 export type ActionMintToken = {
   name: 'mint_tokens';
   inputs: {
-    mintTokensToWallets: Array<{address: string; amount: string}>;
+    mintTokensToWallets: IMintTokenParams[];
   };
   summary: {
-    newTokens: number;
-    tokenSupply: number;
+    newTokens: number | bigint;
```
Fabricevladimir: Why not pass it in only as a `number` and cast the inputs instead?

Fabricevladimir (changes requested):
Good things, good things. Leaving a few comments. 

Also, I thought I had submitted the review previously 🙈

WissenIstNacht (commented):
Thanks Sepehr. 

Fabrice left a lot of good comments already. Really curious about the comment in the `proposal.tsx` file.

packages/web-app/src/pages/proposal.tsx
sepehr2github: Hey, the decoder method on SDK will return one address per action for token minting, It's not working for us because we considered all the minting addresses as one action, not multiple actions, So In this way, I decoded all the actions by an iterator, then bound them into the one action

Fabricevladimir (approved):

packages/web-app/src/utils/library.ts
```diff
@@ -126,6 +133,63 @@ export async function decodeWithdrawToAction(
...
+
+    return Promise.resolve({
+      name: 'mint_tokens',
+      inputs: {
+        mintTokensToWallets: decoded,
+      },
+      summary: {
+        newTokens: Number(formatUnits(newTokens, decimals)),
+        tokenSupply: parseFloat(formatUnits(totalSupply, decimals)),
+        newHoldersCount: decoded.length,
```
WissenIstNacht, line 190: That's _technically_ not correct. The minting could go to addresses who already hold that token.

Since this is a bigger task and it's not gonna bother anyone for the next couple of months, please simply add a TODO so that we don't forget or think it's a bug in the future.

packages/web-app/src/pages/proposal.tsx
WissenIstNacht: Oh, I see! Thanks a lot for the explanation. Two things:

1. That explanation might be worth putting in a comment.
2. So you're bundling the mints to individual addresses that you get from the SDK, you decode them, you bundle them and then splice them back into the array with all the other actions, right? However, in splice, you only add the bundled action back in. Where do you actually delete the individual mint action you get from the SDK? In other words, shouldn't the delete count in splice be a number other than 0?

WissenIstNacht (commented):
Hey there, thanks for the changes.

I'm still a bit uncertain about the code in `proposal.tsx`. The other issue you can just add a TODO and resolve. :)

WissenIstNacht (approved):

PR #518, opened 2022-10-11
Feature: APP-1081 - Proposal Creation Nav: Main & Withdraw Flows

Fabricevladimir:
## Description
- Caches newly published proposal
- Redirects to `ProposalDetails` page once proposal is successfully published
- Uses cached proposal for `ProposalDetails` page if SDK data for proposal not yet available
- Uses cached proposals in DAO proposal lists (`Governance`, `Dashboard`) if SDK data for cached proposal not yet available
- Persists cached proposals if functional cookies are set
- Updates `useDaoToken` hook to use SDK (had to be done for caching)

NOTE: This lays the foundation for the other singular proposal flows, hence the size.
Also, currently directly accessing localstorage; would like to refactor and hide it behind the `useCache` hook

Task: [APP-1081](https://aragonassociation.atlassian.net/browse/APP-1081)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/hooks/useDaoActions.tsx
+++ b/packages/web-app/src/hooks/useDaoActions.tsx
@@ -1,11 +1,11 @@
 import {useTranslation} from 'react-i18next';
 
 import {ActionParameter, HookData} from 'utils/types';
-import {useDaoMetadata} from './useDaoMetadata';
+import {useDaoDetails} from './useDaoDetails';
 
 export function useDaoActions(dao: string): HookData<ActionParameter[]> {
-  const {data, error, isLoading} = useDaoMetadata(dao);
-  const whitelist = data?.packages[0].pkg.__typename === 'AllowlistPackage';
+  const {data: daoDetails, error, isLoading} = useDaoDetails(dao);
+  const whitelist = daoDetails?.plugins[0].id === 'addresslistvoting.dao.eth';
 
   const {t} = useTranslation();
 
--- a/packages/web-app/src/utils/library.ts
+++ b/packages/web-app/src/utils/library.ts
@@ -215,12 +215,8 @@ export const customJSONReplacer = (_: string, value: unknown) => {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const customJSONReviver = (_: string, value: any) => {
   // deserialize uint8array
-  try {
-    if ('flag' in value && value.flag === FLAG_TYPED_ARRAY) {
-      return new Uint8Array(value.data);
-    }
-  } catch (e) {
-    // not an object? do nothing
+  if (value.flag === FLAG_TYPED_ARRAY) {
+    return new Uint8Array(value.data);
   }
 
   if (typeof value === 'string') {
--- a/packages/web-app/src/utils/library.ts
+++ b/packages/web-app/src/utils/library.ts
@@ -5,7 +5,11 @@ import {ApolloClient} from '@apollo/client';
 import {Client, ClientAddressList} from '@aragon/sdk-client';
 
 import {fetchTokenData} from 'services/prices';
-import {SupportedNetworks} from 'utils/constants';
+import {
+  BIGINT_PATTERN,
+  ISO_DATE_PATTERN,
+  SupportedNetworks,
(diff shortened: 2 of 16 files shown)


RakeshUP (approved):

packages/web-app/src/utils/library.ts
```diff
@@ -178,3 +182,55 @@ export async function decodeRemoveMembersToAction(
...
 }
+
+const FLAG_TYPED_ARRAY = 'FLAG_TYPED_ARRAY';
+/**
+ *  Custom serializer that includes fix for BigInt type
+ * @param _ key; unused
+ * @param value value to serialize
+ * @returns serialized value
+ */
+export const customJSONReplacer = (_: string, value: unknown) => {
```
WissenIstNacht, line 193: Why have a function with 2 args if one of them is not used? :p

Couldn't you just remove it and only keep the `value` argument?

packages/web-app/src/utils/library.ts
```diff
@@ -178,3 +182,55 @@ export async function decodeRemoveMembersToAction(
...
+// disabling so forced assertion is not necessary in try catch
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const customJSONReviver = (_: string, value: any) => {
+  // deserialize uint8array
+  try {
+    if ('flag' in value && value.flag === FLAG_TYPED_ARRAY) {
+      return new Uint8Array(value.data);
+    }
+  } catch (e) {
+    // not an object? do nothing
```
WissenIstNacht: Does something throw an Error here? If so, please report it. Swallowed errors will eventually haunt you or someone else.

packages/web-app/src/utils/library.ts
```diff
@@ -178,3 +182,55 @@ export async function decodeRemoveMembersToAction(
...
+ * @param _ key: unused
+ * @param value value to deserialize
+ * @returns deserialized value
+ */
+// disabling so forced assertion is not necessary in try catch
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const customJSONReviver = (_: string, value: any) => {
+  // deserialize uint8array
+  try {
+    if ('flag' in value && value.flag === FLAG_TYPED_ARRAY) {
```
WissenIstNacht: I think you could leave out the first check. If flag's not in there, `value.flag` would simply evaluate to `undefined` and the comparison would return `false`.

packages/web-app/src/hooks/useDaoActions.tsx
```diff
@@ -1,11 +1,11 @@
...
 
 import {ActionParameter, HookData} from 'utils/types';
-import {useDaoMetadata} from './useDaoMetadata';
+import {useDaoDetails} from './useDaoDetails';
 
 export function useDaoActions(dao: string): HookData<ActionParameter[]> {
-  const {data, error, isLoading} = useDaoMetadata(dao);
-  const whitelist = data?.packages[0].pkg.__typename === 'WhitelistPackage';
+  const {data: daoDetails, error, isLoading} = useDaoDetails(dao);
+  const whitelist = daoDetails?.plugins[0].id === 'addresslistvoting.dao.eth';
```
WissenIstNacht, line 8: Just a head-up, you'll have conflicts here.

WissenIstNacht (changes requested):
Thanks a lot, looks like there were a couple of tricky things to solve.

What's the state on the proposals hook? `useProposal` says it should take over when `useDaoProposals` is deprecated, yet both seem quite alive?

packages/web-app/src/utils/library.ts
Fabricevladimir: This is a calback function that is going to be passed a key  `_` (which is unnecessary for what I'm checking for currently) and a value, which of course I actually need. So if I were to "remove" the first argument and keep only `value`, it would actually be the `key` and not the actual value that I need

packages/web-app/src/utils/library.ts
Fabricevladimir: True. That would also allow me to remove the try catch block

WissenIstNacht (approved):
Cheers!

PR #520, opened 2022-10-12
APP 1109 Goerli Migration

WissenIstNacht:
## Description

This PR updates the App to use the Görli Testnet.

Task: [1109](https://aragonassociation.atlassian.net/browse/APP-1057)

## Type of change

### DIFF
--- a/.github/workflows/pull_request_webapp_preview.yml
+++ b/.github/workflows/pull_request_webapp_preview.yml
@@ -56,11 +56,6 @@ jobs:
           VITE_REACT_APP_ALCHEMY_API_KEY: ${{secrets.VITE_REACT_APP_ALCHEMY_API_KEY}}
           VITE_IPFS_API_KEY: ${{secrets.VITE_IPFS_API_KEY}}
           VITE_IPFS_API_URL: ${{secrets.VITE_IPFS_API_URL}}
-          # TODO
-          # Remove when the SDK no longer needs the addresses of the contract
-          VITE_REACT_APP_SDK_DAO_FACTORY_ADDRESS: '0x2290E6dF695C5272cE942015c90aAe24bFB94960'
-          # "address1,address2,address3,...,addressN"
-          VITE_REACT_APP_SDK_WEB3_PROVIDERS: 'https://eth-rinkeby.alchemyapi.io/v2/bgIqe2NxazpzsjfmVmhj3aS3j_HZ9mpr'
           NODE_OPTIONS: '--max-old-space-size=6656'
       - uses: actions/setup-python@v2
         with:
--- a/packages/web-app/src/containers/selectChainForm/index.tsx
+++ b/packages/web-app/src/containers/selectChainForm/index.tsx
@@ -185,8 +185,8 @@ const networks: SelectableNetworks = {
     security: ['ethereum', 'arbitrum', 'polygon'],
   },
   test: {
-    cost: ['mumbai', 'arbitrum-test', 'rinkeby'],
-    popularity: ['mumbai', 'rinkeby', 'arbitrum-test'],
-    security: ['rinkeby', 'arbitrum-test', 'mumbai'],
+    cost: ['mumbai', 'arbitrum-test', 'goerli', 'rinkeby'],
+    popularity: ['mumbai', 'goerli', 'rinkeby', 'arbitrum-test'],
+    security: ['goerli', 'rinkeby', 'arbitrum-test', 'mumbai'],
   },
 };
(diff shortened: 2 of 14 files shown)


packages/web-app/src/utils/constants/chains.ts
```diff
@@ -136,6 +137,23 @@ export const CHAIN_METADATA: ChainList = {
       decimals: 18,
     },
   },
+  goerli: {
+    id: 4,
```
RakeshUP: id should be 5 i guess

packages/web-app/src/hooks/useDaoParam.tsx
```diff
@@ -24,9 +25,9 @@ export function useDaoParam() {
     error,
     loading: isLoading,
   } = useQuery(DAO_BY_ADDRESS, {
-    variables: {id: dao ? dao : ''},
-    client: client[network],
-    fetchPolicy: 'cache-and-network',
+    variables: {id: '0x1cab6f621a41438639e1f1b51c274ae65d41b8cb'},
+    client: client['goerli']!,
```
RakeshUP, line 29: do you think we should switch back to `client[network]` from the hardcoded value after the contracts have been deployed to all networks?

packages/web-app/src/containers/selectChainForm/index.tsx
```diff
@@ -185,8 +185,8 @@ const networks: SelectableNetworks = {
     security: ['ethereum', 'arbitrum', 'polygon'],
   },
   test: {
-    cost: ['mumbai', 'arbitrum-test', 'rinkeby'],
-    popularity: ['mumbai', 'rinkeby', 'arbitrum-test'],
-    security: ['rinkeby', 'arbitrum-test', 'mumbai'],
+    cost: ['mumbai', 'arbitrum-test', 'goerli', 'rinkeby'],
```
RakeshUP, line 188: i think 'rinkeby' can be removed from here safely. also do you plan on removing other references to `rinkeby` later at some point? if so can you create a ticket for that?

WissenIstNacht: Yup, I'll do this separately in [another PR](https://aragonassociation.atlassian.net/browse/APP-1111) 👍🏼

packages/web-app/src/hooks/useDaoParam.tsx
WissenIstNacht: tricky one :D

I was about to do that when I saw [this](https://github.com/aragon/apps/blob/41f7e0c6b1d02e6a5d927c8d229e431d158d51c1/packages/web-app/src/hooks/useDaoParam.tsx#L1). Let's see how this goes...

RakeshUP (approved):

.github/workflows/pull_request_webapp_preview.yml
```diff
@@ -56,11 +56,6 @@ jobs:
           VITE_REACT_APP_ALCHEMY_API_KEY: ${{secrets.VITE_REACT_APP_ALCHEMY_API_KEY}}
           VITE_IPFS_API_KEY: ${{secrets.VITE_IPFS_API_KEY}}
           VITE_IPFS_API_URL: ${{secrets.VITE_IPFS_API_URL}}
-          # TODO
-          # Remove when the SDK no longer needs the addresses of the contract
-          VITE_REACT_APP_SDK_DAO_FACTORY_ADDRESS: '0x2290E6dF695C5272cE942015c90aAe24bFB94960'
-          # "address1,address2,address3,...,addressN"
-          VITE_REACT_APP_SDK_WEB3_PROVIDERS: 'https://eth-rinkeby.alchemyapi.io/v2/bgIqe2NxazpzsjfmVmhj3aS3j_HZ9mpr'
```
Fabricevladimir, line 63: Will the removal of these not affect the flows? Especially preview and deploy?

Fabricevladimir (approved):
Migrate away, my dude

.github/workflows/pull_request_webapp_preview.yml
WissenIstNacht: As far as I can tell, no.
- `VITE_REACT_APP_SDK_DAO_FACTORY_ADDRESS` is not being used at all (since we're currently passing it manually when creating the SDK client objects).
- `VITE_REACT_APP_SDK_WEB3_PROVIDERS` is currently not used either. The info is accessed differently. Which is not ideal. But since the last part of this string is effectively an API key we're not winning awards here either haha

Either way, I'm working on a [subtask](https://aragonassociation.atlassian.net/browse/APP-1110) that's supposed to resolve these issues 👍🏼

PR #521, opened 2022-10-12
Feature: APP-1089 - Proposal Creation Nav: Settings Flow

Fabricevladimir:
## Description
- adds caching to settings proposal
- redirects to proposal details page on proposal successfully published
- extracts new proposal data to cached data function

NOTE: Please review _**after**_  #518 is merged so that the base branch commits don't distract. 

Task: [APP-1089](https://aragonassociation.atlassian.net/browse/APP-1089)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/context/createProposal.tsx
+++ b/packages/web-app/src/context/createProposal.tsx
@@ -7,7 +7,6 @@ import {
   ProposalCreationSteps,
 } from '@aragon/sdk-client';
 import {DaoAction} from '@aragon/sdk-client/dist/internal/interfaces/common';
-import {differenceInSeconds} from 'date-fns';
 import React, {useCallback, useEffect, useMemo, useState} from 'react';
 import {useFormContext} from 'react-hook-form';
 import {useTranslation} from 'react-i18next';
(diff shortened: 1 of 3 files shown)


RakeshUP (approved):

WissenIstNacht (approved):
Thanks, I think we're good.

I have a more general question, though. Why are the proposal stored as objects with `id` as keys? Compare this to the Deposit, where an array structure is used. I'm asking because in a couple of places, this seems to make the code more complicated than it needs to be (copying objects, working through them in for-loops, etc.). I think arrays and the functions they provide would have allowed for a much-needed level of abstraction in those places.

Fabricevladimir:
> Thanks, I think we're good.
> 
> I have a more general question, though. Why are the proposal stored as objects with `id` as keys? Compare this to the Deposit, where an array structure is used. I'm asking because in a couple of places, this seems to make the code more complicated than it needs to be (copying objects, working through them in for-loops, etc.). I think arrays and the functions they provide would have allowed for a much-needed level of abstraction in those places.

Ease of access really, meaning I don't have to filter an array each time I need to find the specific proposal. At the end of the day, I don't _actually_ gain any sort of performance boost really considering at no point are we going to have a significant enough number of cached proposals for that to really matter.

PR #525, opened 2022-10-18
Fix: App 1112 - Setting Flow Fixes

WissenIstNacht:
## Description

This PR fixes two outstanding issue with the Settings flow:
1. The comparison view now properly shows the DAOs current setting for "Old Settings"
2. The minimum participation was properly set throughout the flow

Task: [APP-1112](https://aragonassociation.atlassian.net/browse/APP-1112)

## Type of change


packages/web-app/src/containers/compareSettings/index.tsx
```diff
@@ -10,23 +10,67 @@ import {useTranslation} from 'react-i18next';
...
+    daoDetails?.plugins[0].id as PluginTypes
+  );
+
   const [selectedButton, setSelectedButton] = useState<'new' | 'old'>('new');
 
   const onButtonGroupChangeHandler = () => {
     setSelectedButton(prev => (prev === 'new' ? 'old' : 'new'));
   };
 
+  if (areDetailsLoading || areDetailsLoading || areSettingsLoading) {
```
RakeshUP: `areDetailsLoading` repeated

RakeshUP (approved):

packages/web-app/src/pages/editSettings.tsx
```diff
@@ -38,6 +38,7 @@ const EditSettings: React.FC = () => {
   const {network} = useNetwork();
   const {isMobile} = useScreen();
   const {breadcrumbs, icon, tag} = useMappedBreadcrumbs();
+  const {data: daoId, isLoading: paramAreLoading} = useDaoParam();
```
sepehr2github: Don't want to be sensitive but, I think the right spelling Is ```paramsAreLoading```

sepehr2github (commented):
Hey, Great Work thanks, Not sure if It was part of the PR or not but token information for token-based dao didn't show properly for me 

Screen Shot 2022-10-18 at 1 44 34 PM

WissenIstNacht:
Hey @sepehr2github , regarding your comment. There are two issues:

1. There's an extra row that got introduced during the copy-writing update.
2. There token information (id, name and symbol) are not showing.

The first one I fixed (see latest commit). The second one is a data-fetching issue that I consider out of the scope of this PR. The reason for this is that the error happens in the hook that fetches data. Therefore, this is a more generic issue that touches more than just settings). Also, that hook needs rewriting, as it still relies on direct GraphQl queries, rather than the SDK. And lastly, once the hook that fetches this data is fixed, this issue will be gone (no need to touch these files afterwards).

That being said, I thought about creating a bug ticket for it. But then I saw that Fabrice essentially already took care of this in one of his PRs.

sepehr2github (approved):
Thanks for the changes

PR #530, opened 2022-10-21
Fix/dao creation

WissenIstNacht:
## Description

Fixing an issue during DAO creation, where SDK would not throw an error (`Could not pin the metadata on IPFS`). This was the result of an empty object being passed as a parameter to the SDK.

**Note**
There are still issues with the flow:
- Signing the TX doesn't work. If I understood @sepehr2github  correctly, this is an issue with the SDK and is currently expected behaviour.
- When creating a new token, the encoding step leads to an uncaught error from deep withing the SDK.  As far as I can tell the data we're passing is OK. But this might need further investigation.

Task: [APP-1136](https://aragonassociation.atlassian.net/browse/APP-1136)

## Type of change

### DIFF
--- a/packages/web-app/src/context/createDao.tsx
+++ b/packages/web-app/src/context/createDao.tsx
@@ -5,32 +5,31 @@ import {
   IAddressListPluginInstall,
   ICreateParams,
   IErc20PluginInstall,
+  IPluginSettings,
 } from '@aragon/sdk-client';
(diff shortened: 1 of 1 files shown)


Fabricevladimir (approved):

sepehr2github (approved):
Seems good to me 👍

PR #531, opened 2022-10-25
Feature: App-917

sepehr2github:
## Description

- Add Ens Validation on DAO creation process

Task: [APP-917](https://aragonassociation.atlassian.net/browse/APP-917)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/utils/validators.ts
+++ b/packages/web-app/src/utils/validators.ts
@@ -172,8 +172,9 @@ export async function isDaoNameValid(value: string, provider: InfuraProvider) {
   if (isOnlyWhitespace(value)) return i18n.t('errors.required.name');
   try {
     const ensAddress = await provider?.resolveName(value.replaceAll(' ', '_'));
+    if (ensAddress) return i18n.t('errors.ensDuplication');
   } catch (err) {
-    return i18n.t('errors.required.name');
+    return i18n.t('errors.ensNetworkIssue');
   }
   return true;
 }
--- a/packages/web-app/src/utils/validators.ts
+++ b/packages/web-app/src/utils/validators.ts
@@ -175,6 +175,7 @@ export async function isDaoNameValid(value: string, provider: InfuraProvider) {
     ensAddress = await provider?.resolveName(value.replaceAll(' ', '_'));
 
     if (ensAddress) return i18n.t('errors.ensDuplication');
+    else return true;
   } catch (err) {
     return i18n.t('errors.ensNetworkIssue');
   }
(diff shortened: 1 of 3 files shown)


Fabricevladimir (approved):
Looks good to me

WissenIstNacht:
Thanks for the PR.

For some reason I can not make the validation error pop up when creating a DAO. Could you give me a DAO name that would result in a clash and therefore shows the error?

sepehr2github:
> Thanks for the PR.
> 
> For some reason I can not make the validation error pop up when creating a DAO. Could you give me a DAO name that would result in a clash and therefore shows the error?

Hey, There is no way to test this logic because the backend hasn't implemented the ens logic and It's entirely dummy data, I've just put the logic, couldn't tested it even by myself

packages/web-app/src/utils/validators.ts
```diff
@@ -165,3 +167,24 @@ export function actionsAreValid(
...
+    }, 50);
+
+    clearTimeout(timeoutBuffer.current);
+
+    if (ensAddress) return i18n.t('errors.ensDuplication');
+  } catch (err) {
+    return i18n.t('errors.ensNetworkIssue');
+  }
+  return true;
+}
```
WissenIstNacht: Yes, it is probably part of the problem as well that the SDK doesn't have that logic implemented. However, I think this code has issues as well.

See, this function is **always** going to return `true`. So even if the SDK's `resolveName` ever did return something useful, your validator function would never make use of it.

It's a bit much to explain in a comment. Let me know if you want to talk about it.

WissenIstNacht (changes requested):

packages/web-app/src/utils/validators.ts
```diff
@@ -165,3 +167,15 @@ export function actionsAreValid(
...
   return isValid;
 }
+
+export async function isDaoNameValid(value: string, provider: InfuraProvider) {
+  if (isOnlyWhitespace(value)) return i18n.t('errors.required.name');
+  let ensAddress = null;
+  try {
+    ensAddress = await provider?.resolveName(value.replaceAll(' ', '_'));
+
+    if (ensAddress) return i18n.t('errors.ensDuplication');
```
WissenIstNacht, line 177: Now you're never returning `true` :D

As far as I remember the true is necessary for when there is no error (or rather, when the input is actually valid). So I think adding `else return true` under this line would make it good to go.

WissenIstNacht (approved):
Thanks a lot for the changes.

One small modification left, but I'm approving this already :)

packages/web-app/src/utils/validators.ts
sepehr2github: I was about to add it ;D, thanks will fix it

PR #539, opened 2022-10-31
Fix: APP-1102

sepehr2github:
## Description

- add failed execution to the proposal details page

Task: [APP-1102](https://aragonassociation.atlassian.net/browse/APP-1102)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/components/executionWidget/index.tsx
+++ b/packages/web-app/src/components/executionWidget/index.tsx
@@ -88,16 +88,14 @@ const WidgetFooter: React.FC<FooterProps> = ({
 }) => {
   const {t} = useTranslation();
   switch (status) {
-    // TODO: Commented this piece of code to enable testing. Will be reverted before merging the PR
-    // case 'defeated':
-    //   return (
-    //     <AlertInline
-    //       label={t('governance.executionCard.status.defeated')}
-    //       mode={'warning'}
-    //     />
-    //   );
+    case 'defeated':
+      return (
+        <AlertInline
+          label={t('governance.executionCard.status.defeated')}
+          mode={'warning'}
+        />
+      );
     case 'executable':
-    case 'defeated' as 'executable':
       return (
         <Footer>
           <ButtonText
--- a/packages/web-app/src/components/executionWidget/index.tsx
+++ b/packages/web-app/src/components/executionWidget/index.tsx
@@ -11,6 +11,8 @@ import styled from 'styled-components';
 
 import {Action} from 'utils/types';
 import {ActionsFilter} from './actionsFilter';
+import {CHAIN_METADATA} from 'utils/constants';
+import {useNetwork} from 'context/network';
 
 export type ExecutionStatus =
   | 'defeated'
@@ -85,8 +87,15 @@ type FooterProps = Pick<
 const WidgetFooter: React.FC<FooterProps> = ({
   status = 'default',
   onExecuteClicked,
+  txhash,
 }) => {
   const {t} = useTranslation();
+  const {network} = useNetwork();
+
+  const handleTxViewButtonClick = () => {
+    window.open(CHAIN_METADATA[network].explorer + 'tx/' + txhash, '_blank');
+  };
+
   switch (status) {
     case 'defeated':
       return (
@@ -120,6 +129,7 @@ const WidgetFooter: React.FC<FooterProps> = ({
             iconRight={<IconLinkExternal />}
             size="large"
             bgWhite
+            onClick={handleTxViewButtonClick}
           />
 
           <AlertInline
@@ -137,6 +147,7 @@ const WidgetFooter: React.FC<FooterProps> = ({
             iconRight={<IconLinkExternal />}
             size="large"
             bgWhite
+            onClick={handleTxViewButtonClick}
           />
 
           <AlertInline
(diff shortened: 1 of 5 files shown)


packages/web-app/src/context/proposalTransaction.tsx
```diff
@@ -315,9 +317,11 @@ const ProposalTransactionProvider: React.FC<Props> = ({children}) => {
             break;
         }
         setExecuteParams(undefined);
+        setExecutionFailed(false);
```
Fabricevladimir, line 322: Can you set the transaction hash for both failure and success? That way it gets passed to the `ExecutionWidget` as well

packages/web-app/src/context/proposalTransaction.tsx
```diff
@@ -78,6 +79,7 @@ const ProposalTransactionProvider: React.FC<Props> = ({children}) => {
 
   const [executeParams, setExecuteParams] = useState<IExecuteProposalParams>();
   const [executeSubmitted, setExecuteSubmitted] = useState(false);
+  const [executionFailed, setExecutionFailed] = useState(true);
```
Fabricevladimir: Don't forget to default this to `false` before you merge unless you're going to clear this up after qa

packages/web-app/src/components/executionWidget/index.tsx
```diff
@@ -96,7 +96,7 @@ const WidgetFooter: React.FC<FooterProps> = ({
     //       mode={'warning'}
     //     />
     //   );
-    // case 'executable':
+    case 'executable':
     case 'defeated' as 'executable':
```
Fabricevladimir: Please remove this `case 'defeated' as 'executable'` and uncomment the block preceding for the `defeated` case.

packages/web-app/src/hooks/useDaoProposal.tsx
```diff
@@ -30,6 +30,14 @@ import {customJSONReplacer} from 'utils/library';
  * @returns a detailed proposal
  */
 
+export enum ProposalStatus {
```
Fabricevladimir: I assume this is for demonstration purposes, but please remember to cleanup before merge. Also, SDK exposes a `ProposalStatus` already

packages/web-app/src/hooks/useDaoProposal.tsx
```diff
@@ -154,7 +162,10 @@ export const useDaoProposal = (
         if (proposal && encodedActions) proposal.actions = encodedActions;
 
         if (proposal) {
-          setData({...augmentProposalWithCache(proposal)});
+          setData({
+            ...augmentProposalWithCache(proposal),
+            status: ProposalStatus.SUCCEEDED,
```
Fabricevladimir: Same here, don't forget to cleanup before merge

packages/web-app/src/hooks/useDaoProposal.tsx
```diff
@@ -164,7 +175,9 @@ export const useDaoProposal = (
             }
           }
         } else if (cachedProposal) {
-          setData({...augmentProposalWithCache(cachedProposal)});
+          setData({
```
Fabricevladimir, line 167: [Not a code thing] I'm not sure why this broke up into multiple lines, do we have different formatter issues creeping in again? 😬

packages/web-app/src/utils/proposals.ts
```diff
@@ -299,12 +300,26 @@ export function getProposalStatusSteps(
...
-
-    // TODO - Execution failed: no execution date treated as failure for now
+      if (executionFailed)
+        return [
+          ...getPassedProposalSteps(creationDate, startDate, publishedBlock),
+          {
+            label: i18n.t('governance.statusWidget.failed'),
+            mode: 'failed',
+            date: `${format(
+              endDate,
```
Fabricevladimir: This should be the **execution date** and not `endDate`. Because the SDK proposal will _not_ have the execution date of  a failed transaction, this should then be `new Date()` (ie now) or alternatively a date stored on the context handling the action execution (this would be more accurate technically especially if the failure is to be cached, which we haven't actually talked about but I'm not sure it's worth the hassle right now).

Fabricevladimir (changes requested):
Looking good Sepher! Left a couple comments here and there, mostly about cleaning up before or after merge.

packages/web-app/src/hooks/useDaoProposal.tsx
sepehr2github: No, It might be a VS small issue

Fabricevladimir (approved):
Sweet 🍬

packages/web-app/src/components/executionWidget/index.tsx
```diff
@@ -122,6 +129,7 @@ const WidgetFooter: React.FC<FooterProps> = ({
             iconRight={<IconLinkExternal />}
             size="large"
             bgWhite
+            onClick={handleTxViewButtonClick}
```
WissenIstNacht, line 132: Does this case make sense? Is there a TX when the execution fails?

WissenIstNacht (approved):
I assume my comment isn't relevant, otherwise you or Fabrice would have noticed. Therefore, I'm approving this. If you think my comment is relevant let me know and maybe drop a todo. :)

packages/web-app/src/components/executionWidget/index.tsx
sepehr2github: AFAIK we get tx hash for failed executions, but good point will check with jose and create another ticket if needed on our todo list

PR #546, opened 2022-11-07
Fix: APP-1153

sepehr2github:
## Description

- Replace TimeInputes on proposal creation with the UI component versions
- Refactor TimeInput to fix render loop issue

Note: Didn't fix the input height issue on the desktop because Vale & Selim will re-review all the inputs in the upcoming days

Task: [APP-1153](https://aragonassociation.atlassian.net/browse/APP-1153)

## Type of change

## Checklist:

- [] I have updated the ``CHANGELOG.md`` file in the root folder.


RakeshUP (approved):

WissenIstNacht:
Hey, thanks a lot for taking a look at this. Two things:

1. It's nice you tried to remove the `SimplifiedInput` and use the one from the
component library instead. Unfortunately, there's a little issue with that. The
`SimplifiedInput` was introduced on purpose, because there were issues between
design and implementation that we couldn't resolve. So we made something really
simple in the meantime. One of the issues, for example, is that in a lot of
countries, the "AM" and "PM" system is not a thing. So "17:30PM" makes no sense.
On top of that, there were some weird differences between Firefox and Chrome
when it comes to deciding which time format to use. 
2. What do you mean by "fix render loop issue"? I couldn't find any info on this. What is it and what part of the code solves this?

My suggestion would be to keep the changes you've made to the ui component, as they're actually good changes and a good step in the right direction. But, at the same time, keep using the simplified input in the web-app.

Let me know what you think. :)

sepehr2github:
> Hey, thanks a lot for taking a look at this. Two things:
> 
> 1. It's nice you tried to remove the `SimplifiedInput` and use the one from the
>    component library instead. Unfortunately, there's a little issue with that. The
>    `SimplifiedInput` was introduced on purpose, because there were issues between
>    design and implementation that we couldn't resolve. So we made something really
>    simple in the meantime. One of the issues, for example, is that in a lot of
>    countries, the "AM" and "PM" system is not a thing. So "17:30PM" makes no sense.
>    On top of that, there were some weird differences between Firefox and Chrome
>    when it comes to deciding which time format to use.
> 2. What do you mean by "fix render loop issue"? I couldn't find any info on this. What is it and what part of the code solves this?
> 
> My suggestion would be to keep the changes you've made to the ui component, as they're actually good changes and a good step in the right direction. But, at the same time, keep using the simplified input in the web-app.
> 
> Let me know what you think. :)

Aha, Now I see, Ok I reversed all of those changes on the web app, Also regarding the loop issue, We've been using a state on (TimeInput) component and It was causing some useless re-renders (renders loop indeed) when you wanted to use the input so by this new component refactor, that problem has been solved as well

packages/web-app/src/containers/setupVotingForm/index.tsx
```diff
@@ -31,6 +30,7 @@ import {
 import {DateTimeErrors} from './dateTimeErrors';
 import {useGlobalModalContext} from 'context/globalModals';
 import {StringIndexed} from 'utils/types';
+import {SimplifiedTimeInput} from 'components/inputTime/inputTime';
```
WissenIstNacht, line 33: 😜

WissenIstNacht (approved):

PR #563, opened 2022-11-17
Feature: APP-1293 & APP-1292

sepehr2github:
## Description

- Verification modal has been added
- Etherscan validation has been added

Task: [APP-1293](https://aragonassociation.atlassian.net/browse/APP-1293) - [APP-1292](https://aragonassociation.atlassian.net/browse/APP-1292)

Note: The Etherscan API Verification needs an account from the DevOps team, once the account got ready, The verification method will be replaced by the new API

## Checklist:

### DIFF
--- a/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
+++ b/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
@@ -17,9 +17,9 @@ import {Controller, useFormContext, useWatch} from 'react-hook-form';
 import {CHAIN_METADATA, TransactionState} from 'utils/constants';
 import {useNetwork} from 'context/network';
 import {validateContract} from 'utils/validators';
-import {useProviders} from 'context/providers';
 import {useAlertContext} from 'context/alert';
 import {isAddress} from 'ethers/lib/utils';
+import {EtherscanContractResponse} from 'utils/types';
 
 type Props = {
   isOpen: boolean;
@@ -43,12 +43,14 @@ const ContractAddressValidation: React.FC<Props> = props => {
   );
   const {control} = useFormContext();
   const {network} = useNetwork();
-  const {infura: provider} = useProviders();
   const addressField = useWatch({
     name: 'contractAddress',
     control,
   });
   const {alert} = useAlertContext();
+  const [contractData, setContractData] = useState<
+    EtherscanContractResponse | undefined
+  >();
 
   const label = {
     [TransactionState.WAITING]: t('scc.addressValidation.actionLabelWaiting'),
@@ -57,9 +59,14 @@ const ContractAddressValidation: React.FC<Props> = props => {
     [TransactionState.ERROR]: t('scc.addressValidation.tryAgain'),
   };
 
-  const setContractValid = useCallback((value: boolean) => {
-    if (value) setTransactionState(TransactionState.SUCCESS);
-    else setTransactionState(TransactionState.ERROR);
+  const setContractValid = useCallback(value => {
+    if (value) {
+      setTransactionState(TransactionState.SUCCESS);
+      setContractData(value);
+    } else {
+      setTransactionState(TransactionState.ERROR);
+      setContractData(value);
+    }
   }, []);
 
   // clear field when there is a value, else paste
@@ -137,7 +144,7 @@ const ContractAddressValidation: React.FC<Props> = props => {
           label={label[transactionState] as string}
           onClick={async () => {
             setTransactionState(TransactionState.LOADING);
-            setContractValid(await validateContract(addressField, provider));
+            setContractValid(await validateContract(addressField));
           }}
           iconLeft={
             transactionState === TransactionState.LOADING ? (
@@ -153,7 +160,11 @@ const ContractAddressValidation: React.FC<Props> = props => {
         {transactionState === TransactionState.SUCCESS && (
           <AlertInlineContainer>
             <AlertInline
-              label={t('scc.addressValidation.successLabel') as string}
+              label={
+                t('scc.addressValidation.successLabel', {
+                  contractName: contractData?.ContractName,
+                }) as string
+              }
               mode="success"
             />
           </AlertInlineContainer>
--- a/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
+++ b/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
@@ -12,6 +12,8 @@ import {
   Spinner,
   ValueInput,
 } from '@aragon/ui-components';
+import {shortenAddress} from '@aragon/ui-components/src/utils/addresses';
+
 import {handleClipboardActions} from 'utils/library';
 import {Controller, useFormContext, useWatch} from 'react-hook-form';
 import {CHAIN_METADATA, TransactionState} from 'utils/constants';
@@ -52,6 +54,9 @@ const ContractAddressValidation: React.FC<Props> = props => {
     EtherscanContractResponse | undefined
   >();
 
+  const isTransactionSuccessful = transactionState === TransactionState.SUCCESS;
+  const isTransactionLoading = transactionState === TransactionState.LOADING;
+
   const label = {
     [TransactionState.WAITING]: t('scc.addressValidation.actionLabelWaiting'),
     [TransactionState.LOADING]: t('scc.addressValidation.actionLabelLoading'),
@@ -82,9 +87,8 @@ const ContractAddressValidation: React.FC<Props> = props => {
   );
 
   const isButtonDisabled = useMemo(
-    () =>
-      transactionState === TransactionState.SUCCESS || !isAddress(addressField),
-    [addressField, transactionState]
+    () => isTransactionSuccessful || !isAddress(addressField),
+    [addressField, isTransactionSuccessful]
   );
 
   return (
@@ -95,7 +99,7 @@ const ContractAddressValidation: React.FC<Props> = props => {
         onBackButtonClicked={props.onBackButtonClicked}
         disabled={
           transactionState === TransactionState.LOADING ||
-          transactionState === TransactionState.SUCCESS
+          isTransactionSuccessful
         }
       />
       <Content>
@@ -112,7 +116,6 @@ const ContractAddressValidation: React.FC<Props> = props => {
         </DescriptionContainer>
         <Controller
           name="contractAddress"
-          control={control}
           rules={{
             required: t('errors.required.tokenAddress') as string,
             validate: value => isAddress(value) || t('errors.invalidAddress'),
@@ -127,9 +130,10 @@ const ContractAddressValidation: React.FC<Props> = props => {
                 mode={error ? 'critical' : 'default'}
                 ref={ref}
                 name={name}
-                value={value}
+                value={isTransactionSuccessful ? shortenAddress(value) : value}
                 onBlur={onBlur}
                 onChange={onChange}
+                disabled={isTransactionSuccessful || isTransactionLoading}
                 placeholder="0x ..."
                 adornmentText={value ? t('labels.clear') : t('labels.paste')}
                 onAdornmentClick={() => handleAdornmentClick(value, onChange)}
@@ -147,17 +151,17 @@ const ContractAddressValidation: React.FC<Props> = props => {
             setContractValid(await validateContract(addressField));
           }}
           iconLeft={
-            transactionState === TransactionState.LOADING ? (
+            isTransactionLoading ? (
               <Spinner size="xs" color="white" />
             ) : undefined
           }
           iconRight={icons[transactionState]}
-          isActive={transactionState === TransactionState.LOADING}
+          isActive={isTransactionLoading}
           disabled={isButtonDisabled}
           size="large"
           className="mt-3 w-full"
         />
-        {transactionState === TransactionState.SUCCESS && (
+        {isTransactionSuccessful && (
           <AlertInlineContainer>
             <AlertInline
               label={
--- a/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
+++ b/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
@@ -86,7 +86,6 @@ const ContractAddressValidation: React.FC<Props> = props => {
   );
 
   const addressValidator = (value: string) => {
-    console.log('check', value);
     if (isAddress(value)) return true;
     return t('errors.invalidAddress') as string;
   };
--- a/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
+++ b/packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
@@ -78,12 +78,12 @@ const ContractAddressValidation: React.FC<Props> = props => {
   const handleAdornmentClick = useCallback(
     (value: string, onChange: (value: string) => void) => {
       // when there is a value clear it
-      if (value && !isTransactionSuccessful) {
+      if (value && !isTransactionSuccessful && !isTransactionLoading) {
         onChange('');
         alert(t('alert.chip.inputCleared'));
       } else handleClipboardActions(value, onChange, alert);
     },
-    [alert, isTransactionSuccessful, t]
+    [alert, isTransactionLoading, isTransactionSuccessful, t]
   );
 
   const handleVerificationClick = async () => {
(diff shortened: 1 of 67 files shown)


packages/web-app/src/locales/en/common.json
```diff
@@ -733,6 +734,16 @@
       "title": "No Smart Contract",
       "description": "Add a smart contract to interact with its actions",
       "primaryBtnLabel": "Connect Smart Contract"
+    },
+    "addressValidation": {
+      "modalTitle": "Smart Contract Composer",
```
Fabricevladimir: This should be "Connect Smart Contract"

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+        />
+        <ButtonText
+          label={label[transactionState] as string}
+          onClick={async () => {
+            setTransactionState(TransactionState.LOADING);
+            await validateContract(addressField, provider, setContractValid);
+          }}
+          iconLeft={icons[transactionState]}
+          isActive={transactionState === TransactionState.LOADING}
+          disabled={addressField === ''}
```
Fabricevladimir: Because when the form first loads, `addressField` is undefined, the condition returns false, enabling the button by default instead of disabling it. I think this can be fixed by either checking explicitly if the `addressField` is undefined or by adding the types and default values to the `useForm` declaration on the `demoSCC` page. I lean slightly towards the types if only because there's a lot that will need to be connected and I can see the types benefitting us when we wire up the modals to the form. Choice is yours 👍🏾 

Also, it might be a good idea to check if `addressField` is not an empty space character, because that will trigger the button state to default and not disabled.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+          render={({
+            field: {name, onBlur, onChange, value, ref},
+            fieldState: {error},
+          }) => (
+            <>
+              <ValueInput
+                mode={error ? 'critical' : 'default'}
+                ref={ref}
+                name={name}
+                value={value}
```
Fabricevladimir: It would seem per the designs, that the contract address is shortened in the field so that the user can see both the starting and ending characters

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+              label={t('labels.etherscan') as string}
+              href={`${CHAIN_METADATA[network].explorer}`}
+            />
+          </Description>
+        </DescriptionContainer>
+        <Controller
+          name="contractAddress"
+          control={control}
+          rules={{
+            required: t('errors.required.tokenAddress') as string,
```
Fabricevladimir, line 149: Not sure if this is within the scope, but do we need custom validation for the address? For example, if the user inputs "abc", this should probably show an error and still keep the button disabled.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+                mode={error ? 'critical' : 'default'}
+                ref={ref}
+                name={name}
+                value={value}
+                onBlur={onBlur}
+                onChange={onChange}
+                placeholder="0x ..."
+                adornmentText={value ? t('labels.copy') : t('labels.paste')}
+                onAdornmentClick={() =>
+                  handleClipboardActions(value, onChange, alert)
```
Fabricevladimir: This alert seems to be the _browser_ alert and not the `AlertChip` component

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+            <>
+              <ValueInput
+                mode={error ? 'critical' : 'default'}
+                ref={ref}
+                name={name}
+                value={value}
+                onBlur={onBlur}
+                onChange={onChange}
+                placeholder="0x ..."
+                adornmentText={value ? t('labels.copy') : t('labels.paste')}
```
Fabricevladimir: The adornment should be `Clear` instead of  `Copy`

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+          rules={{
+            required: t('errors.required.tokenAddress') as string,
+          }}
+          defaultValue={''}
+          render={({
+            field: {name, onBlur, onChange, value, ref},
+            fieldState: {error},
+          }) => (
+            <>
+              <ValueInput
```
Fabricevladimir, line 160: While the contract is validating and when it's successfully validated, the input should be disabled.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+  };
+
+  const setContractValid = useCallback((value: boolean) => {
+    if (value) setTransactionState(TransactionState.SUCCESS);
+    else setTransactionState(TransactionState.ERROR);
+  }, []);
+
+  return (
+    <ModalBottomSheetSwitcher isOpen={props.isOpen} onClose={props.onClose}>
+      <ModalHeader
```
Fabricevladimir, line 122: While the contract is validating and has been successfully validated, the back button should be removed/hidden.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+            </>
+          )}
+        />
+        <ButtonText
+          label={label[transactionState] as string}
+          onClick={async () => {
+            setTransactionState(TransactionState.LOADING);
+            await validateContract(addressField, provider, setContractValid);
+          }}
+          iconLeft={icons[transactionState]}
```
Fabricevladimir: When the transaction is successful, the icon should be `iconRight` instead of `iconLeft`

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const addressField = useWatch({
+    name: 'contractAddress',
+    control,
+  });
+
+  const label = {
+    [TransactionState.WAITING]: t('scc.addressValidation.actionLabelWaiting'),
+    [TransactionState.LOADING]: t('scc.addressValidation.actionLabelLoading'),
+    [TransactionState.SUCCESS]: t('scc.addressValidation.actionLabelSuccess'),
+    [TransactionState.ERROR]: t('scc.addressValidation.tryAgain'),
```
Fabricevladimir: Two things concerning the error state:
1. Once the user has hit the error state, I expect clearing the input to reset the state to `WAITING`, which should reset the button label back to the `WAITING` label. That said, can you check with Selim to make sure I'm not just spouting crap? 🤣
2. We don't necessarily have the designs for this, but I think it would make sense to keep the error state consistent with the ABI verification branch error state, meaning, setting the button label to "Retry" and showing the error, "Contract could not be verified" , on the `AlertInline` below the button. Thoughts? 🤔

packages/web-app/src/locales/en/common.json
```diff
@@ -733,6 +734,16 @@
...
+      "title": "Smart Contract Address",
+      "description": "Enter the desired smart contract address. It is best to copy this address directly from",
+      "actionLabelWaiting": "Validate Smart Contract",
+      "actionLabelLoading": "Validate Smart Contract ...",
```
Fabricevladimir: Small typo, text on the designs has it as "Validating Smart Contract ...". Of course, user testing might have us scrambling to change it 😄

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,155 @@
...
+                <AlertInline label={error.message} mode="critical" />
+              )}
+            </>
+          )}
+        />
+        <ButtonText
+          label={label[transactionState] as string}
+          onClick={async () => {
+            setTransactionState(TransactionState.LOADING);
+            await validateContract(addressField, provider, setContractValid);
```
Fabricevladimir: Curious, why not pass in the result of `validateContract` to `setContractValid` instead like so:
```javascript
setContractValid(await validateContract(addressField, provider));
```

No change necessary, just curious.

Fabricevladimir (changes requested):
Thanks Sepehr; left a few comments

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
jamesej: I imagine the full address should be shown if there's enough width in the component, the shortening is a fallback to ensure the user can see the start and end of the address.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
jamesej: Another thing that would be nice to include is to create a `FieldValues` type and use it as a generic parameter to `useFormContext` and `useWatch` so that `addressField` gets a non-`any` type.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
jamesej: I think you should trim whitespace off `addressField` before checking it as it's very easy to cut and paste extra whitespace around something.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
sepehr2github: This is something we could fix in the integration phase when we want to integrate the whole flow, what do you think?

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
sepehr2github: > Another thing that would be nice to include is to create a `FieldValues` type and use it as a generic parameter to `useFormContext` and `useWatch` so that `addressField` gets a non-`any` type.

This is sth that I wanted to do but thought might be good wait for the integration phase, we can add these types later

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
Fabricevladimir: I'm not sure that they have much to do with the integration. The integration is just supposed to link the different parts together

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -54,17 +63,49 @@ const ContractAddressValidation: React.FC<Props> = props => {
...
+      if (value) {
+        onChange('');
+        alert(t('alert.chip.inputCleared'));
+      } else handleClipboardActions(value, onChange, alert);
+    },
+    [alert, t]
+  );
+
+  const addressValidator = (value: string) => {
+    console.log('check', value);
```
Fabricevladimir: I see you

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -27,7 +30,7 @@ type Props = {
 
 const icons = {
   [TransactionState.WAITING]: undefined,
```
Fabricevladimir, line 37: This is a general comment and it might be a preference thing, but there is no actual transaction on this flow. It sort of throws you off because the namings are transaction related if that makes sense. It's something that could have been handled with the isLoading, isError pattern we've used already. I'm not requesting changes there, but a rename might go a long way

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
Fabricevladimir: The issue with waiting for later is that the person who does the integration doesn't necessarily know all the form fields you've used and what the nuances in terms of typing

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -54,17 +63,49 @@ const ContractAddressValidation: React.FC<Props> = props => {
...
+  );
+
+  const addressValidator = (value: string) => {
+    console.log('check', value);
+    if (isAddress(value)) return true;
+    return t('errors.invalidAddress') as string;
+  };
+
+  const isButtonDisabled = useMemo(
+    () => isTransactionSuccessful || !isAddress(addressField),
```
Fabricevladimir: When the contract has been successfully validated, the button shouldn't be disabled.

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -80,28 +121,27 @@ const ContractAddressValidation: React.FC<Props> = props => {
...
                 onBlur={onBlur}
+                value={value}
                 onChange={onChange}
+                disabled={isTransactionSuccessful || isTransactionLoading}
                 placeholder="0x ..."
-                adornmentText={value ? t('labels.copy') : t('labels.paste')}
-                onAdornmentClick={() =>
-                  handleClipboardActions(value, onChange, alert)
-                }
+                adornmentText={value ? t('labels.clear') : t('labels.paste')}
```
Fabricevladimir: Once the contract itself has been validated with your `addressValidator`, 'clear' should should transition to 'copy', which once we've reenabled the adornment, will allow the user to, well, copy the address

packages/web-app/src/utils/validators.ts
```diff
@@ -190,17 +194,22 @@ export async function isDaoNameValid(value: string, provider: InfuraProvider) {
...
+
   try {
-    // TODO: This Method is temporary and Will be replaced by etherscan API Verification
-    const code = await provider.getCode(address);
-    setIsContractValid(code !== '0x');
+    const res = await fetch(url);
+    const data = await res.json();
+    if (data.result[0].ABI !== 'Contract source code not verified')
+      return data.result[0];
+    else return;
```
Fabricevladimir: Unnecessary to return I think

packages/web-app/src/utils/validators.ts
```diff
@@ -190,17 +194,22 @@ export async function isDaoNameValid(value: string, provider: InfuraProvider) {
...
   } catch (error) {
-    setIsContractValid(false);
     console.log(error);
+    return;
```
Fabricevladimir: Same here

Fabricevladimir (changes requested):

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
sepehr2github: Yes, I meant the second point you mentioned, whoever did the ABI thing should add contract ABI verification and these minor changes

Fabricevladimir (approved):

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -0,0 +1,231 @@
...
+    verificationState === TransactionState.SUCCESS;
+  const isTransactionLoading = verificationState === TransactionState.LOADING;
+
+  const label = {
+    [TransactionState.WAITING]: t('scc.addressValidation.actionLabelWaiting'),
+    [TransactionState.LOADING]: t('scc.addressValidation.actionLabelLoading'),
+    [TransactionState.SUCCESS]: t('scc.addressValidation.actionLabelSuccess'),
+    [TransactionState.ERROR]: '',
+  };
+
```
jamesej, line 73: Just a comment maybe for future reference, and this is not about your code but code that was already written, in my view in Typescript it's nicer not to use enum types like TransactionState.WAITING here but to define
`type TransactionState = "waiting" | "loading" | "success" | "error";`
This code would be more concise and readable had that been done. Enum types are good where the name of the thing is different to the value of the thing.

jamesej (approved):

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
sepehr2github: Great, Point. Thanks. Will keep this in mind

PR #574, opened 2022-11-27
Feature: Wallet input

sepehr2github:
## Description

- Wallet input added to UI components
- pumped SDK version

Task: [APP-1323](https://aragonassociation.atlassian.net/browse/APP-1323)

## Type of change

## Checklist:


packages/ui-components/src/components/input/walletInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+  /** Text that appears on the button present on the right side of the input; if no text
+   * is provided, the button will not be rendered */
+  adornmentText?: string;
+  /** Handler for when the button present on the right side of the input  is
+   * clicked */
+  onAdornmentClick?: () => void;
+  /** Changes a input's color schema */
+  mode?: 'default' | 'success' | 'warning' | 'critical';
+  /** Disable the input but keep the adornment button active */
+  disabledFilled?: boolean;
```
Fabricevladimir, line 16: Because we don't have a situation where the adornment is ever actually disabled, I think this is more a situation and less of a state, meaning, perhaps this property is unnecessary. I find it might  make more sense to keep the interface smaller by using the `disabled` property that comes out of the box instead to conditionally render the adornment if there is a value in the disabled input. Also, as it stands, passing `disabled` will override `disabledFilled`

packages/ui-components/src/components/input/walletInput.tsx
```diff
@@ -0,0 +1,85 @@
...
+    } else if (mode === 'warning') {
+      className += 'border-warning-600';
+    } else if (mode === 'critical') {
+      className += 'border-critical-600';
+    }
+
+    return {className};
+  }
+)<StyledContainerProps>`
+  :focus-within {
```
Fabricevladimir, line 82: TW has a focus within, can you use that one instead?

Also, when I focus-within, the input still has an outline: 
image

packages/web-app/src/containers/smartContractComposer/contractAddressValidation.tsx
```diff
@@ -157,13 +157,13 @@ const ContractAddressValidation: React.FC<Props> = props => {
           }) => (
             //TODO: This value input needs to replace with wallet input
```
Fabricevladimir: I think you can remove this now. Also, do you know if we have a task to update the other places on the webapp that requires the `WalletInput`?

Fabricevladimir (changes requested):
Thanks, Sepehr! I left a couple of comments

packages/ui-components/src/components/input/walletInput.tsx
sepehr2github: I couldn't get what you want from me. could you explain a bit more?

jamesej (approved):
LGMT if Fabrice's comment at walletInput.tsx:16 is resolved.

packages/ui-components/src/components/input/walletInput.tsx
Fabricevladimir: I'm still getting this when I run Storybook

Fabricevladimir (changes requested):

packages/ui-components/src/components/input/walletInput.tsx
Fabricevladimir: Resolved with #583

Fabricevladimir (approved):

PR #585, opened 2022-12-05
Feature: APP-1300 - Change ghost variant background

Fabricevladimir:
## Description
- Switches `ButtonIcon` and `ButtonText` ghost variant background to transparent

Feels like there should be more components with ghost variants ... 😰 

Task: [APP-1300](https://aragonassociation.atlassian.net/browse/APP-1300)

## Type of change

## Checklist:


RakeshUP (approved):

sepehr2github (approved):
Thanks, Fabrice, How about other components? should we create another pr for addressing them?

WissenIstNacht:
Quick comment: In the past, `ghost` was sometimes (erroneously) used instead of `secondary` with `bgWhite`. One such case is the `ButtonIcon` used in the `InputNumber`. 

IN general I don't think the ROI is good enough to go through every component both in the library and in the app. But just something to be aware of.

PR #589, opened 2022-12-07
Fix: Deposit APP-1325

sepehr2github:
## Description

- Fix Native & Erc20 depositing issue

Task: [APP-1325](https://aragonassociation.atlassian.net/browse/APP-1325)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/context/deposit.tsx
+++ b/packages/web-app/src/context/deposit.tsx
@@ -4,6 +4,7 @@ import {
   TransferType,
 } from '@aragon/sdk-client';
 import {useFormContext} from 'react-hook-form';
+import {Web3Provider} from '@ethersproject/providers';
 import {generatePath, useNavigate, useParams} from 'react-router-dom';
 import React, {
   createContext,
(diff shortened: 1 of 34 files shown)


packages/web-app/src/context/deposit.tsx
```diff
@@ -226,27 +226,42 @@ const DepositProvider = ({children}: {children: ReactNode}) => {
...
+                  from,
+                  amount: depositParams?.amount,
+                  reference,
+                  type: TransferType.DEPOSIT,
+                  tokenType: 'erc20',
+                  token: {
+                    name: tokenName,
+                    address: tokenAddress,
+                    symbol: tokenSymbol,
+                    // TODO: Fix the decimals value
```
Fabricevladimir: 👀

packages/web-app/src/context/deposit.tsx
```diff
@@ -29,6 +29,7 @@ import {useGlobalModalContext} from './globalModals';
 import {useReactiveVar} from '@apollo/client';
```
Fabricevladimir, line 34: Hey not really part of this specific task, but wherever you're running the generator functions for the deposit, can you move the `try/catch` _around_ the whole `for of` loop instead of having it inside? Metamask update is returning rejected signature errors weirdly and so they won't be caught unless it's the `try/catch` is around the whole thing. The modal will hang despite user having cancelled the transaction.

Fabricevladimir (approved):
Thanks, I can deposit. Left a couple of comments regardless

RakeshUP (approved):

packages/web-app/src/context/deposit.tsx
sepehr2github: Good point will fix in second refactore

PR #599, opened 2022-12-13
Fix: Minimum duration in setup voting step of New Proposal flow

RakeshUP:
## Description

There are some UI artifacts that are left as such for now, as there's going to be a redesign in this step.

Task: [APP-1381](https://aragonassociation.atlassian.net/browse/APP-1381)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/containers/setupVotingForm/index.tsx
+++ b/packages/web-app/src/containers/setupVotingForm/index.tsx
@@ -329,10 +329,10 @@ const SetupVotingForm: React.FC = () => {
                 <Controller
                   name="duration"
                   control={control}
-                  defaultValue={days}
+                  defaultValue={days + 1}
                   rules={{
                     min: {
-                      value: days || 0,
+                      value: days + 1 || 0,
                       message: t('errors.durationTooShort'),
                     },
                     required: t('errors.required.duration'),
@@ -360,7 +360,7 @@ const SetupVotingForm: React.FC = () => {
           ) : (
             <>
               <div className="block space-y-2">
-                {days && days >= 1 && (
+                {days && days >= 1 ? (
                   <div>
                     <Controller
                       name="durationSwitch"
@@ -379,7 +379,7 @@ const SetupVotingForm: React.FC = () => {
                       }}
                     />
                   </div>
-                )}
+                ) : null}
                 <HStack>
                   <Controller
                     name="endDate"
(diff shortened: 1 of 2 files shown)


packages/web-app/src/containers/setupVotingForm/index.tsx
```diff
@@ -357,66 +326,135 @@ const SetupVotingForm: React.FC = () => {
...
+                <AlertInline
+                  label={formState.errors.duration.message}
+                  mode="critical"
+                />
+              )}
+            </>
+          ) : (
+            <>
+              <div className="block space-y-2">
+                {days && days >= 1 && (
```
WissenIstNacht: This displays `0` (literally the character 0 in the UI) when the check doesn't pass. Changing to `days && days >= 1 ? /*component*/ : null` does the trick.

packages/web-app/src/containers/setupVotingForm/index.tsx
```diff
@@ -357,66 +326,135 @@ const SetupVotingForm: React.FC = () => {
                     );
                   }}
                 />
-              </div>
-              <HStack>
                 <Controller
-                  name="endDate"
+                  name="duration"
                   control={control}
+                  defaultValue={days}
```
WissenIstNacht: You'll have to go round up on `default value` and `min` in the input intself. Because if it's not exactly a day you'll be short on the duration. I guess just doing days+1 should do the trick. If people want more granular control they can use the date+time control. But up to you.

WissenIstNacht (changes requested):
Thanks!

sepehr2github (approved):
Thanks Rakash, Once you addressed Vale PRs, It's good to go

packages/web-app/src/containers/setupVotingForm/index.tsx
RakeshUP: Thanks that's a good catch!

packages/web-app/src/containers/setupVotingForm/index.tsx
RakeshUP: Changing to days+1 for now. I guess this step is going to be redesigned pretty soon and Majority Voting plugin is incorporated into the designs. We can come back at that time to make this component more robust

WissenIstNacht (approved):

PR #611, opened 2022-12-21
Fix: Actions List - APP-1421

sepehr2github:
## Description

- Fix proposal title next page validation
- Fix the actions list on the proposal details page

Task: [APP-1421](https://aragonassociation.atlassian.net/browse/APP-1421)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/pages/proposal.tsx
+++ b/packages/web-app/src/pages/proposal.tsx
@@ -204,11 +204,11 @@ const Proposal: React.FC = () => {
           mintTokenActions.actions.length,
           decodedMintToken
         );
-
-        Promise.all(actionPromises).then(value => {
-          setDecodedActions(value);
-        });
       }
+
+      Promise.all(actionPromises).then(value => {
+        setDecodedActions(value);
+      });
     }
   }, [
     apolloClient,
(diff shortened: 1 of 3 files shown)


RakeshUP (approved):

packages/web-app/src/pages/proposal.tsx
```diff
@@ -168,9 +168,10 @@ const Proposal: React.FC = () => {
                 network
               );
             case 'mint':
-              mintTokenActions.actions.push(action.data);
-              if (mintTokenActions.actions.length === 0)
+              if (mintTokenActions.actions.length === 0) {
                 mintTokenActions.index = index;
```
jamesej, line 172: On line 202, below, the splice statement will not in general delete only minting actions from the `actionPromises` list. This is because after this code is run mapping over `proposal.actions`, the `mintTokenActions.index` will simply point to the first minting action in the list. But the other minting actions could be located at any positions in the list after this so a single splice won't be able to delete all of them and would likely delete other actions.

jamesej (changes requested):
One specific comment added, but this is tricky because you need a general solution that will deal with any set of proposal actions. One way I could see to do it is to do a for loop over `proposal.actions` and push all the non-minting actions onto one list and the minting actions onto another, while recording the position of the first minting action. Then you combine the minting actions into one action and insert this into the non-minting actions list.

jamesej (approved):
All good, thanks

PR #616, opened 2023-01-06
Feature: DAO logo upload

RakeshUP:
Task: [APP-1419](https://aragonassociation.atlassian.net/browse/APP-1419)

## Type of change

## Checklist:

[APP-1419]: https://aragonassociation.atlassian.net/browse/APP-1419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui-components/src/components/avatar/avatarDao.tsx
+++ b/packages/ui-components/src/components/avatar/avatarDao.tsx
@@ -1,5 +1,6 @@
-import React, {HTMLAttributes, useMemo, useState} from 'react';
+import React, {HTMLAttributes, useEffect, useMemo, useState} from 'react';
 import styled from 'styled-components';
+import {resolveIpfsCid} from '@aragon/sdk-common';
 
 export interface AvatarDaoProps extends HTMLAttributes<HTMLElement> {
   daoName: string;
@@ -16,6 +17,21 @@ export const AvatarDao: React.FC<AvatarDaoProps> = ({
   ...props
 }) => {
   const [error, setError] = useState(false);
+  const [logoSrc, setLogoSrc] = useState<string | undefined>();
+
+  useEffect(() => {
+    setError(false);
+    if (src) {
+      try {
+        const logoCid = resolveIpfsCid(src);
+        setLogoSrc(`https://ipfs.io/ipfs/${logoCid}`);
+      } catch (err) {
+        setLogoSrc(undefined);
+      }
+    } else {
+      setLogoSrc(undefined);
+    }
+  }, [src]);
 
   const daoInitials = useMemo(() => {
     const arr = daoName.trim().split(' ');
@@ -23,13 +39,13 @@ export const AvatarDao: React.FC<AvatarDaoProps> = ({
     else return arr[0][0] + arr[1][0];
   }, [daoName]);
 
-  return error || !src ? (
+  return error || !logoSrc ? (
     <FallBackAvatar onClick={onClick} size={size} {...props}>
       <DaoInitials>{daoInitials?.toUpperCase()}</DaoInitials>
     </FallBackAvatar>
   ) : (
     <Avatar
-      src={src}
+      src={logoSrc}
       size={size}
       alt="dao avatar"
       onClick={onClick}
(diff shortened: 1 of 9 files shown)


packages/web-app/src/context/createDao.tsx
```diff
@@ -239,10 +250,16 @@ const CreateDaoProvider: React.FC = ({children}) => {
     const metadata: IMetadata = {
       name: daoName,
       description: daoSummary,
-      avatar: daoLogo,
       links: links.filter(r => r.name && r.url),
     };
 
+    if (daoLogo) {
+      const daoLogoBuffer = await readFile(daoLogo as Blob);
```
WissenIstNacht: Could you surround that with a try/catch block? Otherwise the third await might lead to awkward things if one of the two awaits before fail.

packages/web-app/src/containers/navbar/mobile.tsx
```diff
@@ -94,7 +94,7 @@ const FlexOne = styled.div.attrs({
 })``;
 
 const Container = styled.div.attrs({
-  className: 'flex flex-col fixed left-0 bottom-0 w-full',
+  className: 'flex flex-col fixed left-0 bottom-0 w-full z-50',
```
WissenIstNacht: Why though? :p

This change makes it so the navigation modal on mobile is behind the "footer bar". Meaning you can't click on settings.

WissenIstNacht (changes requested):
Thanks man, nice to finally have some logos! :)

I have an additional comment that I didn't really know where to put in code. When opening a DAO for the first time after creating it, the avatar shows something weird. Instead of showing the logo or the initials, it shows a white background with a string saying something like "Avatar Dao". Is that because the view still relies on cached data...? If it's a simple fix, could you change it?

packages/ui-components/src/components/avatar/avatarDao.tsx
```diff
@@ -16,20 +17,35 @@ export const AvatarDao: React.FC<AvatarDaoProps> = ({
...
 }) => {
   const [error, setError] = useState(false);
+  const [logoSrc, setLogoSrc] = useState<string | undefined>();
+
+  useEffect(() => {
+    setError(false);
+    if (src) {
+      try {
+        const logoCid = resolveIpfsCid(src);
+        setLogoSrc(`https://ipfs.io/ipfs/${logoCid}`);
```
Fabricevladimir: I don't know exactly where to put this, but keep in mind that the logo can be elsewhere, especially considering how SDK is allowing storage on different protocols. It might be best to resolve/transform the `src` in the parent component, or even better in our case maybe `useDaoDetails`. That way this component is presentational only and doesn't have to worry about the different protocols. For now we can keep this, but perhaps add a todo or create a tech debt ticket for it.

Fabricevladimir (approved):
Thanks. Approving so you can merge once you've resolved @WissenIstNacht's comments 🚀

packages/web-app/src/containers/navbar/mobile.tsx
RakeshUP: Changed it from `z-50` to `z-10`. This is because some images on the dashboard were on top of the footer bar.

packages/ui-components/src/components/avatar/avatarDao.tsx
RakeshUP: Added this logic to `useDaoDetails` and `useDaos` and removed it from `avatarDao`.

RakeshUP:
@WissenIstNacht, I am not able to reproduce the white background issue that you mentioned. The image is cached as an ArrayBuffer and used from the cache, but for me the actual image is rendered without any issues

packages/web-app/src/hooks/useDaoDetails.tsx
```diff
@@ -45,6 +46,16 @@ export function useDaoDetails(
...
           const dao = await client?.methods.getDao(daoId.toLowerCase());
           if (dao) {
+            if (dao.metadata.avatar) {
+              try {
+                const logoCid = resolveIpfsCid(dao.metadata.avatar);
+                dao.metadata.avatar = `https://ipfs.io/ipfs/${logoCid}`;
+              } catch (err) {
+                dao.metadata.avatar = undefined;
+              }
+            } else {
```
Fabricevladimir: Isn't this else case unnecessary since `metadata.avatar` is undefined in the first place

WissenIstNacht (approved):

PR #621, opened 2023-01-09
Feature: Add newly created DAO to favorites

RakeshUP:
Task: [APP-517](https://aragonassociation.atlassian.net/browse/APP-517)

## Type of change

## Checklist:

[APP-517]: https://aragonassociation.atlassian.net/browse/APP-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/context/apolloClient.tsx
+++ b/packages/web-app/src/context/apolloClient.tsx
@@ -12,6 +12,8 @@ import {
   Erc20Proposal,
   ICreateParams,
   IMetadata,
+  InstalledPluginListItem,
+  IPluginInstallItem,
 } from '@aragon/sdk-client';
 import {RestLink} from 'apollo-link-rest';
 import {CachePersistor, LocalStorageWrapper} from 'apollo3-cache-persist';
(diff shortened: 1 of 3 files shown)


packages/web-app/src/context/createDao.tsx
```diff
@@ -49,12 +59,13 @@ const CreateDaoContext = createContext<CreateDaoContextType | null>(null);
 const CreateDaoProvider: React.FC = ({children}) => {
   const {open} = useGlobalModalContext();
   const navigate = useNavigate();
-  const {isOnWrongNetwork, provider} = useWallet();
+  const {isOnWrongNetwork, provider, chainId} = useWallet();
```
Fabricevladimir: Can you get the network using `CHAIN_METADATA` instead? While the wallet _should_ technically be on the correct chain since it's right after a transaction, we don't have 100% guarantee that the user won't disconnect or do something random

packages/web-app/src/context/apolloClient.tsx
```diff
@@ -127,13 +129,14 @@ const client: Record<
...
 // showing these daos on explorer page
-export type NavigationDao = Omit<DaoListItem, 'metadata'> & {
+export type NavigationDao = Omit<DaoListItem, 'metadata' | 'plugins'> & {
   chain: SupportedChainID;
   metadata: {
     name: string;
     avatar?: string;
     description?: string;
   };
+  plugins: InstalledPluginListItem[] | IPluginInstallItem[];
```
Fabricevladimir, line 140: I'm curious about this type, `IPluginInstallItem`. I assume it's because the the other type is only returned after the plugin has been installed? It's not an issue considering the only thing I need this for is the plugin Id (for now at least), and both types come with it. If we need the rest of `InstalledPluginListItem` props, we can actually favorite the DAO on the Dashboard itself once the subgraph has returned it.

Fabricevladimir (commented):
Thanks, left a couple of comments

packages/web-app/src/context/apolloClient.tsx
RakeshUP: It's just to remove lint errors. I get a few errors in places where `selectedDaoVar` is used.

Fabricevladimir (approved):

sepehr2github (approved):

PR #646, opened 2023-01-23
APP-1623 - Feat: Adapt Settings Flow

WissenIstNacht:
## Description

Updates the settings flow with Majority Voting. I.e., it allows users to create proposal that sets the new parameters provided by Majority Voting.

NOTE: This does currently not involve the proposal threshold parameter. Pending a discussion with the design, this will be tackled in a separate ticket (TBD).

Fixes the bug where the links in the settings flow didn't show and didn't trigger the `settingsChanged` flag that allowed the user to propose settings.

Task: [1623](https://aragonassociation.atlassian.net/browse/APP-1623) & [1616](https://aragonassociation.atlassian.net/browse/APP-1616)

## Type of change

### DIFF
--- a/packages/web-app/src/containers/compareSettings/index.tsx
+++ b/packages/web-app/src/containers/compareSettings/index.tsx
@@ -1,3 +1,4 @@
+import {VotingMode} from '@aragon/sdk-client';
 import {
   AvatarDao,
   ButtonGroup,
@@ -14,8 +15,10 @@ import {Loading} from 'components/temporary';
 import {useNetwork} from 'context/network';
 import {useDaoDetails} from 'hooks/useDaoDetails';
 import {useDaoParam} from 'hooks/useDaoParam';
+import {useDaoToken} from 'hooks/useDaoToken';
 import {PluginTypes} from 'hooks/usePluginClient';
 import {usePluginSettings} from 'hooks/usePluginSettings';
+import {useTokenSupply} from 'hooks/useTokenSupply';
 import {getDHMFromSeconds} from 'utils/date';
 import {EditSettings} from 'utils/paths';
 import {ProposalResource} from 'utils/types';
@@ -30,10 +33,15 @@ const CompareSettings: React.FC = () => {
   const {data: daoDetails, isLoading: areDetailsLoading} = useDaoDetails(
     daoId!
   );
-
   const {data: daoSettings, isLoading: areSettingsLoading} = usePluginSettings(
     daoDetails?.plugins[0].instanceAddress as string,
     daoDetails?.plugins[0].id as PluginTypes
+  );
+  const {data: daoToken, isLoading: tokensAreLoading} = useDaoToken(
+    daoDetails?.plugins?.[0]?.instanceAddress || ''
+  );
+  const {data: tokenSupply, isLoading: tokenSupplyIsLoading} = useTokenSupply(
+    daoToken?.address || ''
   );
 
   const [selectedButton, setSelectedButton] = useState<'new' | 'old'>('new');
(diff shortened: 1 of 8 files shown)


packages/web-app/src/hooks/useTokenSupply.tsx
```diff
@@ -0,0 +1,35 @@
+import {formatUnits} from 'ethers/lib/utils';
+import {useEffect, useState} from 'react';
+
+import {useNetwork} from 'context/network';
+import {useProviders} from 'context/providers';
+import {CHAIN_METADATA} from 'utils/constants';
+import {getTokenInfo} from 'utils/tokens';
+import {HookData} from 'utils/types';
+
+export function useTokenSupply(
```
Fabricevladimir, line 10: What? You've an issue with all the `getTokenInfo` calls all over the place? 😉

packages/web-app/src/containers/compareSettings/index.tsx
```diff
@@ -54,23 +68,47 @@ const CompareSettings: React.FC = () => {
       links: getValues('daoLinks').filter(
```
Fabricevladimir: Why not pull out all the values at once instead of calling `getValues` multiple times? Also, can you add the Avatar as well. Currently we're only passing the DAO name

Fabricevladimir (commented):
Left a couple of comments. Also, lint is failing 😢

packages/web-app/src/pages/proposeSettings.tsx
```diff
@@ -122,30 +118,28 @@ const ProposeSettingWrapper: React.FC<Props> = ({
...
     data: {members},
   } = useDaoMembers(pluginAddress, pluginType as PluginTypes);
-
   const {data: pluginSettings} = usePluginSettings(
     pluginAddress,
     pluginType as PluginTypes
   );
   const {data: daoToken} = useDaoToken(pluginAddress);
-
+  const {data: tokenSupply, isLoading: tokenSupplyIsLoading} = useTokenSupply(
```
Fabricevladimir, line 140: I think you have this on the form already as `tokenTotalSupply`. It was set in Edit Settings.

packages/web-app/src/containers/compareSettings/index.tsx
WissenIstNacht: The avatar component doesn't like the uploaded avatar. For some reason. Maybe because it's a blob or something?

I've created a task to tackle this separately: https://aragonassociation.atlassian.net/browse/APP-1664

Fabricevladimir (approved):
Looking good!

packages/web-app/src/pages/proposeSettings.tsx
```diff
@@ -196,6 +170,8 @@ const ProposeSettingWrapper: React.FC<Props> = ({
         links,
```
Fabricevladimir: Just caught this when looking at the action on staging. Can you filter this array for empty name and url before passing it into the actions?

Edit: We were passing in the wrong links here as well. Should be `daoLinks` instead.

RakeshUP (approved):

packages/web-app/src/pages/proposeSettings.tsx
WissenIstNacht: No, this is the right one. A proposal can have its own set of links, which are not the same as the `daoLinks` that belong to the DAO's metadata

Also, empty links and names should be filtered out by the form during validation.

Fabricevladimir: Yes, a proposal can have its own set of links, but this is the DAO metadata action being encoded. The proposal links are passed down in `getProposalCreationParams`

Fabricevladimir: Concerning the links, the form allows you to go through with empty rows because they are optional I believe. That's why we filter empty ones out from the proposal links already in `getProposalCreationParams`.

WissenIstNacht: My bad, I jumped to the wrong method 👍🏼

Fabricevladimir: Exactly 😄, and I'm saying that's what you should be pulling from that specific `getValues` call because you're encoding the links for the Update DAO Metadata Action

PR #649, opened 2023-01-24
Feature: APP-1568 - Proposal Review (Settings)

Fabricevladimir:
## Description
- Adds the Update Governance and Metadata Setting cards on Proposal creation Review step.

Task: [APP-1568](https://aragonassociation.atlassian.net/browse/APP-1658)

## Type of change

## Checklist:

[APP-1568]: https://aragonassociation.atlassian.net/browse/APP-1568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/hooks/useTokenSupply.tsx
+++ b/packages/web-app/src/hooks/useTokenSupply.tsx
@@ -7,13 +7,18 @@ import {CHAIN_METADATA} from 'utils/constants';
 import {getTokenInfo} from 'utils/tokens';
 import {HookData} from 'utils/types';
 
+type TokenSupplyData = {
+  formatted: number;
+  raw: bigint;
+};
+
 export function useTokenSupply(
   tokenAddress: string
-): HookData<number | undefined> {
+): HookData<TokenSupplyData | undefined> {
   const {network} = useNetwork();
   const {infura} = useProviders();
 
-  const [data, setData] = useState<number>();
+  const [data, setData] = useState<TokenSupplyData>();
   const [error, setError] = useState<Error>();
   const [isLoading, setIsLoading] = useState(false);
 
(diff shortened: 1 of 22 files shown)


sepehr2github (approved):
Thanks, Quick question, On the setting page when I changed early execution the button won't get enabled and the change doesn't detect, Is it in the task area?

Fabricevladimir:
> Thanks, Quick question, On the setting page when I changed early execution the button won't get enabled and the change doesn't detect, Is it in the task area?

You are indeed correct,  @sepehr2github. I missed that.

packages/web-app/src/hooks/useTokenSupply.tsx
```diff
@@ -7,24 +7,30 @@ import {CHAIN_METADATA} from 'utils/constants';
 import {getTokenInfo} from 'utils/tokens';
 import {HookData} from 'utils/types';
 
+type TokenSupplyData = {
+  formatted: number;
+  raw: bigint;
+};
+
 export function useTokenSupply(
```
WissenIstNacht, line 15: Just as a side note: Never got to discuss this with you, but I thougth this might actually make more sense as a part of the daoToken hook. I mean you always need the daoTokens if you need the supply. That's not quite true the other way round (you may only need the token, without total supply). But even then it could be hidden behind a flag and fetched only when desired?

Didn't want to cause a huge refactor, just as an idea for the future.

WissenIstNacht (commented):
All seems good, thanks :)

WissenIstNacht (approved):
All seems good, thanks :)

packages/web-app/src/hooks/useTokenSupply.tsx
Fabricevladimir: No, that makes total sense. I've thought about it before. I'll try to see if I can fit it in somewhere since this one is getting bigger and bigger

PR #663, opened 2023-01-30
Feature: APP-1540 - Multisig - Proposal Duration Warnings

Fabricevladimir:
## Description

- Adds warnings for minimum duration 

Task: [APP-1540](https://aragonassociation.atlassian.net/browse/APP-1540)

## Type of change

## Checklist:

[APP-1540]: https://aragonassociation.atlassian.net/browse/APP-1540?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jamesej (changes requested):
Hi if you choose Specific Date & Time for Start Time, the time field doesn't update properly when you try to edit it. I'd guess the loop between the update event and the rerendering has issues.

packages/web-app/src/containers/dateTimeSelector/index.tsx
```diff
@@ -1,164 +1,37 @@
...
-    dateTimeValidator();
-  }, [utcEnd, dateTimeValidator]); //eslint-disable-line
+const DateTimeSelector: React.FC<Props> = ({
+  mode,
+  value,
+  onUtcClicked,
+  validator,
+  defaultDateOffset,
+}) => {
+  const {days, hours, minutes} = {...defaultOffsets, ...defaultDateOffset};
```
WissenIstNacht, line 40: Since these are a type `Offset` and the only place your using them is as arguments of type `Offset`, you might get by without destructuring the object. But no biggie :)

packages/web-app/src/containers/dateTimeSelector/useDateTimeSelector.ts
```diff
@@ -0,0 +1,241 @@
+// NOTE: Keeping this in the dateTimeSelector folder despite
```
WissenIstNacht: I really wouldn't. The only thing that's really going to change if you put this here is that other people will put other hooks wherever. It very much doubt that it will prevent people from forgetting this hook when using the `DateTimeSelector` as a) they will just copy existing code in the first place and b) mess it up anyway, because people just mess up lol.

Also, the file really should have a tsx ending.

packages/web-app/src/containers/setupVotingForm/tokenVoting.tsx
```diff
@@ -214,240 +111,98 @@ const SetupTokenVotingForm: React.FC<Props> = ({pluginSettings}) => {
...
+                  minDuration={{days, hours, minutes}}
                 />
-              </HStack>
-              {formState.errors?.duration?.message && (
-                <AlertInline
-                  label={formState.errors.duration.message}
-                  mode="critical"
+              ) : (
+                <>
+                  <DateTimeSelector
```
WissenIstNacht: Are you intentionally nesting this inside the `Controller` with `name="durationSwitch`? I mean, the `Duration` and `DateTimeSelector` come with their own `Controllers` already. I'm not sure if or what the implications are of nesting controllers, but it seems wrong somehow.

Edit: Same question for `Duration`, obviously.

packages/web-app/src/containers/dateTimeSelector/useDateTimeSelector.ts
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * This is a hook that returns a validator, handlers and utilities
+ * for working with the DateTimeSelector component
+ * @param minDurationMills minimum duration as defined by the DAO settings
+ * @param minDurationAlert alert shown when the current duration is less than
+ *  the minimum duration defined by the DAO settings
+ * @returns
+ */
+export const useDateTimeSelectorValidation = (
```
WissenIstNacht: I'm struggling a bit with this one. It seems to me like all this code could just be inside the `DateTimeSelector`, no? (Obviously you'd have to pass some getters/setters between parent and child). It seems so tightly linked to the `DateTimeSelector` component that I don't really see the benefit of extracting it into a hook. I mean, you said it yourself, the component and this hook are inseparable.

WissenIstNacht (changes requested):
Thanks a lot Fabrice! Lots of good things in here. I've left some comments on structure. 

Additionally: 
a) I'm not sure where the issue comes from, but your enddates do not consider the start offset. E.g., if the current time is 10AM, the start time will be 10:10 AM but the end date will be 11am instead of 11:10AM. This is visible in the last proposal review when you just step through the flow normally.
b) I think landing on the vote setup screen has reset the times pretty much every time I did it. I don't think that should be the case.

packages/web-app/src/containers/dateTimeSelector/index.tsx
Fabricevladimir: Saves me keystrokes to destructure them 😄

packages/web-app/src/containers/dateTimeSelector/useDateTimeSelector.ts
Fabricevladimir: Got feedback on this as well from James. I can get away with the validator in `DateTimeSelector`, but the other methods, not so much, especially since most are targeting the togglers. I'll take those out and essentially copy them over in both TokenVoting & Multisig forms

packages/web-app/src/containers/setupVotingForm/tokenVoting.tsx
Fabricevladimir: This shouldn't have any effect. Saves from having to watch both toggle values.

jamesej (approved):
OK All good now thanks. Issue with timezones jumping to GMT+0 on dateTimeSelector being corrected to now from time in past now is in another ticket.

WissenIstNacht (approved):
LGTM!

PR #664, opened 2023-01-30
Feat: APP-1694 Minimum Proposal Threshold settings.

WissenIstNacht:
## Description

Allow to change the minimum proposal threshold parameter as part of the normal settings flow for MV DAOs.

Task: [1694](https://aragonassociation.atlassian.net/browse/APP-1694)

## Type of change

### DIFF
--- a/packages/web-app/src/pages/editSettings.tsx
+++ b/packages/web-app/src/pages/editSettings.tsx
@@ -227,7 +227,7 @@ const EditSettings: React.FC = () => {
     setValue('eligibilityTokenAmount', formattedProposerAmount);
     setValue('minimumTokenAmount', formattedProposerAmount);
     setValue('eligibilityType', formattedEligibilityType);
-  }, [daoSettings.minProposerVotingPower, setValue]);
+  }, [formattedEligibilityType, formattedProposerAmount, setValue]);
 
   const setCurrentGovernance = useCallback(() => {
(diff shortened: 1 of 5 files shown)


packages/web-app/src/pages/editSettings.tsx
```diff
@@ -322,6 +361,20 @@ const EditSettings: React.FC = () => {
                 </AccordionContent>
               </AccordionItem>
 
+              <AccordionItem
```
Fabricevladimir, line 364: It's slightly confusing having this as a community change when it appears under governance settings in the actions & action cards, but I'm guessing that's a design thing.

Fabricevladimir (approved):
Looks good 👍🏾

packages/web-app/src/pages/editSettings.tsx
WissenIstNacht: Horrible. But got it confirmed by Selim. The idea being that it needs to align with the DAO creation and the settings views.

jamesej (changes requested):
I came across a minor display issue in that the New Settings display has a decimal '10.0' and the Old Settings display has none
ProposeSettingsOldSettingsProposalThreshold
ProposeSettingsNewSettingsProposalThreshold
 '10'. It might be that the actual new value has a decimal rounding issue and is very slightly larger than 10.

jamesej (approved):
Thanks Vali all good now

PR #665, opened 2023-01-31
fix: App-1684

sepehr2github:
## Description

-revote button will be disabled when the proposal got rejected

Task: [App-1684](https://aragonassociation.atlassian.net/browse/APP-1684)

## Type of change

## Checklist:

### DIFF
--- a/packages/web-app/src/pages/proposal.tsx
+++ b/packages/web-app/src/pages/proposal.tsx
@@ -3,6 +3,7 @@ import {
   DaoAction,
   TokenVotingClient,
   TokenVotingProposal,
+  VotingMode,
   VotingSettings,
 } from '@aragon/sdk-client';
 import {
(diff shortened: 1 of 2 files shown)


RakeshUP (approved):
The PR is still in draft!

packages/web-app/src/pages/proposal.tsx
```diff
@@ -376,13 +378,20 @@ const Proposal: React.FC = () => {
     }
 
     // member, not yet voted
-    else if (canVote) {
+    else if (canVote && !earlyExecution) {
```
WissenIstNacht: This is not gonna work :( 
  
Because if your DAO has EE enabled. this will always evaluate to false, thereby ignoring `canVote`.

// Edit: Note that simply removing the negation will not make it work either. Because than you'd only be able to vote in a DAO with EE.

WissenIstNacht (changes requested):
Hey there, thanks for the PR. Unfortunately, that the doesn't quite fix the error 😅

packages/web-app/src/pages/proposal.tsx
sepehr2github: > This is not gonna work :(
> 
> Because if your DAO has EE enabled. this will always evaluate to false, thereby ignoring `canVote`.
> 
> // Edit: Note that simply removing the negation will not make it work either. Because than you'd only be able to vote in a DAO with EE.

Thanks for the review, Conditions changed to fix the issue

WissenIstNacht (approved):
Hey Sepehr, thanks for the changes. The button seems to work.

PR #669, opened 2023-02-01
Feature: Create multisig dao settings page

josemarinas:
## Description

Task: [APP-1704](https://aragonassociation.atlassian.net/browse/APP-1704)

## Type of change

## Checklist:

[APP-1704]: https://aragonassociation.atlassian.net/browse/APP-1704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/web-app/src/components/multisigMinimumApproval/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const {t} = useTranslation();
+  const {control, setValue} = useFormContext();
+  const [multisigWallets, multisigMinimumApprovals] = useWatch({
+    name: ['multisigWallets', 'multisigMinimumApprovals'],
+    control: control,
+  });
+
+  useEffect(() => {
+    if (multisigMinimumApprovals > multisigWallets.length) {
+      setValue('multisigMinimumApprovals', multisigWallets.length);
```
sepehr2github: In the Controller component, there is a maximum prop that you can set to handle the limitation you need but this method is correct as well ;D

packages/web-app/src/components/multisigMinimumApproval/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <Label
+        label={t('labels.minimumApproval')}
+        helpText={t('createDAO.step4.minimumApprovalSubtitle')}
+      />
+      <Controller
+        name="multisigMinimumApprovals"
+        control={control}
+        defaultValue={Math.ceil(multisigWallets.length / 2)}
+        rules={{
+          validate: value => validateminimumApprovals(value),
```
sepehr2github: It's better to return a proper message that will consider an error and show at as inline Alert

packages/web-app/src/components/multisigMinimumApproval/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+                  placeholder={t('placeHolders.daoName')}
+                />
+              </div>
+
+              <div className="flex flex-1 items-center">
+                <LinearProgressContainer>
+                  <LinearProgress max={multisigWallets.length} value={value} />
+                  <ProgressBarTick />
+                  <ProgressInfo>
+                    {multisigMinimumApprovals !== multisigWallets.length && (
```
sepehr2github: This would be an improvement and nothing more but these two conditions could be merged together!

sepehr2github (commented):
Thanks, Jose! Seems good but some minor changes needed

packages/web-app/src/components/multisigMinimumApproval/index.tsx
josemarinas: oh I didnt know I'll do it that way

packages/web-app/src/components/multisigMinimumApproval/index.tsx
josemarinas: 👍

packages/web-app/src/components/multisigMinimumApproval/index.tsx
josemarinas: I believe this is already fixed with the min max in the input so i dont see use for the validation here

packages/web-app/src/components/multisigMinimumApproval/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <>
+      <Label
+        label={t('labels.minimumApproval')}
+        helpText={t('createDAO.step4.minimumApprovalSubtitle')}
+      />
+      <Controller
+        name="multisigMinimumApprovals"
+        control={control}
+        defaultValue={Math.ceil(multisigWallets.length / 2)}
```
Fabricevladimir: The math is slightly off here because for every even number, this will return half of that number. For ex: `Math.ceil(4 / 2)` will return 2 instead of 3 (which is the correct number since the majority is defined as >50%)

packages/web-app/src/components/multisigMinimumApproval/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                        }}
+                      >
+                        {value}
+                      </p>
+                    ) : (
+                      <p className="font-bold text-right text-primary-500">
+                        {value}
+                      </p>
+                    )}
+                    <p className="text-ui-600 ft-text-sm">
```
Fabricevladimir, line 93: When we hit 100% approval, the label `Addresses` drops down into the linear progress bar. 

Also, there shouldn't be a grey bar inside the linear progress. I think it's for MJ Voting support threshold only.


image

Fabricevladimir (changes requested):
Thanks! Left a couple of comments.

packages/web-app/src/components/multisigMinimumApproval/index.tsx
```diff
@@ -16,7 +16,13 @@ export const MultisigMinimumApproval = () => {
     name: ['multisigWallets', 'multisigMinimumApprovals'],
     control: control,
   });
-
+  const computeDefaultValue = () => {
+    const ceiledApprovals = Math.ceil(multisigWallets.length / 2);
+    if (ceiledApprovals % 2) {
```
Fabricevladimir: ~~I think you need to check if the wallet length is odd or even first. If it's even, divide it by 2 and add one.~~
~~If it's odd, ceil the the value after dividing by two.~~ 

As it stands currently, if we have 1 wallet for example, `ceiledApprovals` will be 1, and because `ceiledApprovals % 2` will return 1 (a truthy value), we will add 1 to `ceiledApprovals` making the final value 2.

EDIT: I just realized you're doing the same thing since you're adding to what was previously odd. 1 is just a special case since it's odd, but adding 1 to it breaks your DAO

josemarinas: I think the condition is reversed, when its even I should add 1, when its odd I shouldnt

Fabricevladimir (approved):

sepehr2github (approved):
Thanks For the changes!

Fabricevladimir (approved):

PR #678, opened 2023-02-04
Feature: APP-1542 - Multisig DAO - Proposal Gating

Fabricevladimir:
## Description

- Gates proposal creation based on membership and DAO settings 

Task: [APP-1542](https://aragonassociation.atlassian.net/browse/APP-1542)

## Type of change

## Checklist:

[APP-1542]: https://aragonassociation.atlassian.net/browse/APP-1542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/components/protectedRoute/index.tsx
+++ b/packages/web-app/src/components/protectedRoute/index.tsx
@@ -74,12 +74,17 @@ const ProtectedRoute: React.FC = () => {
   ]);
 
   const gateMultisigProposal = useCallback(() => {
-    if ((daoSettings as MultisigVotingSettings).onlyListed === false)
+    if ((daoSettings as MultisigVotingSettings).onlyListed === false) {
       close('gating');
-
-    if (filteredMembers.length === 0 && !membersAreLoading) open('gating');
-    else close('gating');
-  }, [membersAreLoading, close, daoSettings, filteredMembers.length, open]);
+    } else if (
+      !filteredMembers.some(mem => mem.address === address) &&
+      !membersAreLoading
+    ) {
+      open('gating');
+    } else {
+      close('gating');
+    }
+  }, [membersAreLoading, close, daoSettings, open, address, filteredMembers]);
 
   /*************************************************
    *                     Effects                   *
--- a/packages/web-app/src/components/protectedRoute/index.tsx
+++ b/packages/web-app/src/components/protectedRoute/index.tsx
@@ -1,4 +1,4 @@
-import {VotingSettings} from '@aragon/sdk-client';
+import {MultisigVotingSettings, VotingSettings} from '@aragon/sdk-client';
 import React, {useCallback, useEffect, useRef} from 'react';
 import {Outlet} from 'react-router-dom';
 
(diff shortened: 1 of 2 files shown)


packages/web-app/src/components/protectedRoute/index.tsx
```diff
@@ -67,20 +65,25 @@ const ProtectedRoute: React.FC = () => {
...
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const checkIfAllowlistedMember = useCallback(() => {
-    if (filteredMembers.length === 0 && !MembershipIsLoading) open('gating');
+  const gateMultisigProposal = useCallback(() => {
+    if ((daoSettings as MultisigVotingSettings).onlyListed === false)
+      close('gating');
+
+    if (filteredMembers.length === 0 && !membersAreLoading) open('gating');
     else close('gating');
```
jamesej: The test here isn't right, it shouldn't be checking if filteredMembers.length is zero but rather if the current user is in that list. I wrote some code that seems to fix it:

```
  const gateMultisigProposal = useCallback(() => {
    if ((daoSettings as MultisigVotingSettings).onlyListed === false) {
      close('gating');
    } else if (
      !filteredMembers.some(mem => mem.address === address) &&
      !membersAreLoading
    ) {
      open('gating');
    } else {
      close('gating');
    }
  }, [membersAreLoading, close, daoSettings, open, address, filteredMembers]);
```

jamesej (changes requested):
Isn't working correctly as it stands, see my comment in files.

packages/web-app/src/components/protectedRoute/index.tsx
Fabricevladimir: Wait, but doesn't `filteredMembers` returning 0 signify that the user isn't in the list of members anyway?🤔

jamesej: Lol OK I didn't check what useDaoMembers actually did, and see that it is filtering by address. I checked again - there is an issue in that you get gated out in a DAO where anyone can make proposals, because in the current code the condition at line 77 doesn't return from the function and reopens the gating at line 80. So actually it's the `else if` you need.

jamesej: Perfect example for the doubters of why loads of if/thens in your code is a bad idea! :wink:

packages/web-app/src/components/protectedRoute/index.tsx
```diff
@@ -102,32 +110,41 @@ const ProtectedRoute: React.FC = () => {
...
-    MembershipIsLoading ||
+    membersAreLoading ||
     settingsAreLoading
   )
     return <Loading />;
 
   return (
     <>
-      <Outlet />
+      {!isGatingOpen && <Outlet />}
```
WissenIstNacht, line 147: A part of me really like this, a part really doesn't haha. No need to change though

WissenIstNacht (approved):
Works for me, thanks.

RakeshUP (approved):

packages/web-app/src/components/protectedRoute/index.tsx
Fabricevladimir: Hahaha! It annoyed me a bit because the scroll bar would suddenly pop up when the proposal creation page actually loaded in the background so I was like, you know what go away!

jamesej (approved):
All good thanks

PR #682, opened 2023-02-07
App 1719 multisig proposal card

WissenIstNacht:
## Description

Display Multisig DAO proposal as Proposal Cards in the Dashboard and Governance view.

Task: [APP-1719](https://aragonassociation.atlassian.net/browse/APP-1719)

Note: Some data is still hard-coded because the actual data is not available from the SDK.

[APP-1719]: https://aragonassociation.atlassian.net/browse/APP-1719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/components/proposalList/index.tsx
+++ b/packages/web-app/src/components/proposalList/index.tsx
@@ -1,7 +1,7 @@
 import {CardProposal, CardProposalProps, Spinner} from '@aragon/ui-components';
-import React from 'react';
+import {React, useMemo} from 'react';
 import {useTranslation} from 'react-i18next';
-import {useNavigate, useParams} from 'react-router-dom';
+import {NavigateFunction, useNavigate} from 'react-router-dom';
 
 import {useNetwork} from 'context/network';
 import {trackEvent} from 'services/analytics';
@@ -12,7 +12,7 @@ import {
 } from 'utils/constants';
 import {translateProposalDate} from 'utils/date';
 import {formatUnits} from 'utils/library';
-import {isErc20VotingProposal} from 'utils/proposals';
+import {isErc20VotingProposal, isMultisigProposal} from 'utils/proposals';
 import {abbreviateTokenAmount} from 'utils/tokens';
 import {ProposalListItem} from 'utils/types';
 import {i18n} from '../../../i18n.config';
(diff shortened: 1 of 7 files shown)


packages/web-app/src/components/proposalList/index.tsx
```diff
@@ -80,68 +99,94 @@ export type CardViewProposal = Omit<CardProposalProps, 'onClick'> & {
...
+    explorer: CHAIN_METADATA[network].explorer,
+    publisherAddress: proposal.creatorAddress,
+    publishLabel: i18n.t('governance.proposals.publishedBy'),
+    process: proposal.status.toLowerCase() as CardProposalProps['process'],
+    onClick: () => {
+      trackEvent('governance_viewProposal_clicked', {
+        proposal_id: proposal.id,
+        dao_address: proposal.dao,
+      });
+      navigate(`proposals/${proposal.id}`);
```
Fabricevladimir: On the Dashboard, clicking on a proposal will navigate to an invalid url. I remeber we have a `generatePath` function or something for similar occurrences.

packages/web-app/src/pages/governance.tsx
```diff
@@ -150,7 +150,12 @@ const Governance: React.FC = () => {
           </ButtonGroup>
         </ButtonGroupContainer>
         <ListWrapper>
-          <ProposalList proposals={displayedProposals} isLoading={isLoading} />
+          <ProposalList
+            proposals={displayedProposals}
+            pluginAddress={daoDetails?.address as string}
```
Fabricevladimir: This should be pluginInstanceAddress I think instead of DAO address.

packages/web-app/src/locales/en/common.json
```diff
@@ -245,6 +245,7 @@
       "Pending": "Starts in {{timeDiff}}",
       "Active": "{{timeDiff}} left",
       "voteTitle": "Winning option",
+      "voteTitleMultisig": "Approval Progress",
```
Fabricevladimir, line 248: Not a big deal, but seems copy has been updated to "Approved by". We have an existing key I think.

Fabricevladimir (commented):
Thanks, looking good!

Left a couple of comments. One other minor thing is that we're missing the approval indicator on the Linear Progress.

packages/web-app/src/components/proposalList/index.tsx
WissenIstNacht: Not sure why it doesn't like the relative path all of a sudden. It was working fine before, I just copied the code into the function. 🤔

packages/web-app/src/locales/en/common.json
WissenIstNacht: Ah nice, wasn't sure about adding keys. I'll do a quick update 👍🏼

Fabricevladimir (approved):

PR #697, opened 2023-02-10
Fix: Dao name validation on settings page

baxri:
## Description

On settings page, when editing, changing the DAO name field then manually reverting it to the original name wrongly results in a ‘Name is already taken’ error.

https://aragonassociation.atlassian.net/browse/APP-1771


sepehr2github (commented):
Thanks for the changes but still one issue left, When I use an invalid ens name such as "sepehr" which is already taken, the edit button is enabled which should be disabled if the name form returns an error

WissenIstNacht (approved):

sepehr2github (approved):

PR #704, opened 2023-02-13
Fix: calendar date input double icon

baxri:
## Description

When the Date Input field is displaying (e.g. on proposal creation when selecting ‘Specific Date + Time’), there is a double display of the calendar icon (Firefox issue only!).

[https://aragonassociation.atlassian.net/browse/APP-1750](https://aragonassociation.atlassian.net/browse/APP-1750)

### DIFF
--- a/packages/ui-components/src/components/input/dateInput.tsx
+++ b/packages/ui-components/src/components/input/dateInput.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {useRef} from 'react';
 import styled from 'styled-components';
 
 import {IconCalendar} from '../icons';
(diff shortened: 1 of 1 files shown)


packages/ui-components/src/components/input/dateInput.tsx
```diff
@@ -35,7 +49,8 @@ allows for hover and active when disabled. */
 type InputContainerProps = Pick<DateInputProps, 'disabled'>;
 
 const InputContainer = styled.div.attrs(({disabled}: InputContainerProps) => {
-  const baseClasses = 'flex items-center p-1 rounded-xl border-2 font-normal';
+  const baseClasses =
+    'flex items-center p-1 rounded-xl border-2 font-normal cursor-pointer';
   let className = `${baseClasses}`;
```
jamesej, line 25: Input container needs to have the class for 'position:relative' added because the Overlay which has position:absolute won't use the input container as the base for it's positioning otherwise. Maybe you had this set up as a manual style and forgot to add it. Also the horizontal positioning of Overlay is wrong - maybe another manual style that didn't get added?

jamesej (changes requested):

WissenIstNacht (changes requested):
Hey there, thanks for the PR, cool stuff.

There is a small issue with the changes, though. As it stands, the date input now has two icons (a chevron down and the calendar icon) when using chrome on mobile.

If you can make that go away and keep the rest of changes, then that's fine with me. Otherwise, I would suggest just getting rid of the custom icon on Firefox. I.e., just delete the `{isFF && ...` section in the component.

WissenIstNacht:
Update: Hey, @baxri,  I didn't see @jamesej had reviewed the PR before me. Sorry, didn't want to step on anyone's toes :D

baxri:
> 

May be it it better to remove custom Icon for now, since there might be some other headaches, it works in a different in all browsers. What do you think? just to lease default icon and behavior for this date input? @jamesej  @WissenIstNacht ?

In future it can be done using some third party library to get same functionality on all browsers

WissenIstNacht (approved):
Hey @baxri, yes this is what I meant. Thanks a lot 😊 👍🏼

WissenIstNacht (approved):

jamesej (approved):
All good thanks

PR #708, opened 2023-02-14
Chore: APP-1748 Proposals improvements

baxri:
## Description

New proposals are added on the bottom of the ‘Proposal Cards list’ at the Governance Page (e.g. see image 1 below), instead of on top of the list. and general improvements.

https://aragonassociation.atlassian.net/browse/APP-1748

### DIFF
--- a/packages/web-app/src/pages/governance.tsx
+++ b/packages/web-app/src/pages/governance.tsx
@@ -16,7 +16,6 @@ import styled from 'styled-components';
 import ProposalList from 'components/proposalList';
 import {Loading} from 'components/temporary';
 import {PageWrapper} from 'components/wrappers';
-import {useDaoDetails} from 'hooks/useDaoDetails';
 import {useDaoParam} from 'hooks/useDaoParam';
 import {PluginTypes} from 'hooks/usePluginClient';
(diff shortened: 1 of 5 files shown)


packages/web-app/src/pages/governance.tsx
```diff
@@ -45,19 +42,24 @@ const Governance: React.FC = () => {
     filterValue !== 'All' ? filterValue : undefined
   );
 
+  // Show proposals loading when loading flag is true and list is empty
+  // Othervise it means there is a show more was clocked
```
sepehr2github: clicked? spelling issue

sepehr2github (changes requested):
Cool, Thanks George, But I found some points such as:
The loading button should be hidden when navigating through the proposal type filter, also filter doesn't work properly

baxri:
> Cool, Thanks George, But I found some points such as: The loading button should be hidden when navigating through the proposal type filter, also filter doesn't work properly

@sepehr2github  good catch, this is problem in current version also of the app. I will fix it.

sepehr2github (approved):
Thanks, Goerge. Much better now. LGTM 👍

PR #725, opened 2023-02-21
Feat: Live updates for proposals

baxri:


packages/web-app/src/pages/proposal.tsx
```diff
@@ -327,6 +336,29 @@ const Proposal: React.FC = () => {
...
+
+      const interval = setInterval(async () => {
+        // remove interval timer once the proposal has started
+        if (proposal.startDate.valueOf() >= new Date().valueOf()) {
+          clearInterval(interval);
+          setIntervalInMills(PROPOSAL_STATUS_INTERVAL);
+        } else if (proposal.status === 'Pending') {
+          // recalculate vote status for pending proposal
+          setVoteStatus(getVoteStatus(proposal, t));
+        }
```
WissenIstNacht, line 352: else? Or is the idea that every other case beside pending is covered in the first if clause? If so, why not simply go else here?

packages/web-app/src/pages/proposal.tsx
```diff
@@ -327,6 +336,29 @@ const Proposal: React.FC = () => {
     }
   }, [voteSubmitted]);
 
+  useEffect(() => {
+    if (proposal) {
```
WissenIstNacht, line 340: I don't quite understand why this needs an interval too. Iin my understanding (correct me if I'm wrong) you have one interval that refetches data and then one interval here that recomputes the status. But isn't it enough to simply have the one that refetches data? Wouldn't the status be recomputed as a consequence of the rerender triggered by the refetched data?

WissenIstNacht (commented):
Hey there, thanks for the PR :)

Things seem to work so far. However, I'm not quite sure I understand some parts of the code. Could you let me know your thoughts in the comments?

packages/web-app/src/pages/proposal.tsx
baxri: @WissenIstNacht  As I disscused @Fabricevladimir  First interval is to get and recalculate statuses and once status is changed then we have an ability into useDaoProposal to start another interval which will update data every 2 minute. useDaoProposal interval is kind of seperate features for proposal fetching we just pass parameter to it and it will do his work to update the proposal. First interval in a page is just for to recalculate status.

packages/web-app/src/pages/proposal.tsx
baxri: @Fabricevladimir  I think we are calculating new status on when status is Pending, other we do nothing right @Fabricevladimir . and when time is passed we quite the interval and starting new interval to update the proposals every 2 min.

WissenIstNacht (approved):

PR #738, opened 2023-02-23
Chore: APP-1885 SDK 0.22 integration

jamesej:
## Description

Integration of SDK 0.22. Changelog for this SDK at [https://github.com/aragon/sdk/blob/develop/modules/client/CHANGELOG.md#0220-beta](https://github.com/aragon/sdk/blob/develop/modules/client/CHANGELOG.md#0220-beta)

Task: [APP-1885](https://aragonassociation.atlassian.net/browse/APP-1885)

## Type of change

## Checklist:

[APP-1885]: https://aragonassociation.atlassian.net/browse/APP-1885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/context/createDao.tsx
+++ b/packages/web-app/src/context/createDao.tsx
@@ -172,8 +172,12 @@ const CreateDaoProvider: React.FC = ({children}) => {
     ];
   }, [getValues]);
 
-  const getVoteSettings = useCallback((): VotingSettings => {
+  const getVoteSettings = useCallback((): [
+    VotingSettings,
+    'goerli' | 'mainnet'
+  ] => {
     const {
+      blockchain,
       minimumApproval,
       minimumParticipation,
       durationDays,
@@ -193,20 +197,23 @@ const CreateDaoProvider: React.FC = ({children}) => {
     else if (earlyExecution) votingMode = VotingMode.EARLY_EXECUTION;
     else votingMode = VotingMode.STANDARD;
 
-    return {
-      minDuration: getSecondsFromDHM(
-        parseInt(durationDays),
-        parseInt(durationHours),
-        parseInt(durationMinutes)
-      ),
-      minParticipation: parseInt(minimumParticipation) / 100,
-      supportThreshold: parseInt(minimumApproval) / 100,
-      minProposerVotingPower:
-        eligibilityType === 'token'
-          ? parseUnits(eligibilityTokenAmount.toString(), 18).toBigInt()
-          : BigInt(0),
-      votingMode,
-    };
+    return [
+      {
+        minDuration: getSecondsFromDHM(
+          parseInt(durationDays),
+          parseInt(durationHours),
+          parseInt(durationMinutes)
+        ),
+        minParticipation: parseInt(minimumParticipation) / 100,
+        supportThreshold: parseInt(minimumApproval) / 100,
+        minProposerVotingPower:
+          eligibilityType === 'token'
+            ? parseUnits(eligibilityTokenAmount.toString(), 18).toBigInt()
+            : BigInt(0),
+        votingMode,
+      },
+      blockchain.network === 'test' ? 'goerli' : 'mainnet',
+    ];
   }, [getValues]);
 
   const getErc20PluginParams =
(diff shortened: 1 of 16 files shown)


packages/web-app/src/context/createDao.tsx
```diff
@@ -150,18 +150,27 @@ const CreateDaoProvider: React.FC = ({children}) => {
...
+      {
         members: multisigWallets.map(wallet => wallet.address),
         votingSettings: {
           minApprovals: multisigMinimumApprovals,
           onlyListed: eligibilityType === 'multisig',
         },
-      };
-    }, [getValues]);
+      },
+      blockchain.network === 'test' ? 'goerli' : 'mainnet',
```
WissenIstNacht, line 171: This might work for now, but will likely break in general. `blokchain.network` is basically a boolean variable that differentiates between test and main networks. There are other Testnets than `goerli` (think `mumbai`) and the same goes for Mainnet. Within the app we agreed to call the Ethereum Mainnet `ethereum` instead of `mainnet` for that exact reason. Or, even better, to use the `chainId` (as a number) over strings whenever possible. But apparently the SDK did not get that memo, seeing as your passing that to `getPluginInstallItem`...

packages/web-app/src/pages/proposeSettings.tsx
```diff
@@ -311,7 +311,7 @@ const ProposeSettingWrapper: React.FC<Props> = ({
     };
 
     const getProposalCreationParams =
-      async (): Promise<ICreateProposalParams> => {
+      async (): Promise<CreateMajorityVotingProposalParams> => {
```
WissenIstNacht, line 314: Will this work if it's a Multisig proposal...?

packages/web-app/src/hooks/useProposals.tsx
```diff
@@ -156,8 +156,7 @@ export function useProposals(
             // TODO: Fix - SDK is still returning approvals as number and not array
             augmentedProposals.unshift({
               ...augmentedProposal,
-              approvals: (augmentedProposal as MultisigProposal).approvals
-                .length,
+              approvals: (augmentedProposal as MultisigProposal).approvals,
```
WissenIstNacht: Does that fix the above TODO?

WissenIstNacht (commented):
Just left some comments

WissenIstNacht:
One more issue... Something happened with the `proposalId`. When clicking on a proposal from either the `dashboard` or the `governance` page (see `proposalList/index.tsx`), the proposals navigate to `not-found`. Apparently the SDK doesn't like the proposal id form or something.

packages/web-app/src/context/createDao.tsx
jamesej: Thanks for this - yes you're right but this will work for now, right? As and when we add another network we can generalise this.

packages/web-app/src/hooks/useProposals.tsx
jamesej: Yes indeed! will remove

Fabricevladimir: @WissenIstNacht, anyone else getting NaN on the proposal card for a multisig proposal?

WissenIstNacht: No that was working I think 🤔

Fabricevladimir: Yeah that was on me, had to delete node modules because SDK package wasn't actually getting updated

packages/web-app/src/pages/proposeSettings.tsx
Fabricevladimir: This works as I've successfully created a proposal, but it will most certainly die on us if the params differ in the slightest.

jamesej: If we need a type that handles params for any type of proposal creation, we should have our own and derive it from the SDK type(s) that way when they make changes we can fix them in one place

WissenIstNacht: If the SDK exports a type called `CreateMajorityVotingProposalParams` that not only work for Multisig proposals, but is also _intended_ to do so, something is wrong. Either rename the type, split it up, or factor out common things, etc. But you can't have a type called `CreateMajorityVotingProposalParams` that also does Multisig.

Aaaaanyway, I think we can all agree that this is something for future patches.

Fabricevladimir: Bro, I'm out of tears.

Fabricevladimir (approved):

PR #739, opened 2023-02-24
Chore: Remove description from deposit flow

baxri:
https://aragonassociation.atlassian.net/browse/APP-1810


WissenIstNacht (approved):
I think this is fine for now.

But this ticket will still need some rework after the launch, as there are still a couple of places in the code that rely on (reading) references.

PR #740, opened 2023-02-24
Fix: Alpha Launch chain Fix

sepehr2github:
## Description

- Fix search input clear action
- Fix dev Related copy issues
- Remove other chains except for goerli & mainnet from dao creation
- Add Network hook to useClient

Task: [APP-1857](https://aragonassociation.atlassian.net/browse/APP-1857) - [APP-1860](https://aragonassociation.atlassian.net/browse/APP-1860) - [APP-1813](https://aragonassociation.atlassian.net/browse/APP-1813)

## Type of change

## Checklist:

[APP-1857]: https://aragonassociation.atlassian.net/browse/APP-1857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APP-1860]: https://aragonassociation.atlassian.net/browse/APP-1860?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APP-1813]: https://aragonassociation.atlassian.net/browse/APP-1813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/containers/selectChainForm/index.tsx
+++ b/packages/web-app/src/containers/selectChainForm/index.tsx
@@ -28,18 +28,18 @@ const SelectChainForm: React.FC = () => {
   // const [sortFilter, setFilter] = useState<SortFilter>('cost');
   const [networkType, setNetworkType] = useState<NetworkType>('test');
 
-  // moving this up so state change triggers translation changes
-  const labels = {
-    cost: {tag: t('labels.cheapest'), title: t('labels.networkCost')},
-    popularity: {
-      tag: t('labels.mostPopular'),
-      title: t('labels.popularity'),
-    },
-    security: {
-      tag: t('labels.safest'),
-      title: t('labels.security'),
-    },
-  };
+  // // moving this up so state change triggers translation changes
+  // const labels = {
+  //   cost: {tag: t('labels.cheapest'), title: t('labels.networkCost')},
+  //   popularity: {
+  //     tag: t('labels.mostPopular'),
+  //     title: t('labels.popularity'),
+  //   },
+  //   security: {
+  //     tag: t('labels.safest'),
+  //     title: t('labels.security'),
+  //   },
+  // };
 
   // const handleFilterChanged = useCallback(
   //   (e: Event) => {
--- a/packages/web-app/src/containers/selectChainForm/index.tsx
+++ b/packages/web-app/src/containers/selectChainForm/index.tsx
@@ -1,22 +1,22 @@
 import {
   ButtonText,
-  Dropdown,
-  IconChevronDown,
-  ListItemAction,
+  // Dropdown,
+  // IconChevronDown,
+  // ListItemAction,
   ListItemBlockchain,
 } from '@aragon/ui-components';
 import styled from 'styled-components';
 import {useTranslation} from 'react-i18next';
 import {Controller, useFormContext} from 'react-hook-form';
-import React, {useCallback, useState} from 'react';
+import React, {useState} from 'react';
 
 import useScreen from 'hooks/useScreen';
 import {CHAIN_METADATA, SupportedNetworks} from 'utils/constants';
 import {useNetwork} from 'context/network';
-import {trackEvent} from 'services/analytics';
+// import {trackEvent} from 'services/analytics';
 
 type NetworkType = 'main' | 'test';
-type SortFilter = 'cost' | 'popularity' | 'security';
+// type SortFilter = 'cost' | 'popularity' | 'security';
 
 const SelectChainForm: React.FC = () => {
   const {t} = useTranslation();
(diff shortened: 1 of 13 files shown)


packages/web-app/src/hooks/useClient.tsx
```diff
@@ -22,17 +29,26 @@ export const useClient = () => {
 };
 
 export const UseClientProvider: React.FC = ({children}) => {
-  const {signer} = useWallet();
+  const {signer, chainId} = useWallet();
   const [client, setClient] = useState<Client>();
+  const {network} = useNetwork();
   const [context, setContext] = useState<SdkContext>();
-  // TODO hook this up to useNetwork once other networks become useable.
+
```
WissenIstNacht, line 30: What does the wallet's `chainId` do here? The wallet is never responsible for telling the app which network it's in. As the todo said, all this needed was to add the line `const {network} = useNetwork();` and then delete the line `const network = 'goerli';`.

//Little detail: Since SDK is going with `mainnet` instead of `ethereum` you'll have to do `network === 'ethereum' ? 'mainnet' : network` when assigning the network in the `contextParams` object.

sepehr2github: Hey, So there are two pages that don't have a network in their path, explore page and dao creation, for dao creation it doesn't matter because It will check the user selection but for explore page which newest daos are there we had to select a chain so we decided to show the active daos according to wallet chain

packages/web-app/src/hooks/useClient.tsx
```diff
@@ -22,17 +29,26 @@ export const useClient = () => {
...
+    //TODO: this should change to ethereum once the contract has been deployed
+    return 'goerli';
+  }, [chainId, network]);
 
   useEffect(() => {
-    const network = 'goerli';
     const contextParams: ContextParams = {
-      network: network,
+      //TODO: replace ethereum with mainnet in activeNetworks
+      network: activeNetwork === 'ethereum' ? 'mainnet' : 'goerli',
```
sepehr2github: @WissenIstNacht It's already there ?!

packages/web-app/src/hooks/useClient.tsx
WissenIstNacht: Well, that's still wrong and you'll have problems sooner or later if you query the `chainId` from the wallet inside this hook. If you want to have a special behavior for the explorer page, than you should do that in the explorer page and not here.

The network context gives you a callback called `setNetwork`, which lets you set the network for the app.  So if you want a specific network on the explorer page (for example, the user's current wallet network), you take the logic that you have here and use it in the explorer page to set the network.

As a side comment: When exactly would `activeNetwork` be the user's wallet network? `network` will always either be `goerli` or `ethereum`, so it will always exit after the first if statement...?

packages/web-app/src/hooks/useClient.tsx
WissenIstNacht: Yes, this you would have to keep. But you will have to replace `goerli` at the end with `network` sooner or later.

sepehr2github: yes but for now, let's go with this condition, will update later on, because it useNetwork type the mainnet input doesn't exist

packages/web-app/src/hooks/useClient.tsx
sepehr2github: I found out the use network has a default value, So in this case It doesn't need setNetwork in explore page also in the future It will keep the last network variable, Which would be fine for now, Also in the use network state the default value should be Ethereum but I used goerli for now because the mainnet contract data is not there

packages/web-app/src/containers/selectChainForm/index.tsx
```diff
@@ -1,62 +1,62 @@
...
   const {t} = useTranslation();
   const {isMobile} = useScreen();
   const {setNetwork} = useNetwork();
   const {control} = useFormContext();
 
-  const [isOpen, setIsOpen] = useState(false);
-  const [sortFilter, setFilter] = useState<SortFilter>('cost');
+  // const [isOpen, setIsOpen] = useState(false);
+  // const [sortFilter, setFilter] = useState<SortFilter>('cost');
   const [networkType, setNetworkType] = useState<NetworkType>('test');
```
jamesej, line 30: This I think should be `useState<NetworkType>('main');` because mainnet is the default selection and you can't see your default selection if this defaults to 'test' as now.

jamesej (changes requested):
Looks good except for this one minor thing about defaulting to the tab that shows the default network in DAO creation

jamesej (approved):
All good! many thanks

PR #811, opened 2023-04-12
Feature: APP-1125 - SCC: Store verified contracts

Fabricevladimir:
## Description
- Stores the verified smart contract 

Task: [APP-1125](https://aragonassociation.atlassian.net/browse/APP-1125)

## Type of change

## Checklist:

[APP-1125]: https://aragonassociation.atlassian.net/browse/APP-1125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/web-app/src/containers/smartContractComposer/components/contractAddressValidation.tsx
```diff
@@ -47,15 +49,15 @@ const icons = {
...
   const {alert} = useAlertContext();
+  const {address} = useWallet();
   const {network} = useNetwork();
 
   const [verificationState, setVerificationState] = useState<TransactionState>(
     TransactionState.WAITING
   );
 
-  const {control, setValue, resetField, setError} =
+  const {control, resetField, setValue, setError} =
```
sepehr2github, line 59: I'm wondering How this change could help, Just kidding

packages/web-app/src/services/cache.ts
```diff
@@ -21,3 +23,70 @@ export async function getFavoritedDaosFromCache(
...
+export function getVerifiedSmartContracts(
+  walletAddress: string | null,
+  chainId?: SupportedChainID
+): SmartContract[] {
+  // Ensure the daoAddress parameter is defined
+  if (!walletAddress) {
+    throw new Error('daoAddress must be defined');
+  }
+
+  const verifiedContracts = JSON.parse(
```
sepehr2github:
```
let verifiedContracts = {} as VerifiedContracts;
try {
  verifiedContracts = JSON.parse(localStorage.getItem(VERIFIED_CONTRACTS_KEY) || '{}');
} catch (error) {
  console.error('Error parsing verified contracts from localStorage:', error);
}
```
this might help because parsing might cause application crashes if wrong data stored

packages/web-app/src/services/cache.ts
```diff
@@ -21,3 +23,70 @@ export async function getFavoritedDaosFromCache(
...
+  walletAddress: string | null,
+  chainId: SupportedChainID
+): void {
+  // Ensure the contract, daoAddress, and chainId parameters are defined
+  if (!contract || !walletAddress || !chainId) {
+    throw new Error('Contract, daoAddress, and chainId must be defined');
+  }
+
+  // get the contracts from local storage
+  const verifiedContracts = JSON.parse(
```
sepehr2github: You can define it on top and use the variable for both funcitons

Fabricevladimir: The issue with pulling data from local storage outside of the function is that you won't get the latest changes as. Also, I don't see any benefit in using a global variable here especially considering the final destination of these functions is going to be on the backend server.

Fabricevladimir: The issue with pulling data from local storage outside of the function is that you won't get any changes. Also, I don't see any benefit in using a global variable here especially considering the final destination of these functions is going to be on the backend server.

packages/web-app/src/services/cache.ts
Fabricevladimir: This is almost paranoia lol . This is why we really should have an Error Boundary. I'm not satisfied how we're handling errors in general.

sepehr2github (approved):
Thanks, Seems good to go

PR #813, opened 2023-04-13
Fix: fetch list of transfers from alchemy api

sepehr2github:
## Description

-refactor useDaoTransfers to fetch erc20 transfers from alchemy api

Task: [APP-2042](https://aragonassociation.atlassian.net/browse/APP-2042)

## Type of change

## Checklist:

- [x] I have updated the ``CHANGELOG.md`` file in the root folder.

[APP-2042]: https://aragonassociation.atlassian.net/browse/APP-2042?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/web-app/src/hooks/useDaoTransfers.tsx
```diff
@@ -6,39 +6,135 @@ import {
...
   daoAddressOrEns: Address
 ): HookData<Transfer[]> => {
   const {client} = useClient();
+  const {network} = useNetwork();
 
   const [data, setData] = useState<Transfer[]>([]);
   const [error, setError] = useState<Error>();
   const [isLoading, setIsLoading] = useState(false);
   const pendingDepositsTxs = useReactiveVar(pendingDeposits);
+  const provider = useSpecificProvider(CHAIN_METADATA[network].id);
```
Fabricevladimir: If we are just using the App network to determine the provider, it's probably better to get the provider from `useProviders` hook since the reference is stable there and won't keep changing on every render.

packages/web-app/src/hooks/useDaoTransfers.tsx
```diff
@@ -6,39 +6,135 @@ import {
...
+
+  function sortByCreatedAt(a: Transfer, b: Transfer): number {
```
Fabricevladimir: This is nitpicking admittedly, but this function is getting recreated every render. Might be more performant to define it outside of the React component.

packages/web-app/src/hooks/useDaoTransfers.tsx
```diff
@@ -6,39 +6,135 @@ import {
...
+                transactionId: hash,
+              };
+            }
+          );
+
+        const erc20DepositsList = await Promise.all(erc20TransfersListPromises);
+
+        if (clientTransfers?.length) {
+          const subgraphTransfers = clientTransfers.filter(
+            t => t.type === TransferType.WITHDRAW || t.tokenType === 'native'
```
Fabricevladimir, line 134: I'm assuming Alchemy isn't providing the native token transfers? Shouldn't there be a way to do so using the category in the options object?

packages/web-app/src/hooks/useDaoTransfers.tsx
```diff
@@ -6,39 +6,135 @@ import {
...
-          const deposits = transfers.filter(
+        // Fetch the token list using the Alchemy API
+        const res = await fetch(url, options);
+        const alchemyTransfersList = await res.json();
+
+        // filter the erc20 token deposits
+        const erc20TransfersListPromises =
+          alchemyTransfersList.result.transfers.map(
+            async ({from, rawContract, metadata, hash}: AlchemyTransfer) => {
+              const {name, symbol, decimals} = await getTokenInfo(
```
Fabricevladimir, line 106: It's kind of weird that Alchemy isn't providing all the token information too, sigh.

Also, this is going to make more duplicate requests the more transfers with the same token that take place. So if you have two transfers with the same token, you're fething the same token information twice. This is a good place for cacheing requests actually. 

That said, these hooks will get refactored eventually with react-query so I'm not requesting a change there unless it's something very quick and simple.

packages/web-app/src/hooks/useDaoTransfers.tsx
```diff
@@ -6,39 +6,135 @@ import {
...
+          );
+
+        const erc20DepositsList = await Promise.all(erc20TransfersListPromises);
+
+        if (clientTransfers?.length) {
+          const subgraphTransfers = clientTransfers.filter(
+            t => t.type === TransferType.WITHDRAW || t.tokenType === 'native'
+          );
+
+          const deposits = clientTransfers.filter(
```
Fabricevladimir, line 137: Why do you need these deposits here considering you're already getting them from Alchemy? You can pass in the transfer type to fetch from the subgraph and get only `WITHDRAW` and use the txHashes from Alchemy deposit response to remove/update pending deposits.

packages/web-app/src/hooks/useDaoTransfers.tsx
```diff
@@ -6,39 +6,135 @@ import {
...
         });
 
-        if (transfers?.length) {
-          const deposits = transfers.filter(
+        // Fetch the token list using the Alchemy API
+        const res = await fetch(url, options);
+        const alchemyTransfersList = await res.json();
+
+        // filter the erc20 token deposits
+        const erc20TransfersListPromises =
```
Fabricevladimir: Can you update the name to reflect that this is only deposits?

Fabricevladimir (changes requested):

packages/web-app/src/hooks/useDaoTransfers.tsx
sepehr2github: Ah sorry forgot to mention that, Yes, Alchemy provides native tokens as well but for one of the daos It seems it had a bug and couldn't return the transfer, on the other hand, SDK Provide native transfers, because I wasn't confident about the API for native tokens I decided to use SDK but We can fetch both from API as well

packages/web-app/src/hooks/useDaoTransfers.tsx
sepehr2github: Yeah, I left the caching part for react Query refactor!

Fabricevladimir (approved):

PR #814, opened 2023-04-14
Chore: APP-2106 - Remove useDaoParam

Fabricevladimir:
## Description

- uses the newly refactored `useDaoDetailsQuery` hook in place of the the deprecated `useDaoDetails` hook
- removes `useDaoParam` hook

Task: [APP-2106](https://aragonassociation.atlassian.net/browse/APP-2106)

## Type of change

## Checklist:

[APP-2106]: https://aragonassociation.atlassian.net/browse/APP-2106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/web-app/src/containers/actionBuilder/index.tsx
```diff
@@ -97,10 +93,10 @@ const Action: React.FC<ActionsComponentProps> = ({name, actionIndex}) => {
 };
 
 const ActionBuilder: React.FC = () => {
-  const {data: daoAddress} = useDaoParam();
+  const {dao: daoAddressOrEns} = useParams();
   const {network} = useNetwork();
   const {selectedActionIndex: index, actions} = useActionsContext();
-  const {data: tokens} = useDaoBalances(daoAddress);
+  const {data: tokens} = useDaoBalances(daoAddressOrEns?.toLowerCase() || '');
```
sepehr2github: Not sure but Might be better to pass the address from daoDetails not daoAddressOrEns

packages/web-app/src/containers/compareSettings/index.tsx
```diff
@@ -65,23 +60,19 @@ const CompareSettings: React.FC = () => {
...
       {/* METADATA*/}
-      <CompareMetadata
-        daoId={daoAddressOrEns}
-        daoDetails={daoDetails}
-        view={selectedButton}
-      />
+      <CompareMetadata daoDetails={daoDetails} view={selectedButton} />
 
       {/* GOVERNANCE */}
       {isTokenVotingSettings(pluginSettings) ? (
```
sepehr2github, line 66: isTokenVotingSettings(pluginSettings) &&  might be better, Just a small thing, not a big deal

packages/web-app/src/containers/configureActions/index.tsx
```diff
@@ -9,14 +10,13 @@ import AddActionMenu from 'containers/addActionMenu';
...
-import {useDaoParam} from 'hooks/useDaoParam';
 
 const ConfigureActions: React.FC = () => {
-  const {data: daoId} = useDaoParam();
+  const {data: daoId} = useParams();
   const {t} = useTranslation();
   const {open} = useGlobalModalContext();
   const {actions} = useActionsContext();
-  const {data: availableActions} = useDaoActions(daoId);
+  const {data: availableActions} = useDaoActions(daoId ?? '');
```
sepehr2github: useDaoActions supports ensName right?

packages/web-app/src/pages/transfers.tsx
```diff
@@ -186,3 +179,4 @@ const Transfers: React.FC = () => {
 };
```
sepehr2github, line 185: please rebase your branch because moving this outside of component block could bring the transfer issue back

packages/web-app/src/pages/transfers.tsx
```diff
@@ -78,12 +77,6 @@ const Transfers: React.FC = () => {
     ]
   );
 
-  const SectionContainer = styled.div.attrs({className: 'my-3 desktop:my-5'})``;
-
-  if (isLoading) {
```
sepehr2github, line 83: Why did you remove this?, wouldn't better to replace with another loading?

packages/web-app/src/pages/proposeSettings.tsx
```diff
@@ -134,7 +132,10 @@ const ProposeSettings: React.FC = () => {
       <FullScreenStepper
         wizardProcessName={t('newProposal.title')}
         navLabel={t('navLinks.settings')}
-        returnPath={generatePath(EditSettings, {network, dao: daoId})}
+        returnPath={generatePath(EditSettings, {
+          network,
+          dao: daoDetails?.address,
```
sepehr2github: I suppose this should be ens first and if not address 

```
toDisplayEns(daoDetails?.ensDomain) || daoDetails?.address
```

sepehr2github:
Great Job my friend, Just left a few comments! Well done ;D

packages/web-app/src/pages/transfers.tsx
Fabricevladimir: The loading should be coming from a mixture of DAO details and the finance hooks, but the finance hooks don't export the loading state, which needs to be fixed when we refactor

packages/web-app/src/pages/transfers.tsx
Fabricevladimir: Branch is already rebased.

packages/web-app/src/containers/actionBuilder/index.tsx
Fabricevladimir: At this point it doesn't matter because `daoDetails` has already been fetched by the parent and has already redirected to the ens url if it's available, or redirected to 404 if there is no proper DAO. But in the spirit of making you happy lol

packages/web-app/src/containers/compareSettings/index.tsx
Fabricevladimir: While that might be slightly more readable, you will have to check every condition every single time regardless of whether the first one is true.

packages/web-app/src/containers/configureActions/index.tsx
Fabricevladimir: Yes and this can be either since the parent is will have already taken care of redirecting to ens url if it's available

packages/web-app/src/pages/proposeSettings.tsx
Fabricevladimir: Good catch, good catch. I'm actually really tired of this toDisplayEns. A cleaner and perhaps more elegant approach would have been to augment the dao details we get back from the sdk with a `daoId` or `navId` prop which would do this once and then navigate using that. Thoughts?

The effect there would of course be I would have to update everywhere that takes in `DaoDetails` as a type to the new custom type

packages/web-app/src/pages/transfers.tsx
sepehr2github: Ok, so you moved this intentionally?

packages/web-app/src/pages/proposeSettings.tsx
sepehr2github: Frankly, depends On you, this PR is already big, So if you update this place and just leave a TODO I won't complain otherwise feel free to fix it, could be a good idea

packages/web-app/src/pages/transfers.tsx
Fabricevladimir: Yes. If you check the transferList component, you'll see that I'm generating a new id even if it's not ideal. There was an issue with having the Styled component inside the component block where it was regenerating this component for every transaction. Also, it was generating a warning for every single transaction because of that.

sepehr2github (approved):
Thanks for the changes. LGTM

PR #834, opened 2023-05-07
Fix: APP-2183 - Polygon DAO Crashes on Resolving ENS Name

Barukimang:
## Description

@sepehr2github with this fix it is possible to open DAOs on Polygon and Mumbai which do have an ENS Domain (whether created from the SDK or from the App somehow). This Polygon DAO with ENS Domain can open e.g.: [click here](http://localhost:3000/#/daos/polygon/0x3782242c028028acb9282b9c88b28810364e66fd/dashboard)

This is by no means a comprehensive fix yet, since many other pages apart from the Dashboard make use of the `toDisplayEns()` function, but a first attempt :) Feel free to take it from here.

- add Ens resolver for L2 daos that could register their ens name
- fix the bug that users could register L2 ens names in dao creation

Task: [APP-2138](https://aragonassociation.atlassian.net/browse/APP-2183)

## Type of change

## Checklist:

[APP-2138]: https://aragonassociation.atlassian.net/browse/APP-2138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/containers/selectChainForm/index.tsx
+++ b/packages/web-app/src/containers/selectChainForm/index.tsx
@@ -23,7 +23,7 @@ const SelectChainForm: React.FC = () => {
   const {t} = useTranslation();
   const {isMobile} = useScreen();
   const {setNetwork, network} = useNetwork();
-  const {control} = useFormContext();
+  const {control, resetField} = useFormContext();
 
   // const [isOpen, setIsOpen] = useState(false);
   // const [sortFilter, setFilter] = useState<SortFilter>('cost');
@@ -153,6 +153,13 @@ const SelectChainForm: React.FC = () => {
                     label: CHAIN_METADATA[selectedNetwork].name,
                     network: networkType,
                   });
+                  if (
+                    CHAIN_METADATA[selectedNetwork].name === 'Polygon' ||
+                    CHAIN_METADATA[selectedNetwork].name === 'Matic'
+                  ) {
+                    // reset daoEnsName if network changed to L2
+                    resetField('daoEnsName');
+                  }
                 }}
                 selected={CHAIN_METADATA[selectedNetwork].id === field.value.id}
                 // tag={index === 0 ? labels[sortFilter].tag : undefined}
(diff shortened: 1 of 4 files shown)


packages/web-app/src/containers/selectChainForm/index.tsx
```diff
@@ -153,6 +153,13 @@ const SelectChainForm: React.FC = () => {
                     label: CHAIN_METADATA[selectedNetwork].name,
                     network: networkType,
                   });
+                  if (
+                    CHAIN_METADATA[selectedNetwork].name === 'Polygon' ||
```
Fabricevladimir: THis works, but I'm wondering why not compare `selectedNetwork` itself considering typescript static analysis will catch any sort of error in the string?

```javascript
if(selectedNetwork === 'polygon')
```

packages/web-app/src/hooks/useDaoDetails.tsx
```diff
@@ -18,13 +21,23 @@ import {useClient} from './useClient';
  */
 async function fetchDaoDetails(
   client: Client | undefined,
-  daoAddressOrEns: string | undefined
+  daoAddressOrEns: string | undefined,
+  provider: JsonRpcProvider,
+  isL2NetworkEns: boolean,
+  redirectDaoToAddress: (address: string | null) => void
```
Fabricevladimir, line 27: Not really a fan of redirecting inside this function as this should have zero knowledge of navigation. Tell me what you think about it @sepehr2github

Also, it would be nice if we disabled the whole thing if isL2networkEns, resolve to the address and then redirect before fetching the query. This would save us an unecessary call each time the ens is given. 

Also, consider tracking the inital query so we remove the ens from the DAO objet returned so that everytime we navigate using a button in the application, we don't have to keep replacing the ens with the address

Fabricevladimir (approved):
Thanks Sepehr and Pieter. I left a couple of comments for improvement.

PR #837, opened 2023-05-09
Feature: APP-2143 - Input Wallet w/ ENS Support

Fabricevladimir:
## Description

- Adds ENS support for input wallet

Note: Because of the breaking changes that will be introduced by the type of the new component, a legacy component is kept in the flows where the new one has not been merged into yet. Currently the only flow supporting the new component is the Withdraw (both main proposal and finance flow)

PS: Couldn't find a proper explorer for ENS lookup on Goerli. Let me know if you can think of one.  
Also, `phitest.ens` can be used on Goerli 

Task: [APP-2143](https://aragonassociation.atlassian.net/browse/APP-2143)

## Type of change

## Checklist:

[APP-2143]: https://aragonassociation.atlassian.net/browse/APP-2143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui-components/src/utils/addresses.tsx
+++ b/packages/ui-components/src/utils/addresses.tsx
@@ -8,7 +8,7 @@ export function shortenAddress(address: Address | null) {
   if (IsAddress(address))
     return (
       address.substring(0, 5) +
-      '...' +
+      '…' +
       address.substring(address.length - 4, address.length)
     );
   else return address;
@@ -18,4 +18,23 @@ export function shortenAddress(address: Address | null) {
 export function IsAddress(address: Address | null) {
   const re = /0x[a-fA-F0-9]{40}/g;
   return Boolean(address?.match(re));
+}
+
+export function isEnsDomain(input: string | null): boolean {
+  if (!input) return false;
+
+  // The pattern for a valid ENS domain:
+  // - starts with an alphanumeric character (case-insensitive)
+  // - followed by zero or more alphanumeric characters, hyphens, or underscores (case-insensitive)
+  // - ends with '.eth'
+  const ensPattern = /^[a-zA-Z0-9]([a-zA-Z0-9-_]*[a-zA-Z0-9])?\.eth$/;
+  return ensPattern.test(input);
+}
+
+export function shortenENS(input: string | null): string {
+  if (!input || !isEnsDomain(input)) return input as string;
+
+  const [subdomain] = input.split('.eth');
+  const shortenedSubdomain = subdomain.slice(0, 7);
+  return `${shortenedSubdomain}…eth`;
 }
--- a/packages/ui-components/src/utils/addresses.tsx
+++ b/packages/ui-components/src/utils/addresses.tsx
@@ -1,9 +1,5 @@
-// NOTE Eventually have these types expressed as string patterns [VR 22-11-2021].
-export type Address = string;
-export type EnsName = string;
-
 // get truncated address
-export function shortenAddress(address: Address | null) {
+export function shortenAddress(address: string | null) {
   if (address === null) return '';
   if (IsAddress(address))
     return (
@@ -15,7 +11,7 @@ export function shortenAddress(address: Address | null) {
 }
 
 // check label type
-export function IsAddress(address: Address | null) {
+export function IsAddress(address: string | null) {
   const re = /0x[a-fA-F0-9]{40}/g;
   return Boolean(address?.match(re));
 }
--- a/packages/ui-components/src/utils/addresses.tsx
+++ b/packages/ui-components/src/utils/addresses.tsx
@@ -22,12 +22,13 @@ export function IsAddress(address: Address | null) {
 
 export function isEnsDomain(input: string | null): boolean {
   if (!input) return false;
-
   // The pattern for a valid ENS domain:
   // - starts with an alphanumeric character (case-insensitive)
   // - followed by zero or more alphanumeric characters, hyphens, or underscores (case-insensitive)
   // - ends with '.eth'
-  const ensPattern = /^[a-zA-Z0-9]([a-zA-Z0-9-_]*[a-zA-Z0-9])?\.eth$/;
+  const ensPattern = new RegExp(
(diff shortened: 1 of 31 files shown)


packages/web-app/src/utils/library.ts
```diff
@@ -553,3 +563,115 @@ export function toDisplayEns(ensName?: string) {
...
+    }
+
+    // Create a new Address instance
+    const addressObj = new Web3Address(provider, addressToSet, ensNameToSet);
+
+    // If a provider is available, try to resolve the missing piece (address or ENS name)
+    if (provider) {
+      if (addressToSet && !ensNameToSet) {
+        ensNameToSet =
+          (await provider.lookupAddress(addressToSet)) ?? undefined;
```
sepehr2github: would be good if we wrap these two promises into a try catch

packages/web-app/src/utils/constants/chains.ts
```diff
@@ -112,6 +115,7 @@ export const CHAIN_METADATA: ChainList = {
       symbol: 'ETH',
       decimals: 18,
     },
+    lookupURL: 'https://etherscan.io/name-lookup-search?id=',
```
sepehr2github, line 122: please set this for goerli as well

packages/web-app/src/context/createProposal.tsx
```diff
@@ -143,19 +140,15 @@ const CreateProposalProvider: React.FC<Props> = ({
...
           actions.push(
             client.encoding.withdrawAction({
               amount: BigInt(
                 Number(action.amount) * Math.pow(10, action.tokenDecimals)
               ),
-              recipientAddressOrEns: receiver,
+
+              /* TODO: SDK doesn't accept ens names, this should be removed once they
+                 fixed the issue */
+              recipientAddressOrEns: action.to.address,
```
sepehr2github, line 155: Could you check with jose again, based on my last updates they fixed the ens acceptance issue in new version

packages/ui-components/src/utils/addresses.tsx
```diff
@@ -19,3 +19,23 @@ export function IsAddress(address: Address | null) {
...
+  const ensPattern = new RegExp(
+    '^([a-z0-9]+(-[a-z0-9]+)*.)*[a-z0-9]+(-[a-z0-9]+)*.eth$'
+  );
+  return ensPattern.test(input);
+}
+
+export function shortenENS(input: string | null): string {
+  if (!input || !isEnsDomain(input)) return input as string;
+
+  const [subdomain] = input.split('.eth');
```
sepehr2github: not a biggie, I think here could be better to use domain because subdomains are names with .dao.eth or any nested domain, therefore might be confusing for future implementations

packages/ui-components/tailwind.config.js
```diff
@@ -178,6 +178,7 @@ module.exports = {
       ringWidth: ['active', 'focus-visible', 'disabled'],
       outline: ['focus-visible', 'active', 'focus', 'hover'],
       rounded: ['focus-visible'],
+      cursor: ['disabled'],
```
sepehr2github: Why is that?

packages/web-app/src/containers/configureWithdraw/index.tsx
```diff
@@ -217,6 +221,22 @@ const ConfigureWithdrawForm: React.FC<ConfigureWithdrawFormProps> = ({
     [alert, t]
   );
 
+  const resolveEnsNameFromAddress = useCallback(
```
sepehr2github, line 224: Please Correct me if I'm wrong, but couldn't we create a memorized instance of Web3Address class here and use its helper methods for resolving and lookup, otherwise might be good to use these as helpers in utils as well

sepehr2github (commented):
Was a wonderful job and now I understand why It took that much xD

packages/web-app/src/utils/constants/chains.ts
Fabricevladimir: I can't find one for Goerli!

packages/ui-components/tailwind.config.js
Fabricevladimir: Selim and I were playing around with the disabled cursor, but it looked horrendous. Good catch

packages/web-app/src/containers/configureWithdraw/index.tsx
Fabricevladimir: We could have, but Selim was adamant about wanting the component to be able to take care of the resolving and lookup. This is an iffy territory for me because this approach removes flexibility, but at the same time, what else are you going to use the input-wallet with ens for if not for resolving ens and addresses? The other point too is that these are technically optional.

Finally, we have to take into account whoever is using this component. If we do the resolving outside of the component, they have to resolve the component everywhere they use this. Right now they merely need to pass in the resolver functions and the component takes care of it by default

packages/web-app/src/context/createProposal.tsx
Fabricevladimir: I tried this, but sadly it died on me. It's not a big deal yet because we do show the ens if available when the proposal has been published. The issue here though is because we're doing all sorts of resolving left, right and center, we lose which one the user put in. I can see it becoming quite confusing when someone gives you their wallet address and then when the proposal is created what you now see is their ens name. You can't even double check to make sure you didn't make a mistake without opening a new tab to lookup the address associated with that ens. I want to raise this and see what the other people think. The `Web3Address` would just be a property on the form but the `to` would be exactly what the user put in

Fabricevladimir:
@sepehr2github I've had to remove the `Address` and `EnsName` type exports as well as move the `shortenAddress` function into the web-app in order to fix the breaking tests. I'm entirely unsure why they started breaking now of all times. The error was related to babel thinking that the type exports are experimental flow syntax.

packages/web-app/src/utils/constants/chains.ts
sepehr2github: Ah you're right, goerli doesn't support that

sepehr2github (approved):
Thanks for the changes

PR #839, opened 2023-05-09
Fix: APP-2176 Multisig proposals should default creation time to now

babarella:
## Description

Multisig wallets are used for speed and performance, not for “voting” (they don’t even vote, but approve tx). It’s understandable we start ERC20Voting proposals +10 minutes after proposal creation, but that shouldn’t happen for multisig proposals where default starting time should be 0 (which is used by the protocol as “now”).

Task: [APP-2176](https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/40?modal=detail&selectedIssue=APP-2176&assignee=712020%3A6ea8272e-a7c7-4fcc-8652-afd007bf3dba)

## Type of change

## Checklist:

[APP-2176]: https://aragonassociation.atlassian.net/browse/APP-2176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/containers/reviewProposal/index.tsx
+++ b/packages/web-app/src/containers/reviewProposal/index.tsx
@@ -3,11 +3,12 @@ import {Link, VoterType} from '@aragon/ui-components';
 import TipTapLink from '@tiptap/extension-link';
 import {EditorContent, useEditor} from '@tiptap/react';
 import StarterKit from '@tiptap/starter-kit';
-import {format} from 'date-fns';
+import {format, formatDistanceToNow, Locale} from 'date-fns';
 import React, {useEffect, useMemo} from 'react';
 import {useFormContext} from 'react-hook-form';
 import {TFunction, useTranslation} from 'react-i18next';
 import styled from 'styled-components';
+import * as Locales from 'date-fns/locale';
 
 import {ExecutionWidget} from 'components/executionWidget';
 import {useFormStep} from 'components/fullScreenStepper';
@@ -43,7 +44,7 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
   defineProposalStepNumber,
   addActionsStepNumber,
 }) => {
-  const {t} = useTranslation();
+  const {t, i18n} = useTranslation();
   const {setStep} = useFormStep();
 
   const {data: daoDetails} = useDaoDetailsQuery();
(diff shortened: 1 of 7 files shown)


packages/web-app/src/context/createProposal.tsx
```diff
@@ -274,16 +274,22 @@ const CreateProposalProvider: React.FC<Props> = ({
...
-          ? new Date(
-              `${getCanonicalDate()}T${getCanonicalTime({
-                minutes: 10,
-              })}:00${getCanonicalUtcOffset()}`
-            )
-          : new Date(
-              `${startDate}T${startTime}:00${getCanonicalUtcOffset(startUtc)}`
-            );
+      let startDateTime: Date;
+      const startMinutesDelay = isMultisigVotingSettings(pluginSettings)
```
Fabricevladimir, line 278: This is a bit incorrect because by the time the transaction is executed, the start date will have already be in the past. 

A better approach is to conditionally include the start date in the object being returned based on whether the multisig proposal should start immediately (now)

If you check the type, you'll see that the start date is optional. By keeping it unset, the proposal will start immediatley. Also, you'll have to do the same thing in the propose settings wrapper.

packages/web-app/src/containers/reviewProposal/index.tsx
```diff
@@ -77,18 +77,20 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
 
   const startDate = useMemo(() => {
     const {startSwitch, startDate, startTime, startUtc} = values;
+
     if (startSwitch === 'now') {
+      const startMinutesDelay = isMultisigVotingSettings(daoSettings) ? 0 : 10;
```
Fabricevladimir, line 83: I left a comment to Selim and Carlos on the corresponding ticket. I believe the start and end dates will be shown differently.That said, I'm not sure if you want to tackle this with the same PR or open a separate one for it. Up to you.

Fabricevladimir (changes requested):
Thanks, Nikita. Good stuff. Unfortunately the proposal creation fails with this approach because the time immediately becomes in the past.

packages/web-app/src/containers/reviewProposal/index.tsx
```diff
@@ -77,29 +78,69 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
...
+      return t('labels.now');
+    }
+
+    return `${format(
+      startDate,
+      KNOWN_FORMATS.proposals
+    )} ${getFormattedUtcOffset()}`;
+  }, [daoSettings, startDate, t, values]);
 
   const formattedEndDate = useMemo(() => {
```
Fabricevladimir, line 111: Might want to be a bit more explicit with the names or add a comment to differentiate the two.

packages/web-app/src/containers/votingTerminal/infoTab.tsx
```diff
@@ -28,6 +28,7 @@ type Props = Pick<
   missingApprovalOrParticipation: number;
   uniqueVoters?: number;
   voteOptions: string;
+  preciseEndDate?: string;
```
Fabricevladimir: Nothing *wrong* with this, but you can pick the property off of the parent props.

packages/web-app/src/context/createProposal.tsx
```diff
@@ -274,16 +274,22 @@ const CreateProposalProvider: React.FC<Props> = ({
       const ipfsUri = await pluginClient?.methods.pinMetadata(metadata);
 
       // getting dates
```
Fabricevladimir, line 276: It would be nice to have these date calculations in one method so we don't have to constantly update this with almost the same code as the one in review page, but that's for another day

packages/web-app/src/context/createProposal.tsx
```diff
@@ -334,11 +342,22 @@ const CreateProposalProvider: React.FC<Props> = ({
         }
       }
 
+      /**
+       * For multisig proposals, in case "now" as start time is selected, we want
```
Fabricevladimir, line 346: Thanks for the comment!

Fabricevladimir (approved):
Nice work, thanks

packages/web-app/src/containers/votingTerminal/infoTab.tsx
```diff
@@ -170,7 +172,18 @@ const InfoTab: React.FC<Props> = ({
         </InfoLine>
         <InfoLine>
           <p>{t('votingTerminal.endDate')}</p>
-          <Strong>{endDate}</Strong>
+          <EndDateWrapper>
+            <Strong>
+              {t('votingTerminal.label.in')} {endDate}
```
Fabricevladimir: This needs to be removed once the proposal has been published so you might have to pass in the "in" as part of the `endDate`

Fabricevladimir (commented):
Just ran into this. Also there's an issue with the proposal not redirecting properly. It sends you to 404

PR #844, opened 2023-05-15
Feat: APP-2122 Add multiple withdrawals

babarella:
## Description

- Be able to add < 1 withdrawals to one withdrawal proposal initiated from the Finance Page
- Use the same accordion item, as we know from the action-builder for withdraw assets
- Instead of Add another action as the further CTA we use the label Add another withdraw (Key: newWithdraw.configureWithdraw.ctaAddAnother)

Task: [APP-2122](https://aragonassociation.atlassian.net/browse/APP-2122)

## Type of change

## Checklist:

[APP-2122]: https://aragonassociation.atlassian.net/browse/APP-2122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/web-app/src/containers/actionBuilder/addAddresses/index.tsx
```diff
@@ -27,12 +27,15 @@ export type CurrentDaoMembers = {
   currentDaoMembers?: MultisigMember[] | BalanceMember[];
 };
 
-type AddAddressesProps = ActionIndex & CustomHeaderProps & CurrentDaoMembers;
+type AddAddressesProps = ActionIndex &
+  CustomHeaderProps &
+  CurrentDaoMembers & {allowRemove?: boolean};
```
Fabricevladimir, line 32: This component already has a version that doesn't contain the accordion and its dropdown menu (see Manage members flow on a multisig DAO). What's the rational behind also adding this prop?

packages/web-app/src/containers/configureActions/index.tsx
```diff
@@ -10,33 +10,90 @@ import AddActionMenu from 'containers/addActionMenu';
...
+  hideAlert?: boolean;
+  addNewActionLabel?: string;
+  onAddNewActionClick?: () => void;
+  addExtraActionLabel?: string;
+  onAddExtraActionClick?: () => void;
+  allowEmpty?: boolean;
+}
+
+const ConfigureActions: React.FC<ConfigureActionsProps> = ({
+  label = i18n.t('newProposal.configureActions.yesOption') || '',
```
Fabricevladimir, line 30: Considering you have the transalation keys, why or it with an empty string?

packages/web-app/src/containers/configureActions/index.tsx
```diff
@@ -10,33 +10,90 @@ import AddActionMenu from 'containers/addActionMenu';
...
+  onAddExtraActionClick,
+  allowEmpty = true,
+}) => {
+  const {data: daoAddressOrEns} = useParams();
   const {t} = useTranslation();
   const {open} = useGlobalModalContext();
-  const {actions} = useActionsContext();
-  const {data: availableActions} = useDaoActions(daoAddressOrEns ?? '');
+  const {actions, addAction} = useActionsContext();
+  const {data: possibleActions} = useDaoActions(daoAddressOrEns ?? '');
```
Fabricevladimir, line 44: You might need to rebase/merge develop into this branch because there was a PR fixing this from "data" -> "dao".

packages/web-app/src/containers/actionBuilder/scc/index.tsx
```diff
@@ -45,7 +54,7 @@ const SCCAction: React.FC<ActionIndex> = ({actionIndex}) => {
           <div className="space-y-2">
             {(actionData.inputs as Input[]).map((input, index) => (
               <div key={input.name}>
-                <div className="mb-1.5 text-base font-bold text-ui-800 capitalize">
+                <div className="mb-1.5 text-base font-bold capitalize text-ui-800">
```
Fabricevladimir: Of all the things for the linter to complain about smh

```suggestion
                 <div className="mb-1.5 text-base font-bold text-ui-800 capitalize">
```

Fabricevladimir (commented):
Saw a lot of good things here! 

Here's a question: While I understand the need for the withdraw to have an allow remove property, what's the rationale behind adding it to all the other cards considering they are all different in terms of use? Add addresses for example is used already in a singular flow and isn't an action that we allow to have multiples of--same for remove addresses and mint tokens.

packages/web-app/src/containers/configureActions/index.tsx
babarella: Because of type inference. `label?: string` which means it's optional to pass. By providing the `string` default value, I narrow down the type inference from `string | undefined` to `string` which is important because removes unnecessary TS checks. 
However, when getting a translation string with `i18n.t('any-key')` => the type of it is `string | undefined` which creates a problem. Even though we know the key with such string definitely exist, Typescript still interprets it as `string | undefined` that why I added this extra `|| ''` to definitely narrow the type to `string`

packages/web-app/src/containers/configureActions/index.tsx
babarella: Rebased with `develop`, this change is not yet there.

Fabricevladimir: Weird, disregard 👍🏾

Fabricevladimir (approved):

PR #849, opened 2023-05-19
Feat: SCC input form validation

RakeshUP:
## Description

- Calling encodeFunctionData on ethers to validate user-entered data with the input data types given by the ABI
- ~~Builds on top of APP-2182. So review this after 2182 is merged.~~

Task: [APP-2203](https://aragonassociation.atlassian.net/browse/APP-2203)

## Type of change

## Checklist:

[APP-2203]: https://aragonassociation.atlassian.net/browse/APP-2203?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui-components/src/components/listItem/action.tsx
+++ b/packages/ui-components/src/components/listItem/action.tsx
@@ -102,7 +102,7 @@ const Container = styled.button.attrs(
 
     switch (mode) {
       case 'disabled':
-        className += ' text-ui-300 border-ui-200';
+        className += ' text-ui-300';
         className += bgWhite ? ' bg-ui-0' : ' bg-ui-50';
         break;
       case 'selected':
--- a/packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
+++ b/packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
@@ -54,7 +54,11 @@ const InputForm: React.FC<InputFormProps> = ({
   const {setValue, resetField} = useFormContext();
   const [formError, setFormError] = useState(false);
 
+  useEffect(() => setFormError(false), [selectedAction]);
+
   const composeAction = useCallback(async () => {
+    setFormError(false);
+
     const etherscanData = await getEtherscanVerifiedContract(
       selectedSC.address,
       network
@@ -83,6 +87,7 @@ const InputForm: React.FC<InputFormProps> = ({
         setValue(`actions.${actionIndex}.contractAddress`, selectedSC.address);
         setValue(`actions.${actionIndex}.contractName`, selectedSC.name);
         setValue(`actions.${actionIndex}.functionName`, selectedAction.name);
+        setValue(`actions.${actionIndex}.notice`, selectedAction.notice);
 
         selectedAction.inputs?.map((input, index) => {
           setValue(`actions.${actionIndex}.inputs.${index}`, {
@@ -117,6 +122,7 @@ const InputForm: React.FC<InputFormProps> = ({
     sccActions,
     selectedAction.inputs,
     selectedAction.name,
+    selectedAction.notice,
     selectedSC.address,
     selectedSC.name,
     setValue,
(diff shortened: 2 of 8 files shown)


packages/ui-components/src/components/listItem/action.tsx
```diff
@@ -78,7 +94,8 @@ type InputContainerProps = Pick<ListItemActionProps, 'mode' | 'bgWhite'>;
 const Container = styled.button.attrs(
   ({mode, bgWhite = false}: InputContainerProps) => {
     const baseLayoutClasses = 'flex items-center gap-x-1.5 w-full';
-    const baseStyleClasses = 'py-1.5 px-2 rounded-xl font-normal';
+    const baseStyleClasses =
+      'py-1.5 px-2 rounded-xl font-normal border-2 border-transparent';
```
Fabricevladimir: This border needs to take into account when the item is disabled. 
image

packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
```diff
@@ -47,6 +52,75 @@ const InputForm: React.FC<InputFormProps> = ({
   const {dao: daoAddressOrEns} = useParams();
   const {addAction, removeAction} = useActionsContext();
   const {setValue, resetField} = useFormContext();
+  const [formError, setFormError] = useState(false);
+
+  const composeAction = useCallback(async () => {
+    const etherscanData = await getEtherscanVerifiedContract(
```
Fabricevladimir, line 62: Do we not have the ABI already from when we verified the contract?

packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
```diff
@@ -97,40 +171,13 @@ const InputForm: React.FC<InputFormProps> = ({
...
-            smart_contract_name: selectedSC.name,
-            method_name: selectedAction.name,
-          });
-        }}
+        className="mt-5 mb-2 w-full desktop:w-max"
+        onClick={composeAction}
       />
+
+      {formError && (
+        <AlertInline label="Error with the inputs" mode="critical" />
```
Fabricevladimir, line 185: Why display the error here only and not under the input that has the error, at least for the ones that we can validate like address? We seem to have access to the component error for some of them in the `ComponentForType` component. We could use this to disable the compose button as well

Also, because this only gets displayed after we try to encode the action, when the error is fixed, this error state is still being displayed--even if we select a different method in the contract from the method that has the error.

Fabricevladimir (commented):
Looking good. Left a couple of comments.

packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
RakeshUP: Changed the staleTime to 5 minutes throughout the app for the etherscan query. It's being de-duped now

packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
RakeshUP: address and bytes can be validated, i don't have an idea about the other types. might have to do a spike to get more info.(tenderly needs to be evaluated as well as a spike)

that being said, just fixed the error state not being reset bug

packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
Fabricevladimir: Ahh. Actually this one is on me. I didn't realize we were using the queryclient from react query. I mean considering the likelyhood of change of a contract, we might as well make this never stale, but yeah, looks good

packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
Fabricevladimir: What I meant was like displaying "invalid address" under the address field instead of under the compose button. Does it make sense in this scenario? What do you think?

Fabricevladimir (approved):

packages/web-app/src/containers/smartContractComposer/components/inputForm.tsx
RakeshUP: I'll go through with the design team for their inputs on validation, as the designs don't have any validation cases. Will merge and try to do validation separately, maybe with tenderly integration

PR #859, opened 2023-05-25
Fix: Empty contract name on sourcify data

RakeshUP:


jamesej (approved):
Makes sense and seems to work, thanks!

PR #861, opened 2023-05-31
Feat: Refetch implementation for Proxy contracts

RakeshUP:
Task: [APP-2220](https://aragonassociation.atlassian.net/browse/APP-2220)

## Type of change

## Checklist:

[APP-2220]: https://aragonassociation.atlassian.net/browse/APP-2220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/hooks/useValidateContract.tsx
+++ b/packages/web-app/src/hooks/useValidateContract.tsx
@@ -22,7 +22,16 @@ export const useValidateContractEtherscan = (
     queryKey: ['verifyContractEtherscan', contractAddress, network],
     staleTime: Infinity,
     queryFn: () => {
-      return fetch(url).then(res => res.json());
+      return fetch(url).then(res => {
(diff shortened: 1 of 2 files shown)


sepehr2github (commented):
thanks, Rakesh, Seems good but two Q:

1) if I add a proxy contract, It will read from the implementation and show the address of the implementation contract, since we don't allow users to add a contract twice, should we avoid them from adding an implementation contract separately as well, because user added the proxy before with implementation actions

2) because we cache the implementation abi, what happened if contract changes the proxy pointer to a new implementation, there is no mechanism currently to deprecate the current implementations abi and add the new one

jamesej (approved):
Works fine, thanks

PR #870, opened 2023-06-09
Fixed a polygonscan api fetch bug

RakeshUP:
Etherscan API key was being used for Polygonscan.


jamesej (approved):
All looks good thanks

PR #874, opened 2023-06-13
fix: native tokens withdrawal action display on proposal page

babarella:
## Description

**Note for reviewer:**
This might seem like a bit of a hacky fix but the main point - it works. We need better support from `findInterface` of SDK function to eliminate the work-around introduced in this PR.

The withdrawal action of Native tokens such as ETH was not visible on proposal page.

Task: [APP-2200](https://aragonassociation.atlassian.net/browse/APP-2200)

## Type of change

## Checklist:

[APP-2200]: https://aragonassociation.atlassian.net/browse/APP-2200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/pages/proposal.tsx
+++ b/packages/web-app/src/pages/proposal.tsx
@@ -230,18 +230,6 @@ const Proposal: React.FC = () => {
           client?.decoding.findInterface(action.data) ||
           pluginClient?.decoding.findInterface(action.data);
 
-        if (!functionParams && action.to && action.value) {
-          return decodeWithdrawToAction(
-            action.data,
-            client,
-            apolloClient,
-            provider,
-            network,
-            action.to,
-            action.value
-          );
-        }
-
         switch (functionParams?.functionName) {
           case 'transfer':
             return decodeWithdrawToAction(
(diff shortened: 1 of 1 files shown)


packages/web-app/src/pages/proposal.tsx
```diff
@@ -230,6 +230,18 @@ const Proposal: React.FC = () => {
           client?.decoding.findInterface(action.data) ||
           pluginClient?.decoding.findInterface(action.data);
 
+        if (!functionParams && action.to && action.value) {
```
Fabricevladimir: While this works, I fear it *might* try to decode any SCC action that has both a to and native token value... But I can't think of another workaround save for running this action throught the `decodeSCCToAction` function if the withdraw decoding fails. That, of course, creates quite a bit of nastiness for an issue that's not UI-side.

UPDATE: Here's a proposal [minting an NFT](staging-app.aragon.org/#/daos/goerli/onelasttime.dao.eth/governance/proposals/0x77e465cc6584aa5fef57083aedb423bf799a2a54_0xe). Because the action has a `to`, some ETH `value`, and no SDK function params from the `findInterface` method, it is treated as a withdraw action.

Fabricevladimir (approved):
Thanks, Nikita. While I do think the native token withdraw is maybe slighly more important than a Smart Contract Composer action with the to and value params, I would suggest getting James' opinion before merging this.

PR #877, opened 2023-06-15
Feat: Network Name on Deposit Modal

RakeshUP:
Task: [APP-2237](https://aragonassociation.atlassian.net/browse/APP-2237)

## Type of change

## Checklist:

[APP-2237]: https://aragonassociation.atlassian.net/browse/APP-2237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/web-app/src/containers/transactionModals/DepositModal.tsx
+++ b/packages/web-app/src/containers/transactionModals/DepositModal.tsx
@@ -1,36 +1,25 @@
-import {
-  ButtonText,
-  IconLinkExternal,
-  Link,
-  WalletInputLegacy,
-  shortenAddress,
-} from '@aragon/ui-components';
+import {AlertInline, ButtonText, WalletInput} from '@aragon/ui-components';
 import React, {useCallback} from 'react';
 import {useTranslation} from 'react-i18next';
 import {generatePath, useNavigate} from 'react-router-dom';
 import styled from 'styled-components';
 
 import ModalBottomSheetSwitcher from 'components/modalBottomSheetSwitcher';
-import {useAlertContext} from 'context/alert';
 import {useGlobalModalContext} from 'context/globalModals';
 import {useNetwork} from 'context/network';
 import {useDaoDetailsQuery} from 'hooks/useDaoDetails';
 import {CHAIN_METADATA} from 'utils/constants';
 import {toDisplayEns} from 'utils/library';
 import {AllTransfers} from 'utils/paths';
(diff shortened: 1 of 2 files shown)


packages/web-app/src/containers/transactionModals/DepositModal.tsx
```diff
@@ -62,90 +41,93 @@ const DepositModal: React.FC = () => {
...
+              {status === 'connected' && !isOnWrongNetwork ? (
+                <AlertInline
+                  label={t('modal.deposit.statusBlockchain')}
+                  mode="success"
+                />
+              ) : (
+                <ConnectButton
+                  onClick={() => {
+                    close('deposit');
+                    open('network');
```
Fabricevladimir: This is something that's slightly off.
1. It would be nice if the `Deposit` modal could immediately pop up again once the user has connected to the proper network, but I'm not sure there's a clean and easy way to do it without a headache of refs since the wallet connection doesn't provide a callback. (not requesting any changes btw)

2. When the wallet is not connected, the `Network` modal goes all kinds of wonky because it doesn't have an address to display, nor a network to direct the user to. We either have to modify the modal or throw the login one before the wrong network one? What do you think?

Fabricevladimir (commented):
Thanks, Rakesh

Fabricevladimir (commented):
Thanks, Rakesh

Fabricevladimir:
Also, the dreaded Crowdin mess happened again. There seem to be keys that were renamed/added/removed, sigh.

packages/web-app/src/containers/transactionModals/DepositModal.tsx
RakeshUP: I've changed it to show the login modal if the user's wallet is not connected

RakeshUP: Passing ref's is a headache. 

But there's a way. We should try to store all kinds of navigation checkpoints right in the URL. Like when opening the `Deposit` modal, should change the URL and when we try to open `Network` modal, we add that in the URL(without removing Deposit), and once `Network` is closed, we could display the `Deposit` modal again. 

This pattern is followed in many apps that have to go through the login flow upon entering some URL and come back to the original URL the user entered

Fabricevladimir (approved):

PR #917, opened 2023-07-11
fix: APP-2253 removed scroll from modals and bottom sheet components

babarella:
## Description

When popup modals are displayed, background page is still scrollable.

Task: [APP-2253](https://aragonassociation.atlassian.net/browse/APP-2253)

## Type of change

## Checklist:

[APP-2253]: https://aragonassociation.atlassian.net/browse/APP-2253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/ui-components/src/components/backdrop/backdrop.tsx
+++ b/packages/ui-components/src/components/backdrop/backdrop.tsx
@@ -29,13 +29,6 @@ export const Backdrop: React.FC<BackdropProps> = ({
 
   useEffect(() => {
     const html = document.querySelector('html');
-    return () => {
-      if (html) html.style.overflow = 'auto';
-    };
-  }, []);
-
-  useEffect(() => {
-    const html = document.querySelector('html');
     if (!html) return;
     html.style.overflow = visible ? 'hidden' : 'auto';
   }, [visible]);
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/backdrop/backdrop.tsx
```diff
@@ -26,6 +26,34 @@ export const Backdrop: React.FC<BackdropProps> = ({
   ...props
 }) => {
   // TODO:Implement Backdrop to use as wrapper
+
+  useEffect(() => {
+    const html = document.querySelector('html');
+    return () => {
+      if (html) html.style.overflow = 'auto';
+    };
+  }, []);
```
RakeshUP: I think this useEffect block is interfering with the one below. I was able to prevent scrolling only when I commented out this block of code. Could you check if that's right?

RakeshUP (changes requested):
I'm not able to test this one. It could be due to the useEffect block that I've mentioned in the comment. Seems like two useEffects's are racing to set `overflow` on the html tag

Fabricevladimir:
@babarella, Can you take a look at Radix UI? We're using their headless Dialog component, which _should_ by default render the content underneath inert/unscrollable

babarella:
> @babarella, Can you take a look at Radix UI? We're using their headless Dialog component, which _should_ by default render the content underneath inert/unscrollable

Applied radix overlay for modals, thanks for suggestion.

RakeshUP (approved):
Thanks!

PR #941, opened 2023-07-26
Fix: Edit page setting issues - APP-2297

sepehr2github:
## Description

- Fix proposal threshold decimals issue
- Fix proposal threshold unexpected reset
- remove null.dao.eth from mobile navbar menu for L2 daos
- Fix edit settings 404 in for /settings/new-proposal
- Add logo resolver to daoDeatils hook
- Fix Update logo issue for setting page & edit setting

Task: [APP-2297](https://aragonassociation.atlassian.net/browse/APP-2297) - [APP-2350](https://aragonassociation.atlassian.net/browse/APP-2350)

## Type of change

## Checklist:

[APP-2297]: https://aragonassociation.atlassian.net/browse/APP-2297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APP-2343]: https://aragonassociation.atlassian.net/browse/APP-2343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APP-2350]: https://aragonassociation.atlassian.net/browse/APP-2350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/containers/compareSettings/index.tsx
```diff
@@ -36,7 +36,9 @@ const CompareSettings: React.FC = () => {
   const [selectedButton, setSelectedButton] = useState<Views>('new');
 
   const daoAddressOrEns =
-    toDisplayEns(daoDetails?.ensDomain) ?? daoDetails?.address;
+    toDisplayEns(daoDetails?.ensDomain) === ''
```
Fabricevladimir, line 39: This is fine, but why not use the or operator?

Fabricevladimir (changes requested):
Nice work, but unfortunately this dies at the propose setting step

src/containers/compareSettings/index.tsx
sepehr2github: because toDisplayEns doesn't return null or undefined

src/containers/compareSettings/compareMetadata.tsx
```diff
@@ -71,7 +66,7 @@ export const CompareMetadata: React.FC<CompareMetadataProps> = ({
         <Dt>{t('labels.logo')}</Dt>
         <Dd>
           {view === 'new' ? (
-            URL_PATTERN.test(daoLogo) ? (
+            URL_PATTERN.test(daoLogo) || daoLogo?.startsWith?.('blob') ? (
```
Fabricevladimir, line 69: Would be nice to add this starts with test into the regex itself.

src/containers/defineMetadata/index.tsx
```diff
@@ -195,7 +195,7 @@ const DefineMetadata: React.FC<DefineMetadataProps> = ({
 
             try {
               // in case url does not need to be created
-              if (URL_PATTERN.test(value)) {
+              if (URL_PATTERN.test(value) || value?.startsWith?.('blob')) {
```
Fabricevladimir, line 200: See comment on `compareMetadata.tsx`

src/hooks/useDaoDetails.tsx
```diff
@@ -39,6 +40,25 @@ async function fetchDaoDetails(
 
   // Note: SDK doesn't support ens names in L2 chains so we need to resolve the address first
   const daoDetails = await client.methods.getDao(daoAddressOrEns.toLowerCase());
+  const avatar = daoDetails?.metadata.avatar;
+  if (avatar)
+    if (avatar && typeof avatar !== 'string') {
```
Fabricevladimir: The check for whether the avatar exits seems to be duplicated for no reason

src/hooks/useDaoDetails.tsx
```diff
@@ -39,6 +40,25 @@ async function fetchDaoDetails(
...
+      daoDetails!.metadata.avatar = URL.createObjectURL(avatar);
```
Fabricevladimir: The non-null assertion is unnecessary because the condition assumes that avatar, a nested property of `daoDetails`, exists.

src/hooks/useDaoDetails.tsx
```diff
@@ -39,6 +40,25 @@ async function fetchDaoDetails(
...
+    } else if (/^ipfs/.test(avatar) && client) {
+      try {
+        const cid = resolveIpfsCid(avatar);
+        const ipfsClient = client.ipfs.getClient();
+        const imageBytes = await ipfsClient.cat(cid); // Uint8Array
+        const imageBlob = new Blob([imageBytes] as unknown as BlobPart[]);
+
+        daoDetails!.metadata.avatar = URL.createObjectURL(imageBlob);
```
Fabricevladimir: Same thing here

src/hooks/useDaoDetails.tsx
```diff
@@ -39,6 +40,25 @@ async function fetchDaoDetails(
...
+      } catch (err) {
+        console.warn('Error resolving DAO avatar IPFS Cid', err);
+      }
+    } else {
+      daoDetails!.metadata.avatar = avatar;
```
Fabricevladimir: And here as well.

src/containers/editSettings/multisig.tsx
```diff
@@ -132,22 +127,34 @@ export const EditMsSettings: React.FC<EditMsSettingsProps> = ({daoDetails}) => {
     resourceLinks,
   ]);
 
-  // metadata setting changes
-  const isMetadataChanged =
-    daoDetails?.metadata.name &&
-    (daoName !== daoDetails.metadata.name ||
+  const isMetadataChanged = useMemo(() => {
+    if (!daoDetails?.metadata.name || !daoName || daoName === '') return false;
```
Fabricevladimir: Something like this might be a bit more concise.
```suggestion
    if (!daoDetails?.metadata.name || !daoName?.trim()) return false;
```

src/containers/editSettings/multisig.tsx
```diff
@@ -193,17 +200,21 @@ export const EditMsSettings: React.FC<EditMsSettingsProps> = ({daoDetails}) => {
 
   useEffect(() => {
     setValue('isMetadataChanged', isMetadataChanged);
-    setValue('areSettingsChanged', isGovernanceChanged);
+    setValue('areSettingsChanged', isMetadataChanged || isGovernanceChanged);
```
Fabricevladimir, line 203: Aha so previously only governance changes were being tracked! Good catch

Fabricevladimir (commented):
Thanks, Sepher. Nice job dealing with these nasty little issues! 
One final thing, considering going directly to settings/new-proposal only shows the page with no data, can you add a check to make sure that when the user goes directly to that page they get redirected to the Edit Settings page if possible? If not, the next best thing would be redirected to the Settings page. What do you think?

src/containers/defineMetadata/index.tsx
sepehr2github: The URL_PATTERN Regex is using for links list as well so I think keeping it link this will be better

src/containers/compareSettings/compareMetadata.tsx
sepehr2github: The URL_PATTERN Regex is using for links list as well so I think keeping it link this will be better

src/pages/proposeSettings.tsx
```diff
@@ -245,88 +246,115 @@ const ProposeSettingWrapper: React.FC<Props> = ({
...
-              ? VotingMode.VOTE_REPLACEMENT
-              : VotingMode.STANDARD,
-          },
-        };
-        setValue('actions', [metadataAction, voteSettingsAction]);
-      } else {
-        const multisigSettingsAction: ActionUpdateMultisigPluginSettings = {
-          name: 'modify_multisig_voting_settings',
+        if (daoDetails && !daoName)
+          navigate(generatePath(Settings, {network, dao: daoDetails?.address}));
```
Fabricevladimir, line 288: Sorry, this is on me, can you switch that from settings to edit settings page instead?

Fabricevladimir (approved):

PR #957, opened 2023-08-03
WIP: Feat: Offchain DAO creation

selankon:
## Description

This PR is developed on the vocdoni `offChain` plugin context.  

The branch implement changes on the DAO creation form to add the option to create `offChain` proposals with a executive committee. 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have selected the correct base branch.


package.json
```diff
@@ -31,8 +31,10 @@
     "@tiptap/pm": "^2.0.3",
     "@tiptap/react": "^2.0.3",
     "@tiptap/starter-kit": "^2.0.3",
-    "@vocdoni/react-providers": "^0.1.8",
-    "@vocdoni/sdk": "^0.3.0",
+    "@vocdoni/react-providers": "^0.1.7",
+    "@vocdoni/sdk": "^0.1.1",
```
elboletaire: Both need upgrade:
- react providers should be at least 0.1.9
- sdk should be at least 0.3.1

src/components/addCommitteeMembers/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }, [address, append, ensName, wallets]);
+
+  // setTimeout added because instant trigger not working
+  const handleAddWallet = () => {
+    append({address: '', ensName: '', amount: 1});
+    setTimeout(() => {
+      trigger(`committee.${controlledFields.length}`);
+    }, 50);
+  };
```
elboletaire, line 53: In a case like this, it's usually better to use an affect for the second action (in this case listening for changes in the addresses field). Doing so, you may not need the timeout anymore.

Still, is not a terrible solution to have a timeout this way

elboletaire: Kinda confusing to see the timeout below set to zero when this one is set to 50 though.

src/components/executionExpirationtime/index.tsx
```diff
@@ -0,0 +1,270 @@
...
+      }
+      trigger([
+        'executionExpirationMinutes',
+        'executionExpirationHours',
+        'executionExpirationDays',
+      ]);
+      onChange(e);
+    },
+    [getValues, setValue, trigger]
+  );
```
elboletaire, line 125: Considering how are you using these methods (`handleDaysChanged`, `handleHoursChanged` and `handleMinutesChanged`), I don't think you need to use `useCallback` at all. These could probably be defined as regular methods, since you're not using state values in the methods (you're using state methods, but not state values). `useCallback` is used mostly when you need to access state values in the function that could change, but I don't see this being the case.

src/components/filteredAddressList/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+    (wallet: TokenVotingWalletField | MultisigWalletField) => {
+      if (searchValue !== '') {
+        const re = new RegExp(searchValue, 'i');
+
+        return wallet?.address?.match(re) || wallet?.ensName.match(re);
+      }
+      return true;
+    },
+    [searchValue]
+  );
```
elboletaire, line 37: In this case it totally makes sense to use a `useCallback` because you're using a state value (`searchValue`) in your function. No need to do anything here, just adding a comment for extending my explanation above.

src/containers/goLive/committee.tsx
```diff
@@ -0,0 +1,126 @@
...
+import {
+  Dd,
+  DescriptionListContainer,
+  Dl,
+  Dt,
+} from '../../components/descriptionList';
+import {IconFeedback, Link} from '@aragon/ods';
+import {CHAIN_METADATA} from '../../utils/constants';
+import React from 'react';
+import CommitteeAddressesModal from '../committeeAddressesModal';
```
elboletaire, line 15: [minor] weird import sorting

src/context/createDao.tsx
```diff
@@ -39,6 +39,17 @@ import {Dashboard} from 'utils/paths';
...
+
+import {
+  OffchainVotingClient,
+  OffchainVotingPluginInstall,
+} from '@vocdoni/offchain-voting';
+import {BigNumber} from '@ethersproject/bignumber';
+import {GaslessPluginVotingSettings} from '@vocdoni/offchain-voting/dist/js-client/src/types';
+import {ErrTokenAlreadyExists} from '@vocdoni/sdk';
+import {useCensus3CreateToken} from '../hooks/useCensus3';
+
```
elboletaire, line 52: [minor] Most, if not all, of these imports should be above the already existing ones. Again, weird import sorting.

Note there's no need to alphabetically sort them, but at least they should be properly grouped (first external imports, then `../`, finally `./`).

src/context/createDao.tsx
```diff
@@ -143,6 +154,47 @@ const CreateDaoProvider: React.FC = ({children}) => {
...
+        votingSettings: vocdoniVotingSettings,
+        ...((tokenType === 'governance-ERC20' || // token can be used as is
+          tokenType === 'ERC-20') && // token can/will be wrapped
+        !isCustomToken // not a new token (existing token)
+          ? {useToken: getErc20PluginParams()}
+          : {newToken: getNewErc20PluginParams()}),
+      };
+    },
+    [getValues]
+  );
```
elboletaire: I'm pretty sure this callback does not need to be a callback either.

src/context/createDao.tsx
```diff
@@ -286,20 +339,27 @@ const CreateDaoProvider: React.FC = ({children}) => {
...
-            network
-          );
-
-        plugins.push(tokenVotingPlugin);
+        if (votingType === 'offChain') {
+          const params = getOffchainPluginInstallParams(votingSettings);
+          const offChainPlugin =
+            OffchainVotingClient.encoding.getPluginInstallItem(params, network);
+          plugins.push(offChainPlugin);
+        } else {
```
elboletaire: Avoid using `else` when possible. In this case you could `break` the execution at the end of the if, and skip the usage of the `else` (and all its related indentation).

Not a required change, since it won't affect in code's efficiency, but would help following coders to read it better.

src/components/addCommitteeMembers/row.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+      // check if address is valid
+      let validationResult = await validateWeb3Address(
+        web3Address,
+        t('errors.required.walletAddress'),
+        t
+      );
+
+      if (validationResult && validationResult !== true)
+        return validationResult;
```
elboletaire, line 54: For consistency, use brackets here (`{}`) or move the return to the same line, so it's a one-liner.

src/pages/createDAO.tsx
```diff
@@ -228,6 +238,40 @@ const CreateDAO: React.FC = () => {
...
+    // if multisig
+    if (membership === 'multisig') {
+      if (!['multisig', 'anyone'].includes(eligibilityType)) {
+        return false;
+      }
+      return true;
+    } else {
+      return !errors.eligibilityTokenAmount;
+    }
+  }, [eligibilityType, errors.eligibilityTokenAmount, membership]);
```
elboletaire, line 273: No need to use `useMemo` here.

elboletaire (changes requested):
Some of the `useMemo` and `useCallbacks` used are not necessary. Ensure you try any changes to these methods, since I can only speak for the code I've seen... I tried to check all the affected methods to ensure I'm not saying it wrong, but I could be wrong in some of them... so double-check it please.

package.json
selankon: As the packages are not yet on npmjs i didn't push the update awaiting the last version before merge to develop. But thanks to remember it!

src/components/addCommitteeMembers/index.tsx
selankon: This code cames from https://github.com/aragon/app/blob/develop/src/components/addWallets/index.tsx#L73-L85 and on the actual develop head, the second timeout have timeout also. I add it manually here 2ea8af38

src/components/executionExpirationtime/index.tsx
selankon: This code cames from https://github.com/aragon/app/blob/develop/src/containers/configureCommunity/index.tsx#L54-L127 , implemented as `useCallback` so I will let as is

src/containers/goLive/committee.tsx
selankon: mmm the import sorting is done by prettier rules and are automatically executed when document is saved. So i cannot change it, the rules are defined by the configuration files of the repository.

src/context/createDao.tsx
selankon: Same as above, I am not sure I can change this manually

src/context/createDao.tsx
selankon: This is a callback because this code is inspired from a similar function from the Aragon app code: https://github.com/aragon/app/blob/develop/src/context/createDao.tsx#L152-L155 .

src/components/addCommitteeMembers/row.tsx
selankon: fix 26c7ced8

src/context/createDao.tsx
selankon: Fix 26c7ced8

src/containers/goLive/committee.tsx
elboletaire: I'd had said prettier does sort as I said (we're using it in ui-components and ui-scaffold following such rules). Are you sure you have the sort on save actually enabled? If so, kinda weird rules IMHO 😅

selankon:
Closed in favour to https://github.com/aragon/app/pull/1074

PR #959, opened 2023-08-03
Feature: APP-2402 - Token Transfers For Base Networks

Fabricevladimir:
## Description

- adds support for token transfer fetching for Base mainnet and testnet
- adds feature flags for networks in DAO creation

Note: Please review _after_ #958 since its branch is the base branch for this PR

Task: [APP-2402](https://aragonassociation.atlassian.net/browse/APP-2402)

## Type of change

## Checklist:

[APP-2402]: https://aragonassociation.atlassian.net/browse/APP-2402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/token/domain/alchemy-token.ts
```diff
@@ -0,0 +1,12 @@
+export type AlchemyTransfer = {
```
cgero-eth: File name should be `alchemy-transfer.ts`

src/services/token/domain/covalent-token.ts
```diff
@@ -34,3 +34,25 @@ export type CovalentTokenBalance = {
...
 };
+
+export type CovalentTokenTransfer = {
+  address: string;
+  items: Array<{
+    transfers: CovalentTransferInfo[];
+  }>;
+};
+
+export type CovalentTransferInfo = {
```
cgero-eth: I'd move these two types on a new `covalent-transfer.ts` file, what do you think?

src/services/token/query-keys.ts
```diff
@@ -6,20 +6,22 @@ import type {
...
   ],
-};
-
-export const tokenBalancesQueryKeys = {
-  address: (params: IFetchTokenBalancesParams): QueryKey => [
+  balances: (params: IFetchTokenBalancesParams): QueryKey => [
     TokenQueryItem.BALANCES,
     params,
   ],
+  transfers: (params: IFetchTokenBalancesParams): QueryKey => [
```
cgero-eth:
```suggestion
  transfers: (params: IFetchTokenTransfersParams): QueryKey => [
```

src/services/token/token-service.ts
```diff
@@ -217,6 +228,116 @@ class TokenService {
...
+
+    // check if network is supported
+    if (!networkId) {
+      console.info(
+        `fetchCovalentErc20Deposits - network ${network} not supported by Covalent`
+      );
+      return null;
+    }
+    // fetch all balances
+    const assetsBalances = await this.fetchTokenBalances({
```
cgero-eth: By making these api calls here we are not using the react-query cache, this means that we will be calling the `fetch-token-balance` api regardless if we already fetched on the UI. We could fix this by passing the token balances as a parameter to this function.

src/hooks/useDaoTransfers.tsx
```diff
@@ -1,192 +1,131 @@
-import {useReactiveVar} from '@apollo/client';
```
cgero-eth, line 1: This hook looks much cleaner now, well done!

cgero-eth (commented):
Thanks for the changes Fabrice!

src/services/token/domain/covalent-token.ts
Fabricevladimir: Might as well, yup

src/services/token/token-service.ts
Fabricevladimir: Good catch 👍🏾

cgero-eth (approved):

PR #964, opened 2023-08-04
fix: APP-2405 proposals multiplication fix

babarella:
## Description

The Miracle of the Multiplication of the Proposal (at the Governance page)

Task: [APP-2405 ](https://aragonassociation.atlassian.net/browse/APP-2405 )

## Type of change

## Checklist:


Fabricevladimir:
While `lodash` fixes the issue, I think an even better approach might be to actually refactor that proposal fetching hook to use `react-query` which will fix this issue on top of providing all the caching goodness. What do you think?

babarella:
> While `lodash` fixes the issue, I think an even better approach might be to actually refactor that proposal fetching hook to use `react-query` which will fix this issue on top of providing all the caching goodness. What do you think?

I think that this solution is good because the previous usage was actually outside of `useDaoDetails` hook.
Would be better I think to go with such fix and maybe plan a fully-fledged refactoring of this hook as a separate ticket in the future.
Ref: `setDisplayedProposals(prev => [...(prev || []), ...proposals]);`

Fabricevladimir (approved):
Approving, but please do add a todo or create a task for the react query refactor

PR #966, opened 2023-08-07
Fix: APP-2344 - Adds proposal creation eligibility to Edit Settings

Fabricevladimir:
## Description

-  Adds proposal creation eligibility to Edit Settings

Task: [APP-2344](https://aragonassociation.atlassian.net/browse/APP-2344)

## Type of change

## Checklist:

[APP-2344]: https://aragonassociation.atlassian.net/browse/APP-2344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/editSettings/multisig.tsx
+++ b/src/containers/editSettings/multisig.tsx
@@ -1,10 +1,10 @@
-import {DaoDetails, MultisigVotingSettings} from '@aragon/sdk-client';
 import {
   AlertInline,
   ButtonText,
   IconGovernance,
   ListItemAction,
 } from '@aragon/ods';
+import {DaoDetails, MultisigVotingSettings} from '@aragon/sdk-client';
 import React, {useCallback, useEffect, useMemo} from 'react';
 import {
   useFieldArray,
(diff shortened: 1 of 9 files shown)


src/containers/editSettings/multisig.tsx
```diff
@@ -151,6 +157,13 @@ export const EditMsSettings: React.FC<EditMsSettingsProps> = ({daoDetails}) => {
     return multisigMinimumApprovals !== settings.minApprovals;
   }, [multisigMinimumApprovals, settings.minApprovals]);
 
+  let daoEligibleProposer: MultisigProposerEligibility = formEligibleProposer;
+  if (Object.keys(settings).length !== 0 && settings.constructor === Object) {
```
cgero-eth, line 161: Why is this check needed here? Is it because the `settings` object might be undefined? In that case I would rather change the type of settings and introduce a null check here to avoid errors in the future:
```
settings = data as MultisigVotingSettings | undefined;

if (settings != null) { ... }
```

cgero-eth (commented):
Thanks @Fabricevladimir, I left only a small comment

src/containers/editSettings/multisig.tsx
Fabricevladimir: I have to check for object properties because for whatever reason the `data` from the `usePluginSettings` hook is being set to an empty object by default. The idea is to change the type to `undefined` in there, but I think it's worth its own chore considering the change will affect 9-10 other files.

cgero-eth: Oh I see, sure let's tackle it on another PR 👌🏻

cgero-eth (approved):

PR #972, opened 2023-08-09
Add wizard skip step feature

selankon:
## Description

It adds a `skipStep` feature on the wizard. 

This is due to the necessity to have conditional steps. For example, a step that is shown depending of previous choices on precedent steps. 

The `Step` `hideWizard` attribute doesn't fit this necessity because the step was not count on the total steps top bar, but is shown anyway when next button clicked. This is because the `hideWizard` property objective is to not to count a step but show it anyway, for example, when the last step is used to review all form data. 

Now, if you add the `skipStep={true}` the step won't be count and neither will be shown. 

## Type of change

## Checklist:

### DIFF
--- a/src/components/fullScreenStepper/fullScreenStepper.tsx
+++ b/src/components/fullScreenStepper/fullScreenStepper.tsx
@@ -50,9 +50,7 @@ export const FullScreenStepper: React.FC<FullScreenStepperProps> = ({
   navLabel,
   returnPath,
 }) => {
-  const skipSteps = useMemo(() => {
-    return children.filter(child => child.props.skipStep !== true);
-  }, [children]);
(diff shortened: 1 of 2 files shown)


src/components/fullScreenStepper/fullScreenStepper.tsx
```diff
@@ -50,11 +50,15 @@ export const FullScreenStepper: React.FC<FullScreenStepperProps> = ({
   navLabel,
   returnPath,
 }) => {
+  const skipSteps = useMemo(() => {
+    return children.filter(child => child.props.skipStep !== true);
```
sepehr2github: Thanks for the changes, But I don't think using a memo here is so useful

sepehr2github (approved):
thanks, Seems good to go, Just please don't forget to sign your previous commits

PR #980, opened 2023-08-11
Refactor wallet connection to make as reusable as possible

sepehr2github:
## Description

- Move wallet login modal to loginMenu context

Task: [APP-2392](https://aragonassociation.atlassian.net/browse/APP-2392)

## Type of change

## Checklist:

[APP-2392]: https://aragonassociation.atlassian.net/browse/APP-2392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/context/loginMenu.tsx
```diff
@@ -0,0 +1,120 @@
+import {LoginRequired} from 'containers/walletMenu/LoginRequired';
```
RuggeroCino, line 1: This currently triggers the wallet connection modal on every page if the user is not connected, I think we should keep this logic only for the protected routes, right?

cgero-eth, line 1: This currently triggers the wallet connection modal on every page if the user is not connected, I think we should keep this logic only for the protected routes, right?

Instead of creating a new context provider, I would maybe move the open/close logic to the global-modal provider, in that way the login-dialog can be opened or closed by any other component. The logic of checking if the user is connected and, otherwise, navigate to back to previous route is specific to the protected-route component. In all other cases (e.g. the delegate flow), we only need to open the login-dialog in case the user clicks a button (delegate tokens) and he is not connected.

sepehr2github: Hey. Thanks for the comment. The Context was a request from James to have a structure for upcoming development. btw the issue resolved

cgero-eth: The dialog isn't displayed now on unprotected routes, well done 👏🏻 But now I need to click on "Connect wallet now" twice in order to display the wallet-connect dialog, could you check this?

sepehr2github:
Closed due to the lack of enough clarity about the corresponding task

PR #986, opened 2023-08-16
Fix: SCC input field validations

RakeshUP:
Task: [APP-2403](https://aragonassociation.atlassian.net/browse/APP-2403)

## Type of change

## Checklist:

[APP-2403]: https://aragonassociation.atlassian.net/browse/APP-2403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/executionWidget/actions/sccExecutionWidget.tsx
+++ b/src/components/executionWidget/actions/sccExecutionWidget.tsx
@@ -8,6 +8,7 @@ import {Input} from 'utils/types';
 export const SCCExecutionCard: React.FC<{
   action: any;
 }> = ({action}) => {
+  console.log('SCCExecutionCard', action);
   return (
     <AccordionMethod
       type="execution-widget"
--- a/src/components/executionWidget/actions/sccExecutionWidget.tsx
+++ b/src/components/executionWidget/actions/sccExecutionWidget.tsx
@@ -8,7 +8,6 @@ import {Input} from 'utils/types';
 export const SCCExecutionCard: React.FC<{
   action: any;
 }> = ({action}) => {
-  console.log('SCCExecutionCard', action);
   return (
     <AccordionMethod
       type="execution-widget"
--- a/src/components/fullScreenStepper/step.tsx
+++ b/src/components/fullScreenStepper/step.tsx
@@ -18,6 +18,6 @@ export type StepProps = {
   onNextButtonDisabledClicked?: () => void;
 };
 
-export const Step: React.FC<StepProps> = props => {
-  return <>{props.children}</>;
+export const Step: React.FC<StepProps> = ({children}) => {
+  return <>{children}</>;
 };
--- a/src/components/fullScreenStepper/step.tsx
+++ b/src/components/fullScreenStepper/step.tsx
@@ -18,6 +18,6 @@ export type StepProps = {
   onNextButtonDisabledClicked?: () => void;
 };
 
-export const Step: React.FC<StepProps> = ({children}) => {
-  return <>{children}</>;
+export const Step: React.FC<StepProps> = props => {
+  return <>{props.children}</>;
 };
--- a/src/containers/actionBuilder/scc/index.tsx
+++ b/src/containers/actionBuilder/scc/index.tsx
@@ -59,28 +59,30 @@ const SCCAction: React.FC<ActionIndex & {allowRemove?: boolean}> = ({
         <FormItem className="space-y-3 rounded-b-xl">
           {actionData.inputs?.length > 0 ? (
             <div className="pb-1.5 space-y-2">
-              {(actionData.inputs as Input[]).map((input, index) => (
-                <div key={input.name}>
-                  <div className="text-base font-bold text-ui-800 capitalize">
-                    {input.name}
-                    <span className="ml-0.5 text-sm normal-case">
-                      ({input.type})
-                    </span>
+              {(actionData.inputs as Input[])
+                .filter(input => input.type)
+                .map((input, index) => (
+                  <div key={input.name}>
+                    <div className="text-base font-bold text-ui-800 capitalize">
+                      {input.name}
+                      <span className="ml-0.5 text-sm normal-case">
+                        ({input.type})
+                      </span>
+                    </div>
+                    <div className="mt-0.5 mb-1.5">
+                      <span className="text-ui-600 ft-text-sm">
+                        {input.notice}
+                      </span>
+                    </div>
+                    <ComponentForType
+                      key={input.name}
+                      input={input}
+                      functionName={actionData.name}
+                      formHandleName={`actions.${actionIndex}.inputs.${index}.value`}
+                      isValid={isValid}
+                    />
                   </div>
-                  <div className="mt-0.5 mb-1.5">
-                    <span className="text-ui-600 ft-text-sm">
-                      {input.notice}
-                    </span>
-                  </div>
-                  <ComponentForType
-                    key={input.name}
-                    input={input}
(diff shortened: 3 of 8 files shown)


src/utils/validators.ts
```diff
@@ -185,21 +191,22 @@ export const alphaNumericValidator = (
  * @param errors List of fields with errors
  * @returns Whether the screen is valid
  */
-export function actionsAreValid(
+export async function actionsAreValid(
```
Fabricevladimir, line 194: This function is awkward because everything is negated!

src/containers/smartContractComposer/components/inputForm.tsx
```diff
@@ -272,16 +234,9 @@ export const ComponentForType: React.FC<ComponentForTypeProps> = ({
...
           name={formName}
-          rules={{
-            required: t('errors.required.walletAddress') as string,
-            validate: value => validateAddress(value),
-          }}
-          render={({
-            field: {name, value, onBlur, onChange},
-            fieldState: {error},
-          }) => (
+          render={({field: {name, value, onBlur, onChange}}) => (
```
Fabricevladimir, line 263: Can you copy this controller with the address validation for the previous case? Since we know how to validate addresses and not really encodedData, we should make use of that. As it is right now, addresses don't show an error when they're invalid, only the whole action shows a required error.

src/components/fullScreenStepper/step.tsx
```diff
@@ -17,6 +17,6 @@ export type StepProps = {
   onNextButtonDisabledClicked?: () => void;
 };
 
-export const Step: React.FC<StepProps> = ({children}) => {
-  return <>{children}</>;
+export const Step: React.FC<StepProps> = props => {
+  return <>{props.children}</>;
```
Fabricevladimir: Why the change here? lol

src/containers/actionBuilder/scc/index.tsx
```diff
@@ -53,27 +59,36 @@ const SCCAction: React.FC<ActionIndex & {allowRemove?: boolean}> = ({
...
-                    input={input}
-                    functionName={actionData.name}
-                    formHandleName={`actions.${actionIndex}.inputs.${index}.value`}
-                  />
-                </div>
-              ))}
+                ))}
+              {!isValid && (
+                <AlertInline
+                  label={t('newProposal.configureActions.alertCritical')}
```
Fabricevladimir, line 88: This is going to need a different string because if an address field is invalid for example, this error is incorrect.

src/containers/actionBuilder/scc/index.tsx
```diff
@@ -20,6 +22,10 @@ const SCCAction: React.FC<ActionIndex & {allowRemove?: boolean}> = ({
     name: [`actions.${actionIndex}`],
   });
   const {alert} = useAlertContext();
+  const {network} = useNetwork();
+  const [isValid, setIsValid] = useState(true);
+
+  validateSCCAction(actionData, network).then(res => setIsValid(res));
```
Fabricevladimir, line 28: Shouldn't this be inside a useEffect? If not, it will run on every render.

src/containers/smartContractComposer/components/inputForm.tsx
```diff
@@ -309,18 +264,21 @@ export const ComponentForType: React.FC<ComponentForTypeProps> = ({
...
             <NumberInput
               name={name}
               onBlur={onBlur}
               onChange={onChange}
               placeholder="0"
               includeDecimal
               disabled={disabled}
-              mode={error?.message ? 'critical' : 'default'}
+              mode={
```
Fabricevladimir, line 325: Why not use the library provided validation here as well?

src/containers/proposalStepper/index.tsx
```diff
@@ -50,16 +50,20 @@ const ProposalStepper: React.FC<ProposalStepperType> = ({
...
+  const [isActionsValid, setIsActionsValid] = useState(false);
 
   const [formActions] = useWatch({
     name: ['actions'],
-    control,
   });
 
   const {errors, dirtyFields} = useFormState({
     control,
   });
```
Fabricevladimir, line 63: I'm assuming you want this running on every render? Why not put it in a useEffect?

Fabricevladimir (changes requested):
Thanks, Rakesh. Left a couple of comments. 

Also, ran into an issue where the data of a previous unadded action isn't being cleared when a new action is selected. Not sure exactly where it's coming from.

src/containers/actionBuilder/scc/index.tsx
RakeshUP: react-hook-form mutates arrays and objects. So useEffect won't run when `actionData` array fields change. 

We either do the validations on the render (or) use something like [use-deep-compare-effect](https://github.com/kentcdodds/use-deep-compare-effect). I think in this component the most re-renders would be due to change in `actionData` and thus the validations would run most of the time. So I chose to put it on the render

src/containers/proposalStepper/index.tsx
RakeshUP: Same as https://github.com/aragon/app/pull/986#discussion_r1319674067

src/containers/smartContractComposer/components/inputForm.tsx
RakeshUP: I'm not aware of any validations for BigNumbers. Could you let me know if you know any?

The only rationale behind this check here is to make sure it's non-empty

RakeshUP:
@Fabricevladimir I'll get the new string after Evan/Selim return from the offsite next week.  

Regarding the inputs not being reset, it can happen when the function and input names are the same for multiple functions(due to function overloading). I'll check if this is the case and add some index/unique property to the form keys

src/containers/smartContractComposer/components/inputForm.tsx
Fabricevladimir: No, what I meant was having the validation done in the controller and displaying the error under the field itself, it's the same thing about the previous address comment.

Using this Goerli contract `0xF7CdB7bD236615f90246b9607D553CcB6E554c5D`, if you try to withdraw Ether, add the action, and then remove the last character of the receiver address, the input gets the error state, but an incorrect message is being shown for the entirety of the action.

image

Fabricevladimir:
> Regarding the inputs not being reset, it can happen when the function and input names are the same for multiple functions(due to function overloading). I'll check if this is the case and add some index/unique property to the form keys

Using the same contract, after filling out the `withdrawEther` action, instead of adding the action, switch to any of the withdraw actions and you'll see that the amount and the receiver from the `withdrawEther` function is being filled in. I think this might just be a case of not clearing the current potential action when we select another function.

src/components/executionWidget/actions/sccExecutionWidget.tsx
```diff
@@ -8,6 +8,7 @@ import {Input} from 'utils/types';
 export const SCCExecutionCard: React.FC<{
   action: any;
 }> = ({action}) => {
+  console.log('SCCExecutionCard', action);
```
Fabricevladimir: You forgot to remove this one!

Fabricevladimir (approved):

src/components/executionWidget/actions/sccExecutionWidget.tsx
jamesej: I fixed this to get this one merged

PR #987, opened 2023-08-16
Fix: APP-2411 - Prioritize ENS in wallet fields

Fabricevladimir:
## Description

Please include a summary of the change and be sure you follow the contributions rules we do provide [here](./CONTRIBUTIONS.md)

Task: [APP-2411](https://aragonassociation.atlassian.net/browse/APP-2411)

## Type of change

## Checklist:

[APP-2411]: https://aragonassociation.atlassian.net/browse/APP-2411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/addWallets/index.tsx
+++ b/src/components/addWallets/index.tsx
@@ -5,7 +5,7 @@ import {
   IconMenuVertical,
   ListItemAction,
 } from '@aragon/ods';
-import React, {useCallback, useEffect, useRef} from 'react';
+import React, {useEffect, useRef} from 'react';
 import {useFieldArray, useFormContext, useWatch} from 'react-hook-form';
 import {useTranslation} from 'react-i18next';
 import styled from 'styled-components';
@@ -22,7 +22,7 @@ import Row from './row';
 const AddWallets: React.FC = () => {
   const {t} = useTranslation();
   const {alert} = useAlertContext();
-  const appendConnectedAddress = useRef(true);
+  const appendConnectedAddress = useRef<boolean>(true);
 
   const {network} = useNetwork();
   const {address} = useWallet();
@@ -53,15 +53,12 @@ const AddWallets: React.FC = () => {
       appendConnectedAddress.current === true
     ) {
       append({address, amount: '1', ensName});
+      appendConnectedAddress.current = false;
+      setTimeout(() => {
+        trigger('wallets.0');
+      }, 50);
     }
-  }, [address, append, controlledFields?.length, ensName]);
-
-  const isConnectedAddress = useCallback(
-    (index: number) =>
-      controlledFields[index].address.toLowerCase() === address?.toLowerCase(),
-
-    [address, controlledFields]
-  );
+  }, [address, append, controlledFields?.length, ensName, trigger]);
 
   const resetDistribution = () => {
     controlledFields.forEach((_, index) => {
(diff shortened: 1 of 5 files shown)


src/components/addWallets/index.tsx
```diff
@@ -36,10 +47,21 @@ const AddWallets: React.FC = () => {
...
 
   useEffect(() => {
-    if (address && !wallets) {
-      append({address, amount: '1'});
+    if (
+      address &&
+      controlledFields?.length === 0 &&
+      appendConnectedAddress.current === true
+    ) {
+      append({address, amount: '1', ensName});
```
cgero-eth, line 55: What if we set the `appendConnectedAddress` to `false` right after this line like:
```
append({address, amount: '1', ensName})
appendConnectedAddress.current = false
```
Would it work as well? In that way we'll still only append the wallet once and we won't need the `isConnectedAddress` check and the new logic on the `handleDeleteRow` function.

src/pages/createDAO.tsx
```diff
@@ -113,6 +139,27 @@ export const CreateDAO: React.FC = () => {
...
+
+    if (prevFormChain.current !== formChain) {
+      prevFormChain.current = formChain;
+      refetchWalletsENS();
+    }
+
+    // intentionally setting only formChain as the dependency
+    // since this update needs to happen only on form network change
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [formChain]);
```
cgero-eth, line 161: Just curious about this one, does `formChain` changes depending on the user connected chain? Shouldn't it be based on the user selection on the first step?

cgero-eth (commented):
Thanks @Fabricevladimir, I only left two questions about the PR

src/components/addWallets/index.tsx
Fabricevladimir: That should work yes.

src/pages/createDAO.tsx
Fabricevladimir: `formChain` changes on the selection the user makes. Initially it *is* based on the user connected chain (or defaults to Ethereum Mainnet if the user is not connected), but afterwards it is only changed via the form itself. See [here](https://github.com/aragon/app/blob/d2281e84dde473c448a05e73fb402a502e8c6c34/src/pages/createDAO.tsx#L117) for more info. 
Here I'm merely checking to see if the user has selected a different network via the form on the first step and updating the ENS associated with each address.

cgero-eth: Oh okay I see, so the use case is that a user selects a chain for the DAO, adds some wallets and then goes back and changes the chain, correct? I'm asking because we might have the same issue in other forms of the application and we could consider fixing this directly on the wallet-input or wrapped-wallet-input components in another tech-debt ticket. Does it makes sense? Should I note this for another ticket?

cgero-eth (approved):

src/pages/createDAO.tsx
Fabricevladimir: As of now, DAO creation is really the only place the network changes, but having the refetching inside the component wrapper itself makes more sense and is cleaner and DRY-er. Feel free to add this to another ticket.

PR #1000, opened 2023-08-22
Feat: APP-2369 - Add entry points to delegation flow

cgero-eth:
## Description

- Display a delegate now button after successfully wrapping the ERC20 tokens to the government tokens;
- Display a voting-power-gating menu if the user has balance but cannot vote because he delegated his voting power
- Add `getVotingPower` / `getPastVotingPower` and relative hooks (`useVotingPower` / `usePastVotingPower`) to check for current and past voting power
- Make sure that the `wagmi` cache is updated after wrapping / unwrapping the tokens to display the correct dialog when opening the "delegateTokens" dialog (if query is not invalidated, the "no-tokens" view will be displayed intstead)

Task: [APP-2369](https://aragonassociation.atlassian.net/browse/APP-2369)

## Type of change

## Checklist:

[APP-2369]: https://aragonassociation.atlassian.net/browse/APP-2369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/delegationGatingMenu/delegationGatingMenu.tsx
+++ b/src/containers/delegationGatingMenu/delegationGatingMenu.tsx
@@ -2,7 +2,7 @@ import ModalBottomSheetSwitcher from 'components/modalBottomSheetSwitcher';
 import {useGlobalModalContext} from 'context/globalModals';
 import styled from 'styled-components';
 import React from 'react';
-import {Trans, useTranslation} from 'react-i18next';
+import {useTranslation} from 'react-i18next';
 import {ButtonText, IllustrationHuman, shortenAddress} from '@aragon/ods';
 import {useDaoDetailsQuery} from 'hooks/useDaoDetails';
 import {useDaoToken} from 'hooks/useDaoToken';
@@ -76,17 +76,13 @@ export const DelegationGatingMenu: React.FC = () => {
           <p className="text-xl text-ui-800">
             {t(`modal.${delegationLabel}.title`)}
           </p>
-          <Trans
-            className="text-ui-600"
-            i18nKey={`modal.${delegationLabel}.desc`}
-            components={{b: <strong />}}
-            parent={props => <p {...props} />}
-            values={{
+          <p className="text-ui-600">
+            {t(`modal.${delegationLabel}.desc`, {
               balance: tokenAmount,
               tokenSymbol: daoToken?.symbol,
               walletAddressDelegation: delegateName,
-            }}
-          ></Trans>
+            })}
+          </p>
         </ContentGroup>
         <ContentGroup>
           <ButtonText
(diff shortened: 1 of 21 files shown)


src/context/govTokensWrapping.tsx
```diff
@@ -147,31 +149,53 @@ const GovTokensWrappingProvider: FC<{children: ReactNode}> = ({children}) => {
...
+  // the wrap / unwrap processes
+  const invalidateDaoTokenBalanceCache = useCallback(() => {
+    wagmiQueryClient.invalidateQueries([
+      {
+        entity: 'balance',
+        address: userAddress,
+        token: wrappedDaoToken?.address,
+      },
+    ]);
+    wagmiQueryClient.invalidateQueries([
```
sepehr2github, line 191: Why do you invalidate Query twice?

src/context/globalModals.tsx
```diff
@@ -186,6 +210,8 @@ export const GlobalModalsProvider: React.FC<Props> = ({children}) => {
       isUtcOpen,
       isWalletOpen,
       isDelegateVotingOpen,
+      isDelegationGatingOpen,
```
sepehr2github, line 213: You could probably create a separate context for it, but this works as well

src/containers/delegationGatingMenu/delegationGatingMenu.tsx
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  const tokenAmount = abbreviateTokenAmount(tokenBalance?.formatted ?? '0');
+
+  const {data: delegateData} = useDelegatee(
+    {tokenAddress: daoToken?.address as string},
+    {enabled: daoToken != null}
+  );
+  const isDelegationActive =
+    delegateData !== '0x0000000000000000000000000000000000000000';
```
sepehr2github: Not important, but you can use AddressZero here 😜

sepehr2github (commented):
Thanks Dear Ruggero for the great work you've done. Left some minor comments

src/containers/delegationGatingMenu/delegationGatingMenu.tsx
cgero-eth: Absolutely 👌🏻

src/context/govTokensWrapping.tsx
cgero-eth: After a wrap or unwrap action, I'm invalidating the cached DAO token and Wrapped DAO token balances, you'll see that the `token` parameter differs between the two query keys (`wrappedDaoToken?.address` first and `daoTokenData?.address` later).

src/context/globalModals.tsx
cgero-eth: Agree, I think it makes sense to open a separate ticket to revisit the dialog implementation. In general I think there will be the need of passing data to dialogs, with this `modalState` we have a generic solution that might work for all dialogs but maybe that's not the best approach

src/context/govTokensWrapping.tsx
sepehr2github: Aw...., Your right, My bad! Thanks for the clarification

sepehr2github (approved):
Thanks for the changes

PR #1014, opened 2023-08-29
chore: Update values of global modal context

cgero-eth:
## Description

- As we are only opening one dialog at a time, we don't really need to keep track of every dialog's open/closed state but we can just use one `activeDialog` variable that is going to be set as the name of the current active dialog;
- With this in place, we can either call the `useGlobalModalContext` hook with the dialog name we are handling to check if the current dialog is open or not (e.g. `const { isOpen } = useGlobalModalContext('transfer')`) or we can use the `activeDialog` variable set on the context (e.g. `const { activeDialog } = useGlobalModalContext; const isOpen = activeDialog === 'transfer';`);
- Similarly, for the `close` function, we don't need to specify which dialog we want to close as there will be only one dialog to close;

## Type of change

## Checklist:

### DIFF
--- a/src/context/globalModals.tsx
+++ b/src/context/globalModals.tsx
@@ -14,7 +14,7 @@ type GlobalModalsContextType<TState = Record<string, unknown>> = {
   modalState?: TState;
   isOpen?: boolean;
   open: (dialog: DialogType, state?: Record<string, unknown>) => void;
-  close: () => void;
+  close: (onClose?: () => void) => void;
 };
 
 export type DialogType =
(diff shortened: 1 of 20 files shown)


src/context/globalModals.tsx
```diff
@@ -52,169 +38,31 @@ export type MenuTypes =
...
-      case 'exportCsv':
-        setIsExportCsvOpen(isOpen);
-        break;
-      case 'delegationGating':
-        setIsDelegationGatingOpen(isOpen);
-        break;
-      default:
-        throw new Error(`GlobalModals: modal ${type} unsupported`);
-    }
+  const close = useCallback(() => {
```
Fabricevladimir: It would be really nice if we can take in a callback here. Like really convenient

Fabricevladimir (approved):
Thanks! Really good refactor there

src/context/globalModals.tsx
cgero-eth: Sure makes sense! 👍🏻

PR #1015, opened 2023-08-30
chore: Remove unused dependencies and functionalities

cgero-eth:
## Description

Elastic APM and error reporting is now working, by checking the Kibana dashboard ([here](https://kibana-monitoring.aragon.org/app/apm/services/aragon-app/overview?comparisonEnabled=true&environment=local&kuery=error.exception.type:%22ReferenceError%22&latencyAggregationType=avg&offset=1w&rangeFrom=now-7d%2Fd&rangeTo=now&serviceGroup=&transactionType=page-load)) we can see lots of `cannot access "x" before initialization` errors. This error is caused by Vite during the HMR process because of circular dependencies on the codebase (see comment [here](https://github.com/vitejs/vite/issues/3033#issuecomment-973247243)) and, in order to start fixing those, we can start by removing everything that is not currently used and needed.

- Remove `graphql-anywhere` and `qs` dependencies as not used (they were used by the `apollo-link-rest` dependency that has been removed)
- Remove unused ABIs, components, hooks and utilities (found by using [knip](https://github.com/webpro/knip) tool)
- Remove unused locale files
- Remove commented out code

## Type of change

## Checklist:


src/components/addWallets/header.tsx
```diff
@@ -24,10 +24,10 @@ const AddWalletsHeader: React.FC = () => {
 
 export default AddWalletsHeader;
 
-export const Container = styled.div.attrs({
```
Fabricevladimir, line 27: Really weird how these were being exported for no conceivable reason

src/containers/actionBuilder/mintTokens/index.tsx
```diff
@@ -405,31 +405,6 @@ export const MintTokenForm: React.FC<MintTokenFormProps> = ({
     trigger,
   ]);
 
-  // const handleCSVUpload = (e: React.ChangeEvent<HTMLInputElement>) => {
```
Fabricevladimir, line 408: Wanted to keep this because this actually works. We just haven't enabled it

src/hooks/useFavoritedDaos.tsx
```diff
@@ -82,49 +75,6 @@ export const useFavoritedDaosInfiniteQuery = (
   });
 };
 
-/**
- * Fetch a favorite DAO from the cache
- * @param daoAddress address of the favorited DAO
- * @param network network of the favorited DAO
- * @returns favorited DAO with given address and network if available
- */
-export const useFavoritedDaoQuery = (
```
Fabricevladimir: I want to keep these two as well. The reason for these two hooks is to get rid of the dependency on the Apollo client reactivar caching.

UPDATE:
These are the functions that should be fetching/mutating favorited DAOs once they've been relocated to the backend.

src/hooks/useIsMounted.tsx
```diff
@@ -1,22 +0,0 @@
-import {useCallback, useEffect, useRef} from 'react';
-
-/**
- * Hook to determine if component is mounted before updating the state,
- * thereby preventing a memory leak
- * @returns Whether the current component is mounted.
- */
-const useIsMounted = () => {
```
Fabricevladimir, line 8: There are quite a few places where we have memory leaks that this could have helped with. That was the original intention anyway. Seems like we all forgot about the hook.

src/locales/en-US/common.json
```diff
@@ -1,604 +0,0 @@
-{
-  "subtitle": "Welcome to",
```
Fabricevladimir, line 2: Gotta keep these translation strings since we are translating the App

src/locales/es-ES/common.json
```diff
@@ -1,692 +0,0 @@
-{
-  "subtitle": "Welcome to",
```
Fabricevladimir, line 2: Same here

src/locales/pt-BR/common.json
```diff
@@ -1,692 +0,0 @@
-{
-  "subtitle": "Bem-vindo a",
```
Fabricevladimir, line 2: Same here

src/locales/ru-UA/common.json
```diff
@@ -1,948 +0,0 @@
-{
```
Fabricevladimir, line 1: And here lol

src/services/cache.ts
```diff
@@ -40,28 +40,6 @@ export async function getFavoritedDaosFromCache(options: {
...
 }
 
-/**
- * Fetch a favorited DAO from the cache if available
- * @param daoAddress the address of the favorited DAO to fetch
- * @param chain the chain of the favorited DAO to fetch
- * @returns a favorited DAO with the given address and chain or null
- * if not found
- */
-export async function getFavoritedDaoFromCache(
```
Fabricevladimir: This goes together with the useFavoritedDaoQuery. Remove this if you remove the query itself. This file is supposed to be a placeholder for backend services. The idea is that these methods get translated since we don't really want these things in the browser cache

src/utils/__tests__/abiDecoder.test.ts
```diff
@@ -243,437 +243,16 @@ describe('abi decoder', () => {
     expect(decodedData!.params[0].type).toEqual('uint256[]');
   });
 
-  it('decode logs without indexed', () => {
```
Fabricevladimir, line 246: Why is the text being deleted?

Fabricevladimir, line 246: Why is this test being deleted?

src/utils/__tests__/tokenInfo.test.txt
```diff
@@ -3,26 +3,12 @@ import {getAlchemyProvider} from 'context/providers';
...
   getTotalHolders,
-  getTotalSupply,
 } from 'services/covalentAPI';
 import {CHAIN_METADATA} from 'utils/constants';
-import {getOwner} from 'utils/tokens';
 
 // Slow tests, only use adhoc
 
 describe('token info', () => {
-  it('get owner succeed', async () => {
```
Fabricevladimir, line 14: Isn't this supposed to be a typescript file instead of a plaintext one?

src/utils/constants/misc.ts
```diff
@@ -132,7 +132,6 @@ export const PROPOSAL_STATE_LABELS = [
 
 // Storage and cacheing keys
 export const FAVORITE_DAOS_KEY = 'favoriteDaos';
-export const PENDING_DEPOSITS_KEY = 'pendingDeposits';
```
Fabricevladimir, line 135: Why does it feel like pending deposits are being completely removed...

Fabricevladimir (commented):
Whew quite a bit in there. Thanks, Ruggero.

Fabricevladimir (commented):
Whew quite a bit in there. Thanks, Ruggero.

src/components/addWallets/header.tsx
cgero-eth: Yeah 🤷🏻

src/containers/actionBuilder/mintTokens/index.tsx
cgero-eth: Usually I prefer not to have commented out code on the codebase because it gets outdated and not needed over time. You can find lots of articles about why this is a bad practice on Google like [this](https://methodpoet.com/commented-out-code/#:~:text=Commented%2Dout%20code%20is%20a%20source%20code%20that%20has%20been,issues%20if%20it's%20not%20removed.). 
I'd suggest to discuss this topic during the next daily and decide if we want to keep such code or not, is it fine for you?

src/locales/en-US/common.json
cgero-eth: We're just using the `en` translations as for now and we never update the others, right? As they're not needed for now I'd just remove them from the codebase because it might be confusing for people that want to contribute

src/utils/__tests__/abiDecoder.test.ts
cgero-eth: Because it tests the `decodeLogs` function on the `abiDecoder` file that has been deleted

src/utils/__tests__/tokenInfo.test.txt
cgero-eth: Yes, there's a comment on this file that says:
> // Slow tests, only use adhoc

I suppose that someone changed the extension to `.txt.` to disable them. A better idea would actually be using the `describe.skip` function from Jest and keep the right file extension, but if you ask me I'd remove the whole file as well as not used 🙈

src/utils/constants/misc.ts
cgero-eth: Mm.. Looks like we're not using the `pendingDeposits` variables anywhere :/ Do you know if they've been removed for some reason?

RuggeroCino:
> Whew quite a bit in there. Thanks, Ruggero.

src/locales/en-US/common.json
Fabricevladimir: Fair enough

src/utils/constants/misc.ts
Fabricevladimir: Ah, my bad, actually it's not needed anymore. This was for when we handled deposits in app rather than just hand the user the treasury address

Fabricevladimir (approved):

src/hooks/useFavoritedDaos.tsx
cgero-eth: Okay thanks for the clarification, I'll restore them and we'll see if we can use them when we have the functionality on the backend 👍🏻

src/services/cache.ts
cgero-eth: I restored them 👍🏻

PR #1029, opened 2023-09-17
APP-2455 Mock Protocol Update SDK function

sepehr2github:
## Description

- Add protocol update sdk function
- fetch protocol version info
- fix Styledlink color CSS issues
- Fix the setting page responsive

Task: [APP-2455](https://aragonassociation.atlassian.net/browse/APP-2455)

## Type of change

## Checklist:

- [] I have updated the `CHANGELOG.md` file in the root folder.

[APP-2455]: https://aragonassociation.atlassian.net/browse/APP-2455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/validators.ts
+++ b/src/utils/validators.ts
@@ -18,7 +18,6 @@ import {
   ActionWithdraw,
   Nullable,
 } from './types';
-import {useQueries} from '@tanstack/react-query';
 
 export type TokenType =
   | 'ERC-20'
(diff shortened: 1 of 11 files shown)


src/utils/validators.ts
```diff
@@ -329,3 +330,59 @@ export async function validateWeb3Address(
 export function isFieldValid<T>(error: FieldError | T | undefined): boolean {
   return !error;
 }
+
+export async function updateVerification(
```
babarella, line 333: nit: This is just an address check (1st step of verification flow), would it be better to rename this function to something like `isAddressVerified()`?

src/utils/validators.ts
```diff
@@ -329,3 +330,59 @@ export async function validateWeb3Address(
...
+    return result;
+  }
+}
+
+/**
+ *  This method is a Mock validation function until the real SDK functions are ready
+ * @param address dao address
+ * @returns an arrea of queries the indicates the status of verifications
+ */
+export function UpdateValidation(address: string) {
```
babarella: nit: I don't see why the name starts with capital letter, usually capital letter names are attributes to classes, components etc.  This is just a validation function.

jamesej (approved):
All looks good

babarella (approved):
LGTM, good job, left minors

src/utils/validators.ts
sepehr2github: good point but since It's a mock function I would recommend waiting for SDK helper functions so we can probably use a better name according to SDK functions

src/utils/validators.ts
sepehr2github: Since I used react query the parent components should be a react component which means has capital letters or turn it to the hook which I decide to turn it into the hook

PR #1039, opened 2023-09-27
Chore: APP-2056 - Refactor proposal fetching hooks

Fabricevladimir:
## Description

- Migrate `useProposals` & `useDaoProposal` hooks to react-query
- update usages across APP

NOTE: This does **not** contain cacheing in local storage, so do not merge until
follow up PR is ready

Task: [APP-2056](https://aragonassociation.atlassian.net/browse/APP-2056)

## Type of change

## Checklist:

[APP-2056]: https://aragonassociation.atlassian.net/browse/APP-2056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/constants/misc.ts
```diff
@@ -162,3 +162,6 @@ export const ETH_TRANSACTION_CALL_LABEL = 'ETH transaction call';
 export const PERSONAL_SIGN_LABEL = 'Personal sign';
 export const PERSONAL_SIGN_BYTES = 32;
 export const PERSONAL_SIGN_SIGNATURE = '0x20c13b0b';
+
+// Proposals
+export const PROPOSALS_PER_PAGE = 6;
```
cgero-eth: As this constant is currently used only by the use-proposals hook, I would move it there

src/services/aragon-sdk/queries/use-proposals.ts
```diff
@@ -0,0 +1,69 @@
...
+export const useProposals = (
+  userParams: Partial<IFetchProposalsParams>,
+  options: UseInfiniteQueryOptions<
+    Array<MultisigProposalListItem> | Array<TokenVotingProposalListItem>
+  > = {}
+) => {
+  const params = {...DEFAULT_PARAMS, ...userParams};
+  const client = usePluginClient(params.pluginType);
+
+  options.enabled = !(!client || !params.daoAddressOrEns);
```
cgero-eth: In this way we are overriding the `enabled` option when defined by the component using the hook. I would just override it when the `client` or `daoAddressOrEns` vars are not defined like:
```
if (client == null || params.daoAddressOrEns == null) {
  opitons.enabled = false
}
```

src/pages/dashboard.tsx
```diff
@@ -342,9 +342,16 @@ const DashboardContent: React.FC<DashboardContentProps> = ({
   pluginAddress,
 }) => {
   const {transfers, totalAssetValue} = useDaoVault();
-  const {data: proposals} = useProposals(daoAddressOrEns, pluginType);
 
-  const proposalCount = proposals.length;
+  const {data: proposals} = useProposals({
+    daoAddressOrEns,
+    pluginType,
+  });
```
cgero-eth: Instead of calling `proposals?.pages.flat() ?? []` below multiple times, I would define this as a variable:
```suggestion
  const {data: proposalsData} = useProposals({
    daoAddressOrEns,
    pluginType,
  });
  const proposals = proposalsData?.pages.flat() ?? [];
```

src/pages/proposal.tsx
```diff
@@ -146,13 +143,20 @@ export const Proposal: React.FC = () => {
...
-    intervalInMills
+    refetch,
+  } = useProposal(
+    {
+      pluginType: pluginType,
+      id: proposalId?.toString() ?? '',
+    },
+
+    {
+      refetchInterval: shouldRefetchOnInterval
```
cgero-eth: Can't we just check the `proposalStatus` here instead of using the `shouldRefetchOnInterval` state?
```suggestion
      refetchInterval: proposalStatus === ProposalStatus.ACTIVE
```

cgero-eth (commented):
Well done @Fabricevladimir!

src/pages/dashboard.tsx
Fabricevladimir: Fair point

src/pages/proposal.tsx
Fabricevladimir: That was my first thought, but turns out, I can't use `proposalStatus` because it hasn't been defined yet. Same with `proposal`, which makes sense since it's kind of a chicken and egg thing with `proposal` and `proposalStatus` being obtained by the very hook that requires it in its setup.

src/services/aragon-sdk/queries/use-proposals.ts
Fabricevladimir: Fair, I guess you could want to disable it wherever the hook is being used using a different set of conditions. Nice catch.

src/utils/constants/misc.ts
Fabricevladimir: Yeah, it was supposed to be used in the Governance and Dashboard pages, but then since I added it as a default in the `use-proposals` hook, that became unnecessary

src/pages/proposal.tsx
cgero-eth: You're right 🙈 Buut I just saw that react-query also accept a function for the `refetchInterval` parameter that takes as input the latest fetched data, can we try doing something like:
```
refetchInterval: (data) => data?.status === ProposalStatus.ACTIVE
```

Fabricevladimir: Sweet!

cgero-eth (approved):
👏🏻👏🏻👏🏻

PR #1043, opened 2023-09-28
feat: Migrate React to v18

cgero-eth:
## Description

- Update React dependencies to v18
- Fix re-render issues

## Type of change

## Checklist:


svgTransform.ts
```diff
@@ -1,6 +1,6 @@
 module.exports = {
```
Fabricevladimir, line 1: Came across something interesting the other day. `vite-plugin-svgr` has evolved to where we don't have to bend over backwards anymore and can use svgs like react components, like so: 
```jsx
import MySvg from 'assets/mysvg.svg?react';

<div><MySvg /></div>
```

src/containers/actionBuilder/addAddresses/index.tsx
```diff
@@ -66,11 +66,11 @@ const AddAddresses: React.FC<AddAddressesProps> = ({
    *************************************************/
   useEffect(() => {
     if (controlledWallets.length === 0) {
-      append({address: '', ensName: ''});
+      replace({address: '', ensName: ''});
```
Fabricevladimir, line 69: Not sure whether it's the replace, but when I try to add a authorize member action to a proposal, the next button stays disabled.

src/containers/actionBuilder/mintTokens/index.tsx
```diff
@@ -171,15 +171,15 @@ export const MintTokenForm: React.FC<MintTokenFormProps> = ({
    *************************************************/
 
   useEffect(() => {
-    // set-up form on first load/reset
+    // Setup form on first load/reset using replace to avoid calling append multiple times
     if (fields.length === 0) {
-      append({web3Address: {address: '', ensName: ''}, amount: '0'});
+      replace({web3Address: {address: '', ensName: ''}, amount: '0'});
```
Fabricevladimir, line 176: Same thing here, when I add a singular wallet to mint tokens, the next button stays disabled as well

src/containers/proposalStepper/index.tsx
```diff
@@ -57,13 +57,21 @@ const ProposalStepper: React.FC<ProposalStepperType> = ({
...
 
-  const {errors, dirtyFields} = useFormState({
-    control,
-  });
+  const {errors, dirtyFields} = useFormState({control});
 
-  actionsAreValid(formActions, actions, errors, network).then(isValid =>
-    setIsActionsValid(isValid)
-  );
+  useEffect(() => {
```
Fabricevladimir, line 62: Ah, this might be the next button not enabled culprit. This doesn't seem to be triggered when as the actions get updated.

src/context/actions.tsx
```diff
@@ -39,18 +41,23 @@ const ActionsProvider: React.FC<ActionsProviderProps> = ({daoId, children}) => {
   const {control} = useFormContext();
   const {remove} = useFieldArray({control, name: 'actions'});
 
-  const addAction = useCallback(newAction => {
-    setActions(current => {
-      const shouldAddEditMultisigAction =
-        updatesMultisigVoting(newAction) && !hasEditMultisigAction(current);
+  const addAction = useCallback(
+    (newAction: ActionItem, allowDuplicate = true) => {
```
Fabricevladimir, line 45: Hey, as far as I can tell there are no side-effects to this, but I'm not entirely sure why allowDuplicated is introduced here. Aren't we indicating which actions can be duplicated in the Add Action Menu actions using the useDaOActions hook response already?

Fabricevladimir (commented):
Whew! Lots of changes! Thanks for the updates; left a couple of comments

Fabricevladimir:
Also, what is up with this list of commits? 🤔

src/containers/proposalStepper/index.tsx
cgero-eth: Good catch, we were validating the form after every re-render, by using the `useWatch` without an array of field names we are now validating after every change on the `actions` field

src/containers/actionBuilder/mintTokens/index.tsx
cgero-eth: See comment [here](https://github.com/aragon/app/pull/1043#discussion_r1356602028)

src/containers/actionBuilder/addAddresses/index.tsx
cgero-eth: See comment [here](https://github.com/aragon/app/pull/1043#discussion_r1356602028)

svgTransform.ts
cgero-eth: Yeah the Icon component on the new ODS is using this to render the icons, but here we still need to parse the svg files because they are not supported by jest testing environment

cgero-eth:
> Also, what is up with this list of commits? 🤔

I accidentally rebased this branch with the `main` branch instead of `develop`, that's why the commit list is messed up 🙈  But we'll squash all the commits of this PR before merging it to `develop` and we'll only have one commit there.

**EDIT**: I squashed them locally so that we have now only one nice commit :D

src/context/actions.tsx
cgero-eth: We're only using the `allowDuplicate` parameter set to false on the `configureActions` component. I found out that React18 in `StrictMode` will call the `useEffect` twice on development (more info [here](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-double-rendering-in-development)), therefore I had to introduce the `replace` and this `allowDuplicate` parameter even though this behaviour only affects the local development.

Fabricevladimir: Ok thanks 👍🏾

svgTransform.ts
Fabricevladimir: Hmm, interesting

src/containers/actionBuilder/scc/index.tsx
```diff
@@ -25,7 +25,14 @@ const SCCAction: React.FC<ActionIndex & {allowRemove?: boolean}> = ({
   const {network} = useNetwork();
   const [isValid, setIsValid] = useState(true);
 
-  validateSCCAction(actionData, network).then(res => setIsValid(res));
+  useEffect(() => {
```
Fabricevladimir, line 28: Ah, same thing here with the validation not triggering. When you add an SCC action with errors, fixing the errors in the form once the action has been added removes the error state of the component, but the error message for the entire action doesn't get removed. This isn't a huge deal, but that means you *can* proceed with an error message.

Fabricevladimir (approved):
Thanks!

src/containers/actionBuilder/scc/index.tsx
cgero-eth: You're right, I applied the same fix on the SCC action-data validation 👍🏻

Fabricevladimir (approved):

src/containers/actionBuilder/scc/index.tsx
Fabricevladimir: Thanks!

PR #1046, opened 2023-09-28
Chore: APP-2056 - Add Cacheing capabilities to proposal hooks

Fabricevladimir:
## Description

- Adds `LocalStorage` utility classes
- refactors browser cacheing to work with `react-query`
- removes reactivar structures pertaining to proposal storage

Note: Please review after #1039 has been merged since it's the base branch and this is also carrying all the commits from it.

~~TODO: Investigate why SDK no longer allows access to function calculating the proposal status: this makes it so that a cached proposal cannot go through the proposal lifecycle until the subgraph has indexed it.~~

Task: [APP-2056](https://aragonassociation.atlassian.net/browse/APP-2056)

## Type of change

## Checklist:

[APP-2056]: https://aragonassociation.atlassian.net/browse/APP-2056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/containers/proposalSnapshot/index.tsx
```diff
@@ -90,7 +102,11 @@ const ProposalSnapshot: React.FC<Props> = ({
     ]
   );
 
-  if (proposals.length === 0 || areMembersLoading) {
+  if (proposalsAreLoading || membersAreLoading) {
+    return <Loading />;
+  }
+
+  if (proposalsFetched && mappedProposals?.length === 0) {
```
cgero-eth: Should we include the `!mappedProposals` case here and return an empty state instead of `null`?

src/pages/dashboard.tsx
```diff
@@ -298,15 +298,15 @@ export const Dashboard: React.FC = () => {
 
         {isDesktop ? (
           <DashboardContent
-            daoAddressOrEns={liveAddressOrEns}
+            daoAddressOrEns={liveDao.address}
```
cgero-eth: Shouldn't we keep here the `liveAddressOrEns`?

src/pages/dashboard.tsx
```diff
@@ -342,8 +342,17 @@ const DashboardContent: React.FC<DashboardContentProps> = ({
...
+  const {data} = useProposals({
+    daoAddressOrEns,
+    pluginType,
+    pluginAddress,
+  });
+  const proposals = data?.pages.flat() ?? [];
+
+  // The SDK does NOT provide a count. This will be incorrect
+  // whenever the number of proposals is greater than the default
+  // page size that we fetch: 6
```
cgero-eth: Are Carlos / the SDK team aware about this? If we have a Jira ticket we could link it here

cgero-eth:
@Fabricevladimir I'll wait for the other PR to be merged on develop so that this PR displays the actual file changed

src/utils/constants/misc.ts
```diff
@@ -162,3 +158,5 @@ export const ETH_TRANSACTION_CALL_LABEL = 'ETH transaction call';
 export const PERSONAL_SIGN_LABEL = 'Personal sign';
 export const PERSONAL_SIGN_BYTES = 32;
 export const PERSONAL_SIGN_SIGNATURE = '0x20c13b0b';
+
+export const PROPOSALS_PER_PAGE = 6;
```
cgero-eth: I would move this to the `useProposals` hook

src/utils/localStorage/proposalStorage.ts
```diff
@@ -0,0 +1,106 @@
...
+   * @returns - An array of retrieved proposals associated with the given plugin address
+   */
+  getProposalsByPluginAddress<
+    T extends MultisigProposalListItem | TokenVotingProposalListItem
+  >(chainId: SupportedChainID, pluginAddress: string): T[] {
+    const key = chainId.toString();
+    const proposals: ProposalCache = this.getItem(key) || {};
+    const matchingProposals = Object.values(proposals)
+      .filter(proposal => proposal.id.startsWith(pluginAddress))
+      .map(proposal => proposal);
```
cgero-eth: The `.map` here is redundant, right?
```suggestion
```

src/pages/proposal.tsx
```diff
@@ -639,12 +637,16 @@ export const Proposal: React.FC = () => {
   /*************************************************
    *                     Render                    *
    *************************************************/
-  if (proposalError) {
+  if (paramsAreLoading || proposalIsLoading || detailsAreLoading) {
+    return <Loading />;
+  }
+
+  if (proposalError || (proposalIsFetched && proposal === null)) {
```
cgero-eth: Same, should we include here the `!proposal` case?

src/utils/localStorage/executionStorage.ts
```diff
@@ -0,0 +1,80 @@
...
+// Type alias for chain data, mapping proposal IDs to their execution details.
+type ExecutionCache = {
+  [proposalId: string]: ExecutionDetail;
+};
+
+/**
+ * ExecutionStorage class provides methods for managing execution statuses
+ * in local storage. The data is structured by chainId and then by proposalId.
+ */
+export class ExecutionStorage extends StorageUtils {
```
cgero-eth, line 19: Instead of instantiating a new `ExecutionStorage` class every time we need it, we could instantiate it once here and export only the instance:
```suggestion
class ExecutionStorage extends StorageUtils {
```
And down below:
```
export const executionStorage = new ExecutionStorage();
```

src/services/aragon-sdk/queries/use-proposals.ts
```diff
@@ -33,40 +44,157 @@ async function fetchProposalsAsync(
...
+    params.pluginAddress == null
+  ) {
     options.enabled = false;
   }
 
+  // get the previously merged local storage proposals from the react-query cache
+  // this helps to restore the state after component is unmounted/remounted
+  const previouslyMergedStoredProposals: Set<string> =
+    queryClient.getQueryData(
+      aragonSdkQueryKeys.localProposals(params.status)
```
cgero-eth, line 70: I don't really get the difference between these proposals and the `allLocalProposals` array, let's maybe sync to let me understand this

src/utils/localStorage/proposalStorage.ts
Fabricevladimir: It is, not sure why I added that.

src/pages/dashboard.tsx
Fabricevladimir: SDK team is aware of this, but I can bring it up again.

src/containers/proposalSnapshot/index.tsx
Fabricevladimir: The tricky thing here is that, if mappedProposals is undefined/null that signals that either proposals are being fetched or there's an error which we should be checking for in this same condition and alerting the user of that. For now I can include it in the previous check, but I don't think, it's the best place for it.

cgero-eth (approved):
💯

PR #1050, opened 2023-09-30
fix: typos

omahs:
fix: typos


cgero-eth (commented):
Thanks for the fixes @omahs, as we are using a third party service to handle the translations we need to update these values there first. I'll approve the PR as soon as the service values are aligned 👍🏻

cgero-eth (approved):

PR #1060, opened 2023-10-09
Fix: APP-2501 SCC field validations for uint, int, and bool

RakeshUP:
## Description

To test, use the `removeLiquidityETHWithPermit` method on `0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D` contract on goerli network.

Task: [APP-2501](https://aragonassociation.atlassian.net/browse/APP-2501)

- Removed ability to add decimals on uint and int fields
- bool only accepts true/false now

## Type of change

## Checklist:

[APP-2501]: https://aragonassociation.atlassian.net/browse/APP-2501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/smartContractComposer/components/inputForm.tsx
+++ b/src/containers/smartContractComposer/components/inputForm.tsx
@@ -353,7 +353,11 @@ export const ComponentForType: React.FC<ComponentForTypeProps> = ({
                   input={component}
                   functionName={`${functionName}.${input.name}`}
                   formHandleName={
-                    formHandleName ? `${formHandleName}[${index}]` : undefined
+                    formHandleName
+                      ? `${formHandleName}[${index}]`
+                      : disabled
+                      ? `${formName}[${index}]`
+                      : undefined
                   }
                   disabled={disabled}
                   isValid={isValid}
--- a/src/containers/smartContractComposer/components/inputForm.tsx
+++ b/src/containers/smartContractComposer/components/inputForm.tsx
@@ -353,11 +353,7 @@ export const ComponentForType: React.FC<ComponentForTypeProps> = ({
                   input={component}
                   functionName={`${functionName}.${input.name}`}
                   formHandleName={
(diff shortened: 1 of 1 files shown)


src/containers/smartContractComposer/components/inputForm.tsx
```diff
@@ -303,6 +305,54 @@ export const ComponentForType: React.FC<ComponentForTypeProps> = ({
         />
       );
 
+    case 'bool':
+      return (
+        <Controller
+          defaultValue=""
+          name={formName}
+          render={({field: {name, value, onBlur, onChange}}) => (
+            <TextInput
```
Fabricevladimir, line 316: I almost wonder if a checkbox wouldn't be a more user friendly component for the boolean.

Fabricevladimir (approved):
Thanks, @RakeshUP. It would be a serious improvement I think to be able to display some sort of message to the user about a field. Right now we're just highlighting the input, but fundamentally know what is wrong with the input. Another thing I noticed as well is that the validation via ABI can fail without a singular field being flagged; that's fine, but the error message is still all fields are required. Of course none of this have to do with the PR, but I'd like to get your thoughts on this.

src/containers/smartContractComposer/components/inputForm.tsx
RakeshUP: I've created a [new ticket for this](https://aragonassociation.atlassian.net/browse/APP-2543?atlOrigin=eyJpIjoiMGNhYWZjNjJkOWViNGJiNDljMDM1MzY2OTNjOTk1NTkiLCJwIjoiaiJ9). Will Evan for his inputs as well there for the checkbox and error messages in SCC actions

PR #1070, opened 2023-10-16
fix: APP-2266 - Add checks to verify dApp connection

cgero-eth:
## Description

- Update the `wcConnect` function to check for connection status and throw error in case the connection is not valid
- Add the following checks in order to verify the status of the dApp connection and display error in case the connection is expired or not valid:
  - Ping the wallet-connect topic to verify its status;
  - Throw error when connection expires by checking the `connection.timeout` attribute;
  - Throw error on internal expiration check (after 60 seconds) in case there's no timeout set on the connection;

Task: [APP-2266](https://aragonassociation.atlassian.net/browse/APP-2266)

## Type of change

## Checklist:

[APP-2266]: https://aragonassociation.atlassian.net/browse/APP-2266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/walletConnect/dAppValidationModal/index.tsx
+++ b/src/containers/walletConnect/dAppValidationModal/index.tsx
@@ -133,7 +133,11 @@ const WCdAppValidation: React.FC<Props> = props => {
     setConnectionStatus(ConnectionState.LOADING);
 
     try {
-      await wcConnect({uri, metadataName: selecteddApp?.name.toLowerCase()});
+      const session = await wcConnect({
+        uri,
+        metadataName: selecteddApp?.name.toLowerCase(),
+      });
+      setSessionTopic(session.pairingTopic);
       setConnectionStatus(ConnectionState.SUCCESS);
     } catch (error: unknown) {
       if (error === METADATA_NAME_ERROR) {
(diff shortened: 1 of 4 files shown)


RakeshUP:
Hey @RuggeroCino, the 60-second timeout is working well, but I'm having an issue with the connection status. There's a race condition when connectionStatus is SUCCESS, but the sessions haven't been updated. This leads to currentSession to be null [here](https://github.com/aragon/app/blob/a02124c5f39c030ec409b7e43f853db38cbb33a3/src/containers/walletConnect/dAppValidationModal/index.tsx#L93).

And this makes the connection status to reset [here](https://github.com/aragon/app/blob/a02124c5f39c030ec409b7e43f853db38cbb33a3/src/containers/walletConnect/dAppValidationModal/index.tsx#L151). 

I tried it multiple times and it's the status is always being reset.

cgero-eth:
> Hey @RuggeroCino, the 60-second timeout is working well, but I'm having an issue with the connection status. There's a race condition when connectionStatus is SUCCESS, but the sessions haven't been updated. This leads to currentSession to be null [here](https://github.com/aragon/app/blob/a02124c5f39c030ec409b7e43f853db38cbb33a3/src/containers/walletConnect/dAppValidationModal/index.tsx#L93).
> 
> And this makes the connection status to reset [here](https://github.com/aragon/app/blob/a02124c5f39c030ec409b7e43f853db38cbb33a3/src/containers/walletConnect/dAppValidationModal/index.tsx#L151).
> 
> I tried it multiple times and it's the status is always being reset.

Good catch, I wasn't setting anymore the session topic after successful connection and `currentSession` was always undefined. I updated the code and it should work now

RakeshUP (approved):

PR #1074, opened 2023-10-16
feat: APP-2572 - Integrate Vocdoni gasless plugin

selankon:
## Description

Integrate the Vocdoni gasless plugin into the Aragon App 

**TODOs**:
- ✅ Sign CLA 
- ✅ Code review
- ✅ Smoke tests (@RuggeroCino)
- ✅ Fix review discussions (@selankon)
- ⏳ Align Crowdin translations (@selankon)
- ✅ Remove local dependencies (@emmdim)

Task: [APP-2572](https://aragonassociation.atlassian.net/browse/APP-2572)

## Type of change

## Checklist:

[APP-2572]: https://aragonassociation.atlassian.net/browse/APP-2572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/containers/configureCommunity/proposalCreation.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <>
+          <DescriptionContainer>
+            <Label
+              label={t('labels.proposalCreation')}
+              helpText={t('createDAO.step3.proposalCreationHelpertext')}
+            />
+          </DescriptionContainer>
+          <SelectEligibility />
+        </>
+      )}
```
elboletaire: Considering the entire component is not rendered with the condition, it would be better to clearly return null after such condition, i.e.

```js
// you could leave the const declaration for better readability, but since it's only
// used once, and for the sake of a shorter example, I've ommited it here
if (!(isCustomToken || tokenType === 'ERC-20' || tokenType === 'governance-ERC20')) {
  return null
}

// normal render of the compnent
return (
  <>
    {/* [...] */}
  </>
)
```

src/containers/configureCommunity/proposalCreation.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <>
+          <MultisigEligibility />
+        </>
+      )}
+      {membership === 'token' && (
+        <>
+          <TokenProposalCreation />
+        </>
+      )}
+    </div>
```
elboletaire: I know these two components did exist already, but I'm wondering if the condition could have been added there instead. Checking them I see they both also access the form context so, unless they're used in different form contexts, with different fields (i.e. `membership` exists in one form, but not in another one already using these two components; or the logic of these two components is just different in another form), this condition could be there.

PS. Remember I'm reviewing by just looking at the code... if you consider this logic makes sense here, just ignore my comment :_)

src/containers/goLive/index.tsx
```diff
@@ -52,13 +53,17 @@ export const GoLiveHeader: React.FC = () => {
...
+
+  const {votingType} = getValues();
 
   return (
     <Container>
       <Blockchain />
       <DaoMetadata />
       <Community />
       <Governance />
+      {votingType === 'offChain' && <Committee />}
```
elboletaire: Maybe this condition could also be in its own component directly (?)

src/context/stepperModal.tsx
```diff
@@ -104,39 +104,48 @@ const StepperModal = <X extends GenericKeyEnum>({
...
+                  <div>{`${nativeCurrency.symbol}`}</div>
+                </StrongText>
+                <p className="text-sm text-right text-ui-500">{totalCost}</p>
+              </VStack>
+            </GasTotalCostEthContainer>
+          </GasCostTableContainer>
+          {gasEstimationError && (
+            <AlertInlineContainer>
+              <AlertInline
+                label={t('TransactionModal.gasEstimationErrorLabel') as string}
```
elboletaire, line 149: `t` also has a `toString()` method, but it's virtually the same you did here (since the only problem of not using `toString()` is typescript complaining about it not being of type string). Just so you know the method exists 😉

src/context/useOffchainVoting.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const client = usePluginClient(GaselessPluginName) as OffchainVotingClient;
+  const {address} = useWallet();
+
+  const isApprovalPeriod = useMemo(() => {
+    if (!proposal) return false;
+    return (
+      proposal.endDate.valueOf() < new Date().valueOf() &&
+      proposal.expirationDate.valueOf() > new Date().valueOf()
+    );
+  }, [proposal]);
```
elboletaire: This is one of those cases where you **could** avoid using `useMemo`. If you still want the function-alike syntax, remember you can use a self-executing function:

```js
const isApprovalPeriod = ((proposal) => {
  if (!proposal) return false;
  return (
    proposal.endDate.valueOf() < new Date().valueOf() &&
    proposal.expirationDate.valueOf() > new Date().valueOf()
  );
})(proposal)
```

Most `useMemo` used below of the aforementioned one could be changed to not be `useMemo`, except for the `approved` one, since there you're traversing an array using `some`, and it's always a good idea to use `useMemo` when working with array functions (specially when those arrays can be gigantic; if you know it'll be short, could be safe to not use it there too).

Also, not one of those crytical changes you should prioritize tbh.. just a heads up about the proper usage of `useMemo`.

src/context/useOffchainVoting.tsx
```diff
@@ -0,0 +1,227 @@
...
+      const canApprove =
+        (await client?.methods.isCommitteeMember(pluginAddress, address!)) ||
+        false;
+      setCanApprove(canApprove);
+    };
+    if (address && client) {
+      approved || !isApprovalPeriod || !proposalCanBeApproved
+        ? setCanApprove(false)
+        : checkCanVote();
+    }
```
elboletaire: This ternary here is quite confusing. I'd suggest:

```js
if (!(address && client)) {
  return
}

if (approved || !isApprovalPeriod || !proposalCanBeApproved) {
  setCanApprove(false)
  return
}
checkCanVote()
```

Edit: reviewing my own review ~20h later... tbh the entire logic in this effect is quite confusing.

src/hooks/useCensus3.tsx
```diff
@@ -0,0 +1,63 @@
...
+    async function checkSupported() {
+      (await census3.getSupportedChains()).includes(chainId)
+        ? setIsSupported(true)
+        : setIsSupported(false);
+    }
+
+    if (chainId && census3) {
+      checkSupported();
+    }
+  }, [census3, chainId]);
```
elboletaire: I'd suggest not using ternaries unless you really need to (i.e. a react component rendering part, or assigning values to an object), since they tend to difficult readability.

Actually, I now do realize you can entirely remove the condition:

```js
useEffect(() => {
  ;(async () => {
    if (chainId && census3) {
      setIsSupported((await census3.getSupportedChains()).includes(chainId))
    }
  })()
}, [census3, chainId]);
```

src/hooks/useCensus3.tsx
```diff
@@ -0,0 +1,63 @@
...
+      // Check if the census is already sync
+      try {
+        const token = await client?.methods.getToken(pluginAddress);
+        if (!token) throw 'Cannot retrieve the token';
+        await census3.createToken(token.address, 'erc20', chainId);
+      } catch (e) {
+        if (e instanceof ErrTokenAlreadyExists) {
+          console.log('DEBUG', 'Token already created');
+        }
+        // todo(kon): handle chain is not supported
```
elboletaire: Friendly reminder to check all comments and, either change what's required, or remove them.

src/hooks/useDaoOffchainProposal.tsx
```diff
@@ -0,0 +1,233 @@
...
+    } else {
+      clearInterval(intervalId);
+    }
+
+    return () => {
+      clearInterval(intervalId);
+    };
+    // This effect only runs when intervalInMills will changed
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [intervalInMills]);
```
elboletaire: No need to use state for an interval id you're not using afterwards:

```js
useEffect(() => {
  let id: number
  if ((intervalInMills || 0) > 0) {
    setNumberOfRuns(value => value + 1);

    id = window.setInterval(() => {
      setNumberOfRuns(value => value + 1);
    }, intervalInMills);

    return
  }
  window.clearInterval(id)

  return () => window.clearInterval(id);
}, [intervalInMills]);
  ```

File: src/hooks/useDaoOffchainProposal.tsx [hunk 10]
```diff
@@ -0,0 +1,233 @@
...
+
+  useEffect(() => {
```
elboletaire: It's always a good idea to add a comment at the top of the effect describing its usage, specially in huge ones like this, and considering the condition to run it is ~60lines below.

src/hooks/useDaoOffchainProposal.tsx
```diff
@@ -0,0 +1,233 @@
...
+        setIsLoading(false);
+      }
+    };
+
+    if (
+      daoAddress &&
+      proposalGuid &&
+      (isMultisigPlugin || (isTokenBasedPlugin && daoToken))
+    )
+      getDaoProposal(proposalGuid);
```
elboletaire: Use brackets for better readability (and consistency with all the code).

Personally, I preffer having these conditions in effects at the very begin, so you clearly see **when** the effect is gonna run. And since I usually use self-calling functions in async effects, the creation of the named method `getDaoProposal` (which is used only once) would not be required, making everything look quite different. But personal preferences, I guess :)

Edit: now I see this is probably a modified copy-paste coming from `useProposals`, so I understand you just changed an already existing flow 😅. There are actually more files very similar... `useProposals`, `useDaoOffchainProposal` and `useDaoProposal` are all quite similar.

src/hooks/useWalletCanVote.tsx
```diff
@@ -43,32 +92,9 @@ export const useWalletCanVote = (
...
-            });
-          });
-          canVote = await Promise.all(canVoteValuesPromises);
-        }
-
-        if (canVote !== undefined) setData(canVote);
-        else setData([false, false, false]);
+        isOffchainVoting
+          ? await fetchCanVoteOffchain()
+          : await fetchOnchainVoting();
```
elboletaire: These ternaries...

```suggestion
        if (isOffchainVoting) {
          return await fetchCanVoteOffchain();
        }
        await fetchOnchainVoting();
```

src/hooks/useWalletCanVote.tsx
```diff
@@ -31,10 +37,53 @@ export const useWalletCanVote = (
...
+    }
+
+    async function fetchCanVoteOffchain() {
+      let canVote = false;
+      if (offchainProposalId) {
+        canVote = await vocdoniClient.isInCensus(offchainProposalId);
+        if (canVote) {
+          canVote =
+            (await vocdoniClient.hasAlreadyVoted(offchainProposalId)) === null;
+        }
```
elboletaire: Could be a one-liner

```suggestion
        canVote = await vocdoniClient.isInCensus(offchainProposalId) &&
          !(await vocdoniClient.hasAlreadyVoted(offchainProposalId));
```

src/pages/proposal.tsx
```diff
@@ -184,16 +206,50 @@ const Proposal: React.FC = () => {
...
+  //       (proposal as GaslessVotingProposal)?.vochain.metadata.description
+  //         .default || ''
+  //     );
+  //   } else {
+  //     setTitle(proposal?.metadata.title || '');
+  //     setDescription(proposal?.metadata.description || '');
+  //   };
+  //   setSummary(proposal?.metadata.summary || '');
+  // }, [proposal]);
+
```
elboletaire: friendly reminder to remove unwanted comments

src/pages/proposal.tsx
```diff
@@ -398,26 +454,30 @@ const Proposal: React.FC = () => {
...
+        ? false
+        : isTokenVotingSettings(daoSettings)
         ? isEarlyExecutable(
             mappedProps?.missingParticipation,
             proposal,
-            mappedProps?.results,
+            mappedProps?.results as ProposalVoteResults,
             daoSettings.votingMode
           )
         : (proposal as MultisigProposal)?.approvals?.length >=
```
elboletaire: Ouch 😅. And I was complaining about ternaries 😂

src/utils/committeeVoting.ts
```diff
@@ -0,0 +1,60 @@
...
+    return t('offchainVotingTerminal.btnLabel.executed');
+  } else if (approved) {
+    return t('offchainVotingTerminal.btnLabel.approved');
+  } else if (notBegan) {
+    return t('offchainVotingTerminal.btnLabel.approve');
+  } else if (voted) {
+    return t('offchainVotingTerminal.btnLabel.voted');
+  } else if (canApprove) {
+    return t('offchainVotingTerminal.btnLabel.approve');
+  }
```
elboletaire: Maybe use switch here

```suggestion
  switch (true) {
    case executed:
      return t('offchainVotingTerminal.btnLabel.executed');
    case approved:
      return t('offchainVotingTerminal.btnLabel.approved');
    case notBegan:
      return t('offchainVotingTerminal.btnLabel.approve');
    case voted:
      return t('offchainVotingTerminal.btnLabel.voted');
    case canApprove:
      return t('offchainVotingTerminal.btnLabel.approve');
  }
```

I say "maybe" since using typescript may require adding a `default` option with a `throw Error`, and would end up being more code (and that error could bring more issues, considering this seems to be a view helper/render method).

Another option is to not use the `else`, since it's not required in this case (because you're always returning after a condition is met):


```suggestion
  if (executed) {
    return t('offchainVotingTerminal.btnLabel.executed');
  }
  if (approved) {
    return t('offchainVotingTerminal.btnLabel.approved');
  }
  if (notBegan) {
    return t('offchainVotingTerminal.btnLabel.approve');
  }
  if (voted) {
    return t('offchainVotingTerminal.btnLabel.voted');
  }
  if (canApprove) {
    return t('offchainVotingTerminal.btnLabel.approve');
  }
```

src/components/addCommitteeMembers/row.tsx
```diff
@@ -0,0 +1,169 @@
...
+        committee?.some(
+          (wallet, walletIndex) =>
+            ((web3Address.address &&
+              wallet.address.toLowerCase() ===
+                web3Address.address.toLowerCase()) ||
+              (web3Address.ensName &&
+                wallet.ensName.toLowerCase() ===
+                  web3Address.ensName.toLowerCase())) &&
+            walletIndex !== index
+        )
```
elboletaire: Considering the amount of logic required by the `some` param, would be more clear if you define it as a `const` the line before (outside of the `if` condition).

src/containers/defineCommittee/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        }%`,
+                      }}
+                    >
+                      {value}
+                    </p>
+
+                    <p className="flex-shrink-0 text-ui-600">
+                      {t('createDAO.step5.alerts.ofAddresses', {
+                        count: committeeCount,
+                      })}
```
elboletaire: `count` in `i18next` is usually for plurals: https://www.i18next.com/translation-function/plurals

Doesn't mean it won't work this way, just another heads-up. Using an extractor would generate 2 or 3 strings when finding `count` as an interpolation value in a translation.

elboletaire (changes requested):

src/hooks/useDaoOffchainProposal.tsx
selankon: This hook doesn't exists anymore

src/pages/proposal.tsx
selankon: After rebase this ternary was moved into a if else statements

.env.development
```diff
@@ -1,3 +1,4 @@
+
```
cgero-eth: Let's remove this new space here so that all the `.env` files have the same format

.env.staging
```diff
@@ -8,4 +8,7 @@ VITE_FEATURE_FLAG_MONITORING=true
 
 VITE_FEATURE_FLAG_DELEGATION=true
 
-VITE_FEATURE_FLAG_OSX_UPDATES=true
\ No newline at end of file
+VITE_FEATURE_FLAG_OSX_UPDATES=true
+
+# Enable gasless plugin on DAO creation
+VITE_FEATURE_FLAG_GASLESS_PLUGIN=false
```
cgero-eth: I'd suggest to enable the plugin on staging, QA can start testing it there
```suggestion
VITE_FEATURE_FLAG_GASLESS_PLUGIN=true
```

.gitignore
```diff
@@ -1,3 +1,5 @@
+.idea
+
```
cgero-eth: The `.idea` entry is duplicated

package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.5.8",
-    "@aragon/sdk-client": "^1.14.0",
+    "@aragon/sdk-client": "1.14.0",
```
cgero-eth: Let's restore this version
```suggestion
    "@aragon/sdk-client": "^1.14.0",
```

src/components/StepperProgress/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const StepListItem = styled.div.attrs({
+  className: 'flex justify-between text-ui-600',
+})``;
+
+const IconAndMessage = styled.div.attrs({
+  className: 'flex space-x-2 items-center',
+})``;
+
+export default StepperModalProgress;
```
cgero-eth: We should avoid default exports and have one component per file, but I know this is not currently enforced across the app. Feel free to just resolve the comment if you won't have time to do so, I'll create a Jira ticket to fix this anyway

src/hooks/usePluginClient.tsx
```diff
@@ -2,31 +2,50 @@ import {MultisigClient, TokenVotingClient} from '@aragon/sdk-client';
...
   ? TokenVotingClient
   : T extends 'multisig.plugin.dao.eth'
   ? MultisigClient
+  : T extends GaselessPluginType
+  ? GaslessVotingClient
   : never;
 
 export function isTokenVotingClient(
-  client: TokenVotingClient | MultisigClient
+  client: TokenVotingClient | MultisigClient | GaslessVotingClient
```
cgero-eth: I would suggest to introduce a new type like:
```
type PluginClient = TokenVotingClient | MultisigClient | GaslessVotingClient
```
And use this when needed

src/hooks/useVocdoniSdk.tsx
```diff
@@ -0,0 +1,17 @@
+import {Signer} from '@ethersproject/abstract-signer';
+import React, {PropsWithChildren} from 'react';
+import {useWallet} from './useWallet';
+import {ClientProvider} from '@vocdoni/react-providers';
+import {EnvOptions} from '@vocdoni/sdk';
+
+// todo(kon): move this to be set by .env file
+export const VocdoniEnv: EnvOptions = EnvOptions.STG;
```
cgero-eth: We can easily add a new environment variable on the `.env` files like:
```
VITE_VOCDONI_ENV=dev|stg|prod
```
And use it here as:
```
export const CovdoniEnv: EnvOptions = import.meta.env.VITE_VOCDONI_ENV;
```

src/components/addCommitteeMembers/footer.tsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import styled from 'styled-components';
+import {Label} from '@aragon/ods-old';
+import {useTranslation} from 'react-i18next';
+
+type WalletsFooterProps = {
+  totalAddresses: number;
+};
+
+const AddWalletsFooter: React.FC<WalletsFooterProps> = ({totalAddresses}) => {
```
cgero-eth: It would be nice if we could improve the file name <=> component name matching. The folder is called `/addCommitteeMembers` but the component is `AddWalletsFooter`, I would probably rename these components to `AddCommittee*` like `AddCommitteeFooter` (and the folder to `/addCommittee` ?)

src/components/addCommitteeMembers/footer.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const FooterRow = styled.div.attrs({
+  className: 'flex',
+})``;
+
+export const FooterItem1 = styled.div.attrs({
+  className: 'flex-1',
+})``;
+
+export const FooterItem2 = styled.div.attrs({
```
cgero-eth: Let's not export components if they're not used. You can find a list of unused exports/files/dependencies by running the [knip](https://github.com/webpro/knip) tool: `npx knip`
```suggestion
const FooterItem1 = styled.div.attrs({
  className: 'flex-1',
})``;

const FooterItem2 = styled.div.attrs({
```

src/components/addCommitteeMembers/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+    control,
+  });
+  const {alert} = useAlertContext();
+
+  const controlledFields = fields.map((field, index) => {
+    return {
+      ...field,
+      ...(wallets && {...wallets[index]}),
+    };
+  });
```
cgero-eth: Is the `controlledFields` variable needed here? As far as I can see we're only using the fields id and total length and we might remove the `useWatch` call to the `committee` field as well

src/components/addCommitteeMembers/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+  useEffect(() => {
+    if (address && !wallets) {
+      // uncomment when minting to treasury is ready
+      // insert(1, {address: address, amount: '0'});
```
cgero-eth: I would remove the commented out code

src/components/addCommitteeMembers/row.tsx
```diff
@@ -0,0 +1,168 @@
...
+        web3Address,
+        t('errors.required.walletAddress'),
+        t
+      );
+
+      if (validationResult && validationResult !== true) {
+        return validationResult;
+      }
+
+      const isDuplicateAddress = committee?.some(
```
cgero-eth: Can we use the `walletInWalletList` utility here? Looks like it has been introduced but it is not used

src/components/StepperProgress/index.tsx
```diff
@@ -0,0 +1,100 @@
+import React from 'react';
```
cgero-eth: Folder name should be `stepperProgress` and not `StepperProgress`

src/components/executionExpirationtime/index.tsx
```diff
@@ -0,0 +1,269 @@
+import React, {useCallback} from 'react';
```
cgero-eth: Folder name should be `executionExpirationTime` and not `executionEpirationtime`

src/components/executionExpirationtime/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+                <AlertInline label={error.message} mode="critical" />
+              )}
+            </TimeLabelWrapper>
+          )}
+        />
+      </DurationContainer>
+      {executionExpirationDays ===
+      COMMITTEE_EXECUTION_MAX_DURATION_DAYS.toString() ? (
+        <AlertInline
+          label={t('alert.maxDurationAlert') as string}
```
cgero-eth: The cast to string is redundant here and below as well, does it throws some type error for you?
```suggestion
          label={t('alert.maxDurationAlert')}
```

src/components/filteredAddressList/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+            pageSize={filteredAddressList.length}
+            LoadMoreLabel={t('community.votersTable.loadMore')}
+            explorerURL={CHAIN_METADATA[network].explorer}
+          />
+        ) : (
+          // this view is temporary until designs arrive
+          <span>{t('AddressModal.noAddresses')}</span>
+        )}
+      </Container>
+      {/*<ModalHeader>*/}
```
cgero-eth: This commented out code can be removed, right?

src/components/filteredAddressList/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+import {TokenVotingWalletField} from 'components/addWallets/row';
+import {MultisigWalletField} from 'components/multisigWallets/row';
+import {getUserFriendlyWalletLabel} from 'utils/library';
+import {CHAIN_METADATA} from 'utils/constants';
+import {useNetwork} from 'context/network';
+
+type FilteredAddressListProps = {
+  wallets: TokenVotingWalletField[] | MultisigWalletField[];
+  tokenSymbol?: string;
+  // tokenMembership: boolean;
```
cgero-eth: Let's remove this if not needed
```suggestion
```

src/containers/communityAddressesModal/index.tsx
```diff
@@ -75,29 +24,10 @@ const CommunityAddressesModal: React.FC<CommunityAddressesModalProps> = ({
       onClose={close}
       data-testid="communityModal"
     >
-      <ModalHeader>
```
cgero-eth, line 78: The `<ModalHeader />` and `<Container />` components below can now be removed

cgero-eth (commented):
I reviewed roughly half of the changed files and left some initial comments. I saw that you're still actively pushing changes to the branch, probably it makes sense for me to continue the review when you're at least done with all the type and eslint issues, what do you think?

src/containers/configureCommunity/proposalCreation.tsx
```diff
@@ -0,0 +1,54 @@
...
+    name: ['membership', 'isCustomToken'],
+  });
+
+  return (
+    <div>
+      {membership === 'multisig' && (
+        <>
+          <MultisigEligibility />
+        </>
+      )}
```
cgero-eth: Using the React fragment is redundant here, right?
```suggestion
      {membership === 'multisig' && <MultisigEligibility />}
```

src/containers/configureCommunity/proposalCreation.tsx
```diff
@@ -0,0 +1,54 @@
...
+      {membership === 'token' && (
+        <>
+          <TokenProposalCreation />
```
cgero-eth: I would suggest to merge the logic of the `ProposalCreation` and `TokenProposalCreation` components in one component instead of creating two small components:
```
{membership === 'token' && isAllowedToConfigureVotingEligibility && <SelectEligibility />}
```

src/containers/delegationGatingMenu/delegationGatingMenu.tsx
```diff
@@ -98,6 +99,10 @@ export const DelegationGatingMenu: React.FC = () => {
     needsSelfDelegation,
   });
 
+  if ((daoDetails?.plugins[0].id as PluginTypes) === GaselessPluginName) {
+    return;
+  }
```
cgero-eth: Do we accidentally open this dialog for the gaseless plugin? We should fix this by not opening the dialog instead of not rendering it

src/containers/configureCommunity/proposalCreation.tsx
selankon: Well, it seems that the `MultisigEligibility` component appears also on the settings edition, which i am not sure if use a pretty different form contexts. 

As we have not yet implemented the editSettings page, I will let this open until is implemented to check if this component can be refactored.

src/containers/goLive/index.tsx
selankon: Mmm I'm not sure: the committee is a conditional step, so on this way, you can see that is a "condtional step". Could be done also on the component itself, but on the way its implemented, you avoid to execute some of the hooks that are on top of the component. 

Imo this can be let like this

src/containers/editSettings/majorityVoting.tsx
```diff
@@ -66,7 +68,9 @@ export const EditMvSettings: React.FC<EditMvSettingsProps> = ({daoDetails}) => {
       pluginAddress,
       pluginType,
     });
-  const votingSettings = pluginSettings as VotingSettings | undefined;
+  const votingSettings = isGasless
+    ? (pluginSettings as GaslessPluginVotingSettings)
+    : (pluginSettings as VotingSettings | undefined);
```
cgero-eth: During the initial render, the `pluginSettings` variable will be `undefined`
```suggestion
  const votingSettings = pluginSettings as GaslessPluginVotingSettings | VotingSettings | undefined;
```

src/containers/editSettings/majorityVoting.tsx
```diff
@@ -436,6 +513,21 @@ export const EditMvSettings: React.FC<EditMvSettingsProps> = ({daoDetails}) => {
                   <ConfigureCommunity isSettingPage />
                 </AccordionContent>
               </AccordionItem>
+
+              {/*todo(kon): this is not going to work until the executive committe list can be retrieved */}
+              {/*<AccordionItem*/}
+              {/*  type="action-builder"*/}
```
cgero-eth, line 530: We usually avoid pushing commented-out code as it is a bad practice, can we get rid of all the commented code before merging this PR to develop?

src/hooks/useDaoProposal.tsx
cgero-eth, line 1: File is empty and can be removed

src/utils/committeeVoting.ts
```diff
@@ -0,0 +1,64 @@
...
+  t: TFunction
+) {
+  let label = '';
+  if (
+    proposal.status === 'Pending' ||
+    proposal.status === 'Active' ||
+    proposal.status === 'Succeeded'
+  ) {
+    // Uncomment line below is causing SyntaxError: ambiguous indirect export: default 🤷‍♀️
+    // const locale = (Locales as Record<string, Locale>)[i18n.language];
```
cgero-eth: Does this still happen?

src/services/aragon-sdk/queries/use-voting-settings.ts
```diff
@@ -57,6 +61,13 @@ export function isMultisigVotingSettings(
   return settings ? 'minApprovals' in settings : false;
 }
 
+export function isGaslessVotingSettings(
+  settings: SupportedVotingSettings | undefined | null
+): settings is GaslessPluginVotingSettings {
+  if (!settings || Object.keys(settings).length === 0) return false;
+  return 'onlyMultisigProposalCreation' in settings;
```
cgero-eth: We should not be passing an empty object to this function:
```suggestion
  return settings ? 'onlyMultisigProposalCreation' in settings : false;
```

src/containers/proposalStepper/index.tsx
```diff
@@ -153,7 +153,7 @@ const ProposalStepper: React.FC<ProposalStepperType> = ({
         wizardTitle={t('newProposal.configureActions.heading')}
         wizardDescription={t('newProposal.configureActions.description')}
         isNextButtonDisabled={!areActionsValid}
-        {...(type === 'os-update' && {skipStep: true, hideWizard: true})}
+        // {...(type === 'os-update' && {skipStep: true, hideWizard: true})}
```
cgero-eth: Why is this commented out?

src/hooks/useFunctionStepper.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export interface StepData {
+  status: StepStatus;
+  errorMessage?: string;
+}
+
+export type GenericKeyEnum = string | number | symbol;
+
+export type StepsMap<X extends GenericKeyEnum> = Record<X, StepData>;
```
cgero-eth: If we're not actually using `number | symbol` types, I would avoid using this `GenericKeyEnum` type everywhere and write it as `extends string`. It would also be better if, instead of just `X`, we give a better name to the generic type so that it's easier to understand what it represents, like:
```suggestion
export type StepsMap<TStepKey extends string> = Record<TStepKey, StepData>;
```

src/components/executionExpirationtime/index.tsx
selankon: I cannot see `executionEpirationtime` string anywhere on the whole app, where did you find it?

.env.staging
selankon: a8227c7f829332c2105b782a48f4b11fbcef430a

.gitignore
selankon: a8227c7f829332c2105b782a48f4b11fbcef430a

src/components/executionExpirationtime/index.tsx
selankon: a8227c7f829332c2105b782a48f4b11fbcef430a

src/components/filteredAddressList/index.tsx
selankon: a8227c7f829332c2105b782a48f4b11fbcef430a

src/containers/communityAddressesModal/index.tsx
selankon: a8227c7f829332c2105b782a48f4b11fbcef430a

src/components/executionExpirationtime/index.tsx
cgero-eth: @selankon It's on the folder and component name, the `t` on the `time` word should be uppercased

src/components/filteredAddressList/index.tsx
selankon: a8227c7f829332c2105b782a48f4b11fbcef430a

src/components/executionExpirationtime/index.tsx
selankon: @elboletaire just show me that you are referring to the camel case, I will fix it :P

src/utils/proposals.ts
```diff
@@ -431,7 +440,78 @@ export function getLiveProposalTerminalProps(
...
+      symbol: proposal.token.symbol,
+    };
+
+    // voters
+    voters =
+      proposal.voters?.map(voter => {
+        return {wallet: voter, src: voter, option: 'none'} as VoterType;
+      }) ?? [];
+
+    console.log(voters);
```
cgero-eth:
```suggestion
```

src/hooks/useDaoToken.tsx
```diff
@@ -13,7 +21,12 @@ export function useDaoToken(
   const [error, setError] = useState<Error>();
   const [isLoading, setIsLoading] = useState(false);
 
-  const pluginClient = usePluginClient('token-voting.plugin.dao.eth');
+  const {data: daoDetails, isLoading: daoDetailsLoading} = useDaoDetailsQuery();
+  const {id: pluginType} =
+    daoDetails?.plugins[0] || ({} as InstalledPluginListItem);
```
cgero-eth: Looks like the cast to `InstalledPluginListItem` is not needed
```suggestion
  const {id: pluginType} = daoDetails?.plugins[0] ?? {};
```

src/context/createGaslessProposal.tsx
```diff
@@ -0,0 +1,267 @@
...
+          .filter(key => isNaN(Number(key)))
+          .map((key, i) => ({
+            title: key,
+            value: i,
+          }))
+      );
+
+      const cost = await vocdoniClient.calculateElectionCost(election);
+
+      console.log('DEBUG', 'Estimated cost', cost, 'balance', account!.balance);
```
cgero-eth: Are these debug console logs still needed?

src/containers/settings/gaslessVoting/index.tsx
```diff
@@ -0,0 +1,200 @@
...
+import {PluginTypes} from 'hooks/usePluginClient';
+import {useVotingSettings} from 'services/aragon-sdk/queries/use-voting-settings';
+import {IPluginSettings} from 'pages/settings';
+import {getDHMFromSeconds} from 'utils/date';
+import {GaslessPluginVotingSettings} from '@vocdoni/gasless-voting';
+
+const GaslessVotingSettings: React.FC<IPluginSettings> = ({daoDetails}) => {
+  const {t} = useTranslation();
+  // const {network} = useNetwork();
+  // const navigate = useNavigate();
```
cgero-eth, line 21: Let's remove all the commented out code on this file as well 🙏🏻

src/containers/transactionModals/gaslessVotingModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+}: {
+  vote: VoteProposalParams | undefined;
+  showVoteModal: boolean;
+  setShowVoteModal: React.Dispatch<React.SetStateAction<boolean>>;
+  setVoteSubmitted: React.Dispatch<React.SetStateAction<boolean>>;
+  onVoteSubmitted: (
+    proposalId: string,
+    vote: VoteValues,
+    isApproval?: boolean
+  ) => Promise<void>;
```
cgero-eth: Let's move the props definition to a separate type and update the return type of the function to `React.FC`:
```
type GaslessVotingModalProps = { ... }

export const GasslessVotingModal: React.FC<GaslessVotingModalProps> = (...) => { ... }
```

src/containers/transactionModals/gaslessVotingModal.tsx
```diff
@@ -0,0 +1,116 @@
...
+  vote,
+  showVoteModal,
+  setShowVoteModal,
+  setVoteSubmitted,
+  onVoteSubmitted,
+}: {
+  vote: VoteProposalParams | undefined;
+  showVoteModal: boolean;
+  setShowVoteModal: React.Dispatch<React.SetStateAction<boolean>>;
+  setVoteSubmitted: React.Dispatch<React.SetStateAction<boolean>>;
```
cgero-eth: The two callbacks can be probably simplified as:
```suggestion
  setShowVoteModal: (showModal: boolean) => void;
  setVoteSubmitted: (voteSubmitted: boolean) => void;
```

src/containers/votingTerminal/committeeVotingTerminal.tsx
```diff
@@ -0,0 +1,300 @@
...
+}: {
+  votingStatusLabel: string;
+  votingTerminal: ReactNode;
+  proposal: GaslessVotingProposal;
+  pluginAddress: string;
+  statusRef: React.MutableRefObject<{
+    wasNotLoggedIn: boolean;
+    wasOnWrongNetwork: boolean;
+  }>;
+  pluginType: PluginTypes;
```
cgero-eth: Let's move these props type definition to a `CommitteeVotingTerminalProps` type

src/services/aragon-sdk/queries/use-proposal.ts
```diff
@@ -7,26 +7,46 @@ import {
...
+    }
+    data = await (client as GaslessVotingClient).methods.getProposal(
+      params!.id,
+      '',
+      ''
+    );
+    return data;
+  }
+  params = params as IFetchProposalParams;
+  data = await client?.methods.getProposal(params.id);
```
cgero-eth: Looks like the `daoName` and `daoAddress` params of the `GaslessVotingClient.getProposal` function are optional. Also this function is not called if `params.id` is not set:
```suggestion
  const data = await client?.methods.getProposal(params.id);
```

src/pages/proposeSettings.tsx
```diff
@@ -600,6 +600,9 @@ const ProposeSettingWrapper: React.FC<Props> = ({
     }
     if (!proposalCreationData) return;
 
+    // todo(kon): implement this
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
     return pluginClient?.estimation.createProposal(proposalCreationData);
```
cgero-eth: Instead of adding a `@ts-ignore` directive and disable the eslint rule, I would just return in case the client is a gasless voting client and write a comment:
```suggestion
    // The propose settings flow is not currently handled by the gasless voting client
    if (!proposalCreationData || isGaslessVotingClient(pluginClient)) return;

    return pluginClient?.estimation.createProposal(proposalCreationData);
```

src/pages/proposeSettings.tsx
```diff
@@ -652,8 +655,24 @@ const ProposeSettingWrapper: React.FC<Props> = ({
...
+    //   proposalIterator =
+    //     pluginClient.methods.createProposal(proposalCreationData);
+    // }
+
+    // todo(kon): implement this
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const proposalIterator = pluginClient.methods.createProposal(
+      proposalCreationData as any
+    );
```
cgero-eth: Same here:
```suggestion
    // The propose settings flow is not currently handled by the gasless voting client
    if (isGaslessVotingClient(pluginClient)) {
      return;
    }

    const proposalIterator =
      pluginClient.methods.createProposal(proposalCreationData);
```

src/context/prepareUpdate.tsx
```diff
@@ -143,9 +143,13 @@ const PrepareUpdateProvider: React.FC<{children: ReactElement}> = ({
   // estimate creation fees
   const estimateCreationFees = useCallback(async () => {
     if (!daoUpdateData) return;
-    if (showModal.type === 'plugin')
+    if (showModal.type === 'plugin') {
+      // todo(kon): implement this on the min sdk
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
cgero-eth: I wouldn't add a `@ts-ignore` here but instead just return void if the client is a `GaslessVotingClient`, same for the `preparePluginIterator` handling below

src/pages/createDAO.tsx
```diff
@@ -265,6 +279,40 @@ export const CreateDAO: React.FC = () => {
...
+  //   // if multisig
+  //   if (membership === 'multisig') {
+  //     if (!['multisig', 'anyone'].includes(eligibilityType)) {
+  //       return false;
+  //     }
+  //     return true;
+  //   } else {
+  //     return !errors.eligibilityTokenAmount;
+  //   }
+  // }, [eligibilityType, errors.eligibilityTokenAmount, membership]);
```
cgero-eth: This can be removed, right?
```suggestion
```

src/pages/proposal.tsx
```diff
@@ -224,10 +234,10 @@ export const Proposal: React.FC = () => {
 
   // set editor data
   useEffect(() => {
-    if (proposal && editor) {
+    if (proposal && editor && proposal.metadata.description) {
       editor.commands.setContent(
         // Default list of allowed tags and attributes - https://www.npmjs.com/package/sanitize-html#default-options
-        sanitizeHtml(proposal.metadata.description, {
+        sanitizeHtml(proposal.metadata.description!, {
```
cgero-eth: We're already checking that `proposal.metadata.description` is not null, right?
```suggestion
        sanitizeHtml(proposal.metadata.description, {
```

src/pages/proposal.tsx
```diff
@@ -709,42 +767,33 @@ export const Proposal: React.FC = () => {
               />
             )}
 
-          {votingSettings && (
-            <VotingTerminal
-              status={proposalStatus}
+          {votingSettings && isGaslessProposal(proposal) ? (
+            <CommitteeVotingTerminal
+              votingTerminal={<VTerminal />}
```
cgero-eth: Instead of passing the `VotingTerminal` as prop, I would suggest to pass it as a child component:
```
<CommitteeVotingTerminal
  proposal={proposal}
  {...}
>
  <VotingTerminal />
</CommitteeVotingTerminal>
```

src/context/useGaslessVoting.tsx
```diff
@@ -0,0 +1,229 @@
...
+      PUBLISH_VOTE: {
+        status: StepStatus.WAITING,
+      },
+    } as GaslessVotingSteps,
+  });
+
+  const submitVote = useCallback(
+    async (vote: VoteProposalParams, electionId: string) => {
+      const vocVote = new Vote([vote.vote - 1]); // See values on the enum, using vocdoni starts on 0
+      console.log('DEBUG', 'ElectionId and vote', electionId, vocVote);
```
cgero-eth: Let's remove the `console.log` if they're not needed anymore

src/context/createProposal.tsx
```diff
@@ -574,7 +647,36 @@ const CreateProposalWrapper: React.FC<Props> = ({
...
           token: daoToken ?? null,
           votes,
-        };
+        } as TokenVotingProposal;
+        proposalStorage.addProposal(CHAIN_METADATA[network].id, proposal);
+      } else if (isGaslessVotingSettings(votingSettings)) {
+        const proposal = {
+          ...baseParams,
+          // todo(kon)(cache): is this needed to be implemented?
+          // expirationDate: Date;
```
cgero-eth: We probably need this to display the proposal even when not yet indexed by subgraph

src/context/createProposal.tsx
```diff
@@ -727,20 +881,42 @@ const CreateProposalWrapper: React.FC<Props> = ({
...
+            gaslessGlobalState !== StepStatus.LOADING
+          }
+          maxFee={maxFee}
+          averageFee={averageFee}
+          gasEstimationError={gasEstimationError}
+          tokenPrice={tokenPrice}
+          title={t('TransactionModal.createProposal')}
+          // buttonLabel={t('TransactionModal.createProposal')}
+          // buttonLabelSuccess={t('TransactionModal.goToProposal')}
+          // disabledCallback={disableActionButton}
```
cgero-eth: Let's remove these if not currently needed:
```suggestion
```

cgero-eth (commented):
That's the last round of feedbacks regarding the code and the changed files, I will now be doing some local tests to double check that current functionalities are not affected

src/components/StepperProgress/index.tsx
selankon: https://github.com/aragon/app/commit/a8227c7f829332c2105b782a48f4b11fbcef430a

src/components/executionExpirationtime/index.tsx
selankon: (on your first comment you wrote `executionEpirationtime` which has a typo, and i was looking for this :P )

src/containers/configureCommunity/proposalCreation.tsx
selankon: Absolutely yes, maybe i split this for some purpose during the coding and I forget to merge it again. I will fix it

src/containers/delegationGatingMenu/delegationGatingMenu.tsx
selankon: True, this is done here https://github.com/aragon/app/blob/19d3c319778fb51ae02125459e7382b02fbf4532/src/pages/proposal.tsx#L573-L577 and shouldn't be open.

I'll delete this lines

src/containers/editSettings/majorityVoting.tsx
selankon: Ok, but we are awaiting a the GraphQl implementation to have a feature that we need to implement the commented code. Hopefully before merge. I will let this comment open as reminder

src/utils/committeeVoting.ts
selankon: Not anymore.  fixing

src/containers/proposalStepper/index.tsx
selankon: Commented by development purposes by @emmdim . 

This is because if there are a pending os-update, you cannot add actions, and we disabled it for development purposes. I'll let this comment open as a reminder

src/hooks/useFunctionStepper.tsx
selankon: Mmm not sure why I add `number | symbol` as key types... Maybe at the beginning i was thinking on other behavior. Thanks for point to it!

I implement your solution, but i have a question: could not be even better to still use a type alias to store the `StepsMap` key type?

```ts
export type GenericKeyEnum = string;
```

On this way, you will ensure readability or scalability for future refactors, imho

selankon: f947d5a8

src/utils/proposals.ts
selankon: f947d5a8

src/services/aragon-sdk/queries/use-voting-settings.ts
selankon: f947d5a8

src/utils/committeeVoting.ts
selankon: f947d5a8

src/hooks/useDaoProposal.tsx
selankon: f947d5a8

src/containers/editSettings/majorityVoting.tsx
selankon: f947d5a8

src/containers/delegationGatingMenu/delegationGatingMenu.tsx
selankon: f947d5a8

src/containers/configureCommunity/proposalCreation.tsx
selankon: f947d5a8

src/containers/configureCommunity/proposalCreation.tsx
selankon: f947d5a8

src/components/executionExpirationtime/index.tsx
selankon: f947d5a8

src/containers/editSettings/majorityVoting.tsx
cgero-eth: @selankon Sounds good, we can also keep it and merge the PR if this is going to be fixed soon 👍🏻

.env.development
selankon: Lol, something on my OS is adding a new line at the end of the file. I tried with different editors without success, and for some editors the line is not there, so strange i have to research deeper 😵

src/components/executionExpirationtime/index.tsx
cgero-eth: Oops 🙈

src/hooks/useDaoToken.tsx
selankon: I took this line from here

https://github.com/aragon/app/blob/develop/src/context/govTokensWrapping.tsx#L53C1-L54

Should i modify also there? Is Aragon app already existing code

src/context/createGaslessProposal.tsx
selankon: I guess not!

src/hooks/useFunctionStepper.tsx
cgero-eth: @selankon You're right, having a type alias would make the functionality/component easier to extend if we plan to use other types as step keys

src/containers/settings/gaslessVoting/index.tsx
selankon: This is for the same as the commented here: https://github.com/aragon/app/pull/1074#discussion_r1384549315

I'll let this open as a reminder

src/hooks/useDaoToken.tsx
cgero-eth: @selankon If we don't get type issues there as well yes please 🙏🏻

selankon: 7abb8407

src/context/createGaslessProposal.tsx
selankon: 7abb8407

src/containers/transactionModals/gaslessVotingModal.tsx
selankon: 7abb8407

src/pages/proposal.tsx
selankon: 7abb8407

src/pages/proposal.tsx
selankon: 7abb8407

src/context/useGaslessVoting.tsx
selankon: 7abb8407

src/context/createProposal.tsx
selankon: 7abb8407

.env.development
cgero-eth: A new line at the end of the file is fine (I think my IDE does it as well), but here the new line is on the first line

src/context/createProposal.tsx
selankon: d691c862

.env.development
elboletaire: Not your os @selankon, but the editorconfig defined in the project: https://github.com/aragon/app/blob/develop/.editorconfig#L9

elboletaire: Correction on my previous comment (now removed): indeed the new line at the beginning of the file is really weird. FYI the new line at the end is there [due to `.editorconfig` defined that way in the project](https://github.com/aragon/app/blob/develop/.editorconfig#L9) (btw kinda curious that it has the new line at the end defined, but the `.editorconfig` file itself does not have that new line at the end 😅).

src/context/createGaslessProposal.tsx
```diff
@@ -0,0 +1,249 @@
...
+  metadata,
+  data,
+  census,
+}: IProposalToElectionProps): UseCreateElectionProps => {
+  return {
+    title: metadata.title,
+    description: metadata.description,
+    question: metadata.summary,
+    startDate: data?.startDate ?? undefined,
+    endDate: data?.endDate ?? new Date(),
```
emmdim: @selankon as we have repeated various times, there was even an issue about it that you closed, the endDate cannot be the current date

emmdim: You either need to add minDuration or ensure that the value is going to be there

cgero-eth:
@emmdim @selankon I run some smoke tests locally and didn't find any issue. I only saw this new error on the console for some DAO page (e.g. http://localhost:3000/#/daos/ethereum/safari.dao.eth/dashboard):
Screenshot 2023-11-09 at 13 17 29

.env.example
```diff
@@ -23,3 +23,5 @@ VITE_WALLET_CONNECT_PROJECT_ID=YOUR_VALUE_HERE
 
 # Covalent - unified API bringing visibility to billions of Web3 data points https://www.covalenthq.com/platform/auth/login/
 VITE_COVALENT_API_KEY=YOUR_VALUE_HERE
+
+VITE_VOCDONI_ENV='stg'
```
cgero-eth, line 27: This should probably be on the `.env.staging` file and not here

package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "dev": "vite --host",
     "dev:watch": "cross-env WATCH=true vite --host",
-    "build": "tsc && cross-env NODE_ENV=production node --max_old_space_size=16384 ./node_modules/vite/bin/vite.js build --sourcemap true > build.txt",
+    "build": "tsc && cross-env NODE_ENV=development node --max_old_space_size=16384 ./node_modules/vite/bin/vite.js build --sourcemap true > build.txt",
```
cgero-eth: Just a reminder to switch it back before merging the PR
```suggestion
    "build": "tsc && cross-env NODE_ENV=production node --max_old_space_size=16384 ./node_modules/vite/bin/vite.js build --sourcemap true > build.txt",
```

cgero-eth (approved):

PR #1105, opened 2023-10-30
Feature: APP-2573 - Arbitrum One & Arbitrum Goerli Support

Fabricevladimir:
## Description

- Adds support for Arbitrum One and Arbitrum Goerli
- Adds feature flags for both chains

Task: [APP-2573](https://aragonassociation.atlassian.net/browse/APP-2573)

## Type of change

## Checklist:

[APP-2573]: https://aragonassociation.atlassian.net/browse/APP-2573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/tokens.ts
```diff
@@ -89,12 +94,32 @@ export async function getPastVotingPower(
...
   blockNumber: number,
-  provider: EthersProviders.Provider
+  provider: EthersProviders.Provider,
+  network: SupportedNetworks
 ) {
-  const contract = new ethers.Contract(address, votesUpgradeableABI, provider);
   try {
-    return (await contract.getPastVotes(account, blockNumber)) as BigNumber;
-  } catch (err) {
+    // for Arbitrum, there is a mismatch between the l1 and l2 block numbers.
```
cgero-eth, line 101: Proposals are also created on Arbitrum, right? Why do we have a L1 / L2 block number mismatch if we're fetching the past votes for the proposal creation block number?

src/context/providers.tsx
```diff
@@ -118,14 +118,17 @@ export function getAlchemyProvider(
 
   const networkishOptions: Networkish = {
     chainId: CHAIN_METADATA[network]?.id,
-    name: translateToNetworkishName(network),
+    name:
+      network !== 'arbitrum-goerli'
+        ? translateToNetworkishName(network)
+        : 'arbitrum-goerli',
```
cgero-eth, line 124: The `translateToNetworkishName` handles the `arbitrum-goerli` case, we can just use the function, right?

src/context/providers.tsx
```diff
@@ -152,14 +155,17 @@ export function getInfuraProvider(
...
```
cgero-eth, line 161: Same here

src/services/token/token-service.ts
```diff
@@ -212,8 +213,12 @@ class TokenService {
     }
 
     return data.items.flatMap(({native_token, ...item}) => {
-      // ignore zero balances if indicated
-      if (ignoreZeroBalances && BigNumber.from(item.balance).isZero())
+      if (
+        ignoreZeroBalances &&
+        ((native_token && nativeTokenBalance === BigInt(0)) ||
```
cgero-eth, line 218: Why do we need the `nativeTokenBalance` here? Can't we use the `item.balance`?

src/services/token/queries/use-token.ts
```diff
@@ -57,11 +59,24 @@ export const useTokenList = (
 
 export const useTokenBalances = (
   params: IFetchTokenBalancesParams,
-  options?: UseQueryOptions<AssetBalance[] | null>
+  options: UseQueryOptions<AssetBalance[] | null> = {}
 ) => {
+  const {data: nativeToken, isFetched} = useBalance({
+    address: params.address as `0x${string}`,
```
cgero-eth: You can import the `Address` type from `wagmi`:
```suggestion
    address: params.address as Address,
```

cgero-eth (commented):
Thanks for the updates @Fabricevladimir , let some comments, some of them are just questions

src/utils/tokens.ts
Fabricevladimir: The subgraph is storing the L2 block number for the proposal creation. When we go directly to the contract, apparently it expects the L1 block number, which we don't have. This is a work around where we're making the provider go to the L2 block number to get the votes at that proposal creation block number. 

A better solution, which OS is already aware of, is to provide a mechanism to use the information from the subgraph since it indexes all voting power.

src/context/providers.tsx
Fabricevladimir: Unfortunately, nope. `translateToNetworkishName` returns `arbitrumGoerli` but both alchemy and infura are expecting `arbitrum-goerli` for whatever reason

src/context/providers.tsx
Fabricevladimir: See above response

src/services/token/token-service.ts
Fabricevladimir: Apparently covalent has had an issue with indexing the native token balance and so now we're querying the contract directly for the balance and using that value instead. @sepehr2github can you provide more context?

src/context/providers.tsx
cgero-eth: Okay I see 🫤 Ideally we should remove this `translateToNetworkishName` map and add this information directly to the `CHAIN_METADATA` record, maybe with two fields: `sdkNetwork` and `providerNetwork` or something similar as apparently the two don't match. Can we open a Jira ticket to refactor this?

src/utils/tokens.ts
cgero-eth: > it expects the L1 block number

That's pretty weird 🤔 Do you know if there's a Jira ticket on the OS board that we can link here?

src/services/token/token-service.ts
cgero-eth: I see, I tried both `arbitrum` and `arbitrum-goerli` Covalent APIs [here](https://www.covalenthq.com/docs/api/balances/get-token-balances-for-address/) and it correctly returned the native balance

sepehr2github: It's not a daily issue, It comes and go. James is also aware of that and happened to me few days ago. The issue was the token indexed and was in list but the balance was zero

src/context/providers.tsx
Fabricevladimir: Here's the corresponding ticket: https://aragonassociation.atlassian.net/browse/APP-2591

src/services/token/token-service.ts
cgero-eth: @sepehr2github So this is not specific to Arbitrum?

cgero-eth: @Fabricevladimir Fine for me to merge the PR as it is, let's just write a comment to document why we're fetching the native balance as well. Hopefully we'll remove such workaround when having a backend service handling this logic.

Fabricevladimir: Commented and ticketed [here](https://aragonassociation.atlassian.net/browse/APP-2592).

cgero-eth (approved):
Thanks @Fabricevladimir ✅

jamesej (commented):
Did a second review to check, code looks good.

PR #1110, opened 2023-11-03
feat: APP-2457 token based dao proposals can start immediately with n…

babarella:
As a DAO operator in a token based DAO, I want my proposals to start immediately after creating them as opposed to having to wait 10 minutes so that I can vote immediately after creating the proposal.

Task: [APP-2457](https://aragonassociation.atlassian.net/browse/APP-2457)

## Type of change

## Checklist:

[APP-2457]: https://aragonassociation.atlassian.net/browse/APP-2457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/reviewProposal/index.tsx
+++ b/src/containers/reviewProposal/index.tsx
@@ -155,7 +155,6 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
       durationHours,
       durationMinutes,
       durationSwitch,
-      startSwitch,
       endDate,
       endTime,
       endUtc,
--- a/src/containers/reviewProposal/index.tsx
+++ b/src/containers/reviewProposal/index.tsx
@@ -83,18 +83,15 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
     const {startSwitch, startDate, startTime, startUtc} = values;
 
     if (startSwitch === 'now') {
-      const startMinutesDelay = isMultisig ? 0 : 10;
       return new Date(
-        `${getCanonicalDate()}T${getCanonicalTime({
-          minutes: startMinutesDelay,
-        })}:00${getCanonicalUtcOffset()}`
+        `${getCanonicalDate()}T${getCanonicalTime()}:00${getCanonicalUtcOffset()}`
       );
     } else {
       return Date.parse(
         `${startDate}T${startTime}:00${getCanonicalUtcOffset(startUtc)}`
       );
     }
-  }, [isMultisig, values]);
+  }, [values]);
 
   const formattedStartDate = useMemo(() => {
     const {startSwitch} = values;
@@ -182,17 +179,14 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
 
     // adding 10 minutes to offset the 10 minutes added by starting now
     if (startSwitch === 'now') {
-      const startMinutesDelay = isMultisig ? 0 : 10;
-      endDateTime = new Date(
-        endDateTime.getTime() + minutesToMills(startMinutesDelay)
-      );
+      endDateTime = new Date(endDateTime.getTime());
     }
 
     return `~${format(
       endDateTime,
       KNOWN_FORMATS.proposals
     )} ${getFormattedUtcOffset()}`;
-  }, [isMultisig, values]);
+  }, [values]);
 
   const terminalProps = useMemo(() => {
     if (votingSettings) {
--- a/src/containers/reviewProposal/index.tsx
+++ b/src/containers/reviewProposal/index.tsx
@@ -176,11 +176,6 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
       );
     }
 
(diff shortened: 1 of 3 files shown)


src/containers/reviewProposal/index.tsx
```diff
@@ -182,17 +178,14 @@ const ReviewProposal: React.FC<ReviewProposalProps> = ({
 
     // adding 10 minutes to offset the 10 minutes added by starting now
     if (startSwitch === 'now') {
-      const startMinutesDelay = isMultisig ? 0 : 10;
-      endDateTime = new Date(
-        endDateTime.getTime() + minutesToMills(startMinutesDelay)
-      );
+      endDateTime = new Date(endDateTime.getTime());
     }
```
Fabricevladimir: Is all this still necessary since `startSwitch` no longer affects the end date?

src/context/createProposal.tsx
```diff
@@ -379,15 +374,11 @@ const CreateProposalWrapper: React.FC<Props> = ({
       }
 
       if (startSwitch === 'now') {
-        endDateTime = new Date(
-          endDateTime.getTime() + minutesToMills(startMinutesDelay)
-        );
+        endDateTime = new Date(endDateTime.getTime());
       } else {
```
Fabricevladimir: This is overwriting what we have set up in lines 341.

src/context/createProposal.tsx
```diff
@@ -379,15 +374,11 @@ const CreateProposalWrapper: React.FC<Props> = ({
...
         if (startDateTime.valueOf() < new Date().valueOf()) {
           startDateTime = new Date(
```
Fabricevladimir, line 404: Why not set `startDateTime` the value it would get if the `startSwitch` is "now"? This seems to be slightly dangerous in terms of the start time already being in the past by the time the tx is actually executed, which will cause it to revert.

You could keep the delay from the original start time in another variable so you can recalculate the end time, although we really should be making the user fix their issue.

src/context/createProposal.tsx
```diff
@@ -410,22 +401,20 @@ const CreateProposalWrapper: React.FC<Props> = ({
...
+       * to keep startDate and endDate undefined, so it's automatically evaluated.
        * If we just provide "Date.now()", than after user still goes through the flow
        * it's going to be date from the past. And SC-call evaluation will fail.
        */
-      const finalStartDate =
-        startSwitch === 'now' && isMultisigVotingSettings(votingSettings)
-          ? undefined
-          : startDateTime;
+      const finalStartDate = startSwitch === 'now' ? undefined : startDateTime;
+      const finalEndDate = startSwitch === 'now' ? undefined : endDateTime;
```
Fabricevladimir: Isn't this overwriting the user-defined endDate? Doesn't this automatically set the proposal to be the minimum duration set on the voting plugin?

src/pages/proposeSettings.tsx
```diff
@@ -516,15 +511,11 @@ const ProposeSettingWrapper: React.FC<Props> = ({
         }
 
         if (startSwitch === 'now') {
```
Fabricevladimir: We're calculating start time twice. See comments on `CreateProposal.tsx`

src/pages/proposeSettings.tsx
```diff
@@ -547,22 +538,21 @@ const ProposeSettingWrapper: React.FC<Props> = ({
         }
 
         /**
-         * For multisig proposals, in case "now" as start time is selected, we want
-         * to keep startDate undefined, so it's automatically evaluated.
+         * In case "now" as start time is selected, we want
+         * to keep startDate and endDate undefined, so it's automatically evaluated.
```
Fabricevladimir: Why end date undefined? Maybe they want to start the proposal now but have it ended on some day later than the plugin minimum duration.

Fabricevladimir (commented):
Thanks! This was much needed; left a couple of comments/questions.

src/context/createProposal.tsx
babarella: Because in case `startSwitch === now`, we will actually never need  `startDateTime` anymore, since we want to start immediately. And that's only possible by passing `undefined` value to SDK as `startDate`.

src/pages/proposeSettings.tsx
babarella: In case of `startTime` we are not, as I believe code above is just some sensible default for further possible calculations. But we do with `endDateTime` here, so I patched the place.

src/context/createProposal.tsx
babarella: Very good insight, thank you

src/pages/proposeSettings.tsx
babarella: Very good insight, thank you

Fabricevladimir (approved):
Looking good! Nice work 🚀

PR #1115, opened 2023-11-06
feat: APP-2561 - Update delegation gating error messages

cgero-eth:
## Description

Task: [APP-2561](https://aragonassociation.atlassian.net/browse/APP-2561)

## Type of change

## Checklist:

[APP-2561]: https://aragonassociation.atlassian.net/browse/APP-2561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/aragon-sdk/queries/use-member.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+`;
+
+const fetchMember = async (
+  {pluginAddress, blockNumber}: IFetchMemberParams,
+  client?: TokenVotingClient
+): Promise<TokenVotingMember> => {
+  invariant(client != null, 'fetchMember: client is not defined');
+  const userAddress = await client.web3.getSigner().getAddress();
```
Fabricevladimir: Why get the address here and not via a param?

src/pages/proposal.tsx
```diff
@@ -516,14 +491,10 @@ export const Proposal: React.FC = () => {
...
 
-  // this tracks whether the voting button should be enabled
-  const enableVotingButton =
-    proposal?.status === ProposalStatus.ACTIVE && // active proposal
-    (address == null || // wallet disconnected
-      (voted === false && canVote === true) || // haven't voted and can vote
-      (isTokenVotingPlugin && voted && canRevote === true) || // voted but vote replacement enabled
-      // can vote but no voting power (voted should status should never be true if shouldDisplayDelegationVotingGating)
-      (voted === false && shouldDisplayDelegationVoteGating === true));
+  const votingDisabled =
```
Fabricevladimir, line 494: This is changing like the wind

src/pages/proposal.tsx
```diff
@@ -580,24 +552,17 @@ export const Proposal: React.FC = () => {
...
-    alertMessage = isErc20VotingProposal(proposal)
-      ? t('votingTerminal.status.ineligibleTokenBased', {
-          token: proposal.token.name,
-        })
-      : t('votingTerminal.status.ineligibleWhitelist');
-  }
+    !isOnWrongNetwork && // on proper network
+    !voted && // haven't voted
+    !canVote && // cannot vote
+    isMultisigPlugin; // is multisig plugin
```
Fabricevladimir: Really this should be right up there with whether the proposal is active lol

src/pages/proposal.tsx
```diff
@@ -554,15 +525,16 @@ export const Proposal: React.FC = () => {
     } else if (isOnWrongNetwork) {
       open('network');
       statusRef.current.wasOnWrongNetwork = true;
-    } else if (shouldDisplayDelegationVoteGating) {
-      return open('delegationGating');
+    } else if (displayDelegationVoteGating) {
+      return open('delegationGating', {proposal});
```
Fabricevladimir, line 529: Considering the modals have state and callbacks, we should refactor this page so the wallet/network situation doesn't rely on all these refs.

Fabricevladimir (commented):
Nice work! Left a few comments and an actual question

src/services/aragon-sdk/queries/use-member.ts
cgero-eth: Mm.. you're right, a `getMember` functionality from the SDK will probably get the `address` as a param, let me update this 👍🏻

src/pages/proposal.tsx
cgero-eth: Yeah, now we always display the delegation gating menu if the user cannot vote (if he didn't vote yet and the proposal is active)

src/pages/proposal.tsx
cgero-eth: Agree, I created a tech debt ticket to remove such `useRef` on the application => [APP-2594](https://aragonassociation.atlassian.net/browse/APP-2594)

src/containers/delegationGatingMenu/delegationGatingMenu.tsx
```diff
@@ -70,33 +86,33 @@ export const DelegationGatingMenu: React.FC = () => {
     delegateData === null ? (address as string) : delegateData;
```
Fabricevladimir, line 86: No change needed, just a heads up: This is an interesting comment because that is technically not true on subgraph. If delegate is `null` that means delegation hasn't even been enabled for this address/account, which is different from the address delegating to itself. This is actually quite crucial when it comes to a known auto-delegate bug that we have in the GovernanceErc20 contract. Of course in this scenario you're also checking the balance so our `needsSelfDelegation` calculation is correct.

EDIT: Bug [#2](https://www.notion.so/aragonorg/GovernanceERC20-Issues-2ac3fdf7ec4a4e3e89c1af7c245183ff)

src/services/aragon-sdk/query-keys.ts
```diff
@@ -62,4 +64,8 @@ export const aragonSdkQueryKeys = {
     AragonSdkQueryItem.VOTING_SETTINGS,
     params,
   ],
+  getMember: (
+    baseParams: IAragonSdkBaseParams,
+    params: IFetchMemberParams
+  ): QueryKey => [AragonSdkQueryItem.GET_MEMBER, baseParams, params],
```
Fabricevladimir, line 70: Both `baseParams` and `params` includes the wallet address. Would be nice to merge/remove one. Not a big deal though.

Fabricevladimir (approved):
Thanks! Looking good

src/services/aragon-sdk/query-keys.ts
cgero-eth: Yeah I wanted to reuse the `IAragonSdkBaseParams` interface for this case :/ Maybe best would be just pass these as `params` and remove the `baseParams`

PR #1125, opened 2023-11-09
Feature: APP-2547 - Updated wallet-based "You can't create proposals" modal

Barukimang:
## Description

Updates wallet-based "You can't create proposals" modal. Required string updates will be done by Evan.

Task: [APP-2574](https://aragonassociation.atlassian.net/browse/APP-2574)

## Type of change

## Checklist:

[APP-2574]: https://aragonassociation.atlassian.net/browse/APP-2574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/gatingMenu/index.tsx
+++ b/src/containers/gatingMenu/index.tsx
@@ -38,7 +38,7 @@ export const GatingMenu: React.FC = () => {
 
   const {data: daoDetails} = useDaoDetailsQuery();
   const {plugins, ensDomain, address} = daoDetails ?? {};
-  const daoName =
+  const daoDisplayName =
     toDisplayEns(ensDomain) !== '' ? toDisplayEns(ensDomain) : address;
 
   const {data: daoToken} = useDaoToken(plugins?.[0].instanceAddress);
@@ -51,11 +51,13 @@ export const GatingMenu: React.FC = () => {
 
   const handleCloseMenu = () => close();
   const handleWrapTokens = () => {
-    const communityPath = generatePath(Community, {network, dao: daoName});
+    const communityPath = generatePath(Community, {network, dao: daoDisplayName});
     navigate(communityPath);
     close();
     handleOpenModal();
   };
+
+  const daoName = daoDetails?.metadata.name;
 
   const pluginType = plugins?.[0].id as PluginTypes;
   const isTokenBasedDao = pluginType === 'token-voting.plugin.dao.eth';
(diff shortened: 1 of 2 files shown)


Fabricevladimir:
Thanks @Barukimang! Waiting on the string updates to approve 👍🏾

Barukimang:
@Fabricevladimir ! Thanks for reviewing.

Just included a fix for the DAO name not properly showing, you can find more details of the issue [here](https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/40?selectedIssue=APP-2556) in this ticket for token based (not wallet-based) DAOs:

For the second issue, with the modal loop, found the issue is only there on Staging (Production is fine). Seems it has to do with recent changes which were made to `gatingMenu/index.tsx` which you can find [here](https://github.com/aragon/app/pull/1102/files)

Guess we could leave that for Sepehr?

Barukimang:
Just jammed with @sepehr2github and we fixed the looping issue as can be seen in the video [here] !(https://aragonassociation.atlassian.net/jira/software/c/projects/APP/boards/40?selectedIssue=APP-2556)

And updated the strings for the modal @Fabricevladimir

src/containers/gatingMenu/index.tsx
```diff
@@ -49,9 +50,20 @@ export const GatingMenu: React.FC = () => {
     pluginType: plugins?.[0].id as PluginTypes,
   });
 
-  const handleCloseMenu = () => close();
+  const handleCloseMenu = () => {
+    const governancePath = generatePath(Governance, {
+      network,
+      dao: daoDisplayName,
+    });
+    navigate(governancePath);
```
Fabricevladimir, line 58: Does it make sense to navigate to the governance page regardless of where you tried to create the proposal? Another option is to navigate back to the page where the proposal creation was triggered with `navigate(-1)`. What do you think?

Fabricevladimir (approved):
LGTM! Left a question.

src/containers/gatingMenu/index.tsx
Barukimang: It definitely makes sense :) @Fabricevladimir  Had tried it before and just now with this code:

`const handleCloseMenu = () => {
    navigate(-1);
    close();
  };`

Issue is that the modal then does not close anymore after the first click. Found that after clicking the state of `isOpen` renders:

`false
false
true`

Haven't found what sets it back to `true`. Think it might be `ProtectedRoute` but am not sure. Do you maybe have an idea of what it could be?

Fabricevladimir: You might get away with closing the modal first and wrapping the navigation in a `setTimeout`. Don't worry about it though if it's not working, it was a stray observation anyway.

PR #1126, opened 2023-11-09
fix: APP-2579 tweaked token withdraw amount display on mobile

babarella:
## Description

When opening a Token Withdrawal Proposal at Mobile, you cannot see the complete token amount (see image 1, token amount is 2,000 USD, see image 2). So the issue is that you don’t know exactly what you are voting on.

Task: [APP-2579](https://aragonassociation.atlassian.net/browse/APP-2579)

## Type of change

## Checklist:

[APP-2579]: https://aragonassociation.atlassian.net/browse/APP-2579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/executionWidget/actions/withdrawCard.tsx
+++ b/src/components/executionWidget/actions/withdrawCard.tsx
@@ -5,6 +5,7 @@ import styled from 'styled-components';
 
 import {AccordionMethod} from 'components/accordionMethod';
 import {ActionWithdraw} from 'utils/types';
+import {NumberFormat, formatterUtils} from '@aragon/ods';
 
 export const WithdrawCard: React.FC<{
   action: ActionWithdraw;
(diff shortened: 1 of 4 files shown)


src/components/executionWidget/actions/withdrawCard.tsx
```diff
@@ -34,10 +35,12 @@ export const WithdrawCard: React.FC<{
           tokenCount={action.amount}
```
Fabricevladimir: This might need to be formatted as well as it's causing everything else on the card to shrink.
image

Fabricevladimir (changes requested):
Thanks, @babarella. Unfortunately, ran into an issue on develop where the withdraw flow immediately crashes. Can you take a look at that as well?

src/components/executionWidget/actions/withdrawCard.tsx
```diff
@@ -5,13 +5,35 @@ import styled from 'styled-components';
...
   action: ActionWithdraw;
   daoName: string;
 }> = ({action, daoName}) => {
   const {t} = useTranslation();
 
+  const amount = Number(action.amount) || 0;
+  const tokenPrice = Number(action.tokenPrice) || 0;
+
+  const tokenCountDisplay =
+    amount > 99999
```
Fabricevladimir, line 20: Interesting little thing, on desktop, we would probably be fine all the way up to millions; on mobile it would be nice to see more numbers. My vote is that this card gets refactored, but that's a whole other story.

Fabricevladimir (approved):
Good work!

PR #1100, opened 2023-10-30
Connect SDK For Update flow

sepehr2github:
## Description

- OS Update has been implemented
- Plugin preparation flow has been connected
- Display available versions only

Notes:
1) Execution Widget could not be shown since decoders only return data not functions 
2) ~~Update flow is handling by Rakesh~~
3) prepared list doesn't work properly
4) update plugin proposal through error while execution

**Note: All tasks have been migrated to [APP-2623](https://aragonassociation.atlassian.net/browse/APP-2623)**

Task: [APP-2494](https://aragonassociation.atlassian.net/browse/APP-2494)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I ran all tests with success and extended them if necessary.
- [ ] I have updated the `CHANGELOG.md` file in the root folder.
- [x] I have tested my code on the test network.

[APP-2494]: https://aragonassociation.atlassian.net/browse/APP-2494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APP-2623]: https://aragonassociation.atlassian.net/browse/APP-2623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1130, opened 2023-11-14
Aragon Updates entrypoints 

RakeshUP:
## Description

Added isMember check and isUpdateAvailable check to the "Update Banners" on the settings page and the top navbar

## Type of change

## Checklist:

### DIFF
--- a/src/containers/navbar/updateBanner.tsx
+++ b/src/containers/navbar/updateBanner.tsx
@@ -14,6 +14,8 @@ import {featureFlags} from 'utils/featureFlags';
 import {useDaoDetailsQuery} from 'hooks/useDaoDetails';
 import {PluginTypes, usePluginClient} from 'hooks/usePluginClient';
 import {useWallet} from 'hooks/useWallet';
+import {DaoDetails} from '@aragon/sdk-client';
+import {useAvailableUpdateVersions} from 'hooks/useAvailableUpdateVersions';
 
 const UpdateBanner: React.FC = () => {
   const {t} = useTranslation();
@@ -27,6 +29,8 @@ const UpdateBanner: React.FC = () => {
   const pluginAddress = daoDetails?.plugins?.[0]?.instanceAddress as string;
   const pluginClient = usePluginClient(pluginType);
   const [isMember, setIsMember] = useState(false);
+  const {osxAvailableVersions, pluginAvailableVersions} =
+    useAvailableUpdateVersions(pluginType, daoDetails as DaoDetails);
 
   useEffect(() => {
     pluginClient?.methods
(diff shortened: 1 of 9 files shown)


src/hooks/useUpdateProposal.tsx
```diff
@@ -1,31 +1,26 @@
...
     queryKey: ['aragonVerifiedUpdateProposalCheck', proposalId],
-    queryFn: () =>
-      new Promise(resolve => {
-        setTimeout(() => resolve(Boolean(getRandomInt(2))));
-      }),
+    queryFn: () => {
+      return (
+        client?.methods.isDaoUpdate(proposalActions) ||
+        client?.methods.isPluginUpdate(proposalActions)
+      );
```
Fabricevladimir, line 22: This function isn't making any actual data fetching call. Why make it a `react-query` and not just a simple util function that takes in the client and proposal actions? Even the proposal id doesn't seem to be used either.

Also, I checked this with `dao.dao.eth` on Goerli, but unfortunately, both calls returned false. The proposal is a Dao update proposal.

src/hooks/useAvailableUpdateVersions.tsx
```diff
@@ -0,0 +1,100 @@
...
+import {useClient} from './useClient';
+import {compareVersions} from 'utils/library';
+import {OSX} from 'context/update';
+import {useProtocolVersions} from './useDaoVersions';
+
+export const useAvailableUpdateVersions = (
+  pluginType: PluginTypes,
+  daoDetails: DaoDetails
+) => {
+  const [pluginAvailableVersions, setPluginAvailableVersions] = useState(
```
Fabricevladimir, line 17: Just an observation, but I wonder why the map is being used considering we don't seem to be making use of it's features really. Regardless, I have refactored this a bit in the main branch. I think it's fine to merge it in, and I'll remove the file again.

src/pages/settings.tsx
```diff
@@ -54,7 +72,9 @@ export const Settings: React.FC = () => {
   }
 
   const daoUpdateEnabled =
```
Fabricevladimir, line 74: Apparently I screwed up way back, but should we only be showing the version card when an update is available?

src/containers/navbar/updateBanner.tsx
```diff
@@ -11,23 +11,48 @@ import {
...
+  const {osxAvailableVersions, pluginAvailableVersions} =
+    useAvailableUpdateVersions(pluginType, daoDetails as DaoDetails);
+
+  useEffect(() => {
+    pluginClient?.methods
+      .isMember({
+        address: address as string,
+        pluginAddress,
+      })
+      .then(value => setIsMember(value));
```
Fabricevladimir, line 41: See this I think would benefit from react query caching since it is making an asynchronous call.

Fabricevladimir (changes requested):
Thanks, Rakesh. Left a couple of comments. Additionaly, the update banner doesn't seem to get closed once we click on it

Fabricevladimir:
Closing because this is now being handled in #1100

PR #1136, opened 2023-11-16
F/gasless voting bugfixes

emmdim:
## Description

Bugfixing for gasless plugin and update to the lates vocdoni-sdk

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] I have selected the correct base branch.
- [x ] I have performed a self-review of my own code.
- [x ] I have tested my code on the test network.


src/components/proposalList/index.tsx
```diff
@@ -204,15 +201,6 @@ export function proposal2CardProps(
       ),
       title: proposal.vochain.metadata.title.default,
       description: proposal.vochain.metadata.questions[0].title.default,
```
Fabricevladimir, line 203: No longer allowing navigation to the created proposal?

src/context/proposalTransaction.tsx
```diff
@@ -375,8 +375,16 @@ const ProposalTransactionProvider: React.FC<Props> = ({children}) => {
           voteToPersist
         );
       }
+      invalidateProposalQueries();
```
Fabricevladimir: The reason I moved the query invalidation to the modal closing callback is that for whatever reason react-query wasn't picking it up after the invalidation here, which is really weird considering local storage interaction is synchronous. This is happening after the community vote right now

src/locales/en/common.json
```diff
@@ -1396,12 +1404,12 @@
...
       "sourcifyStatusPending": "Validating on Sourcify …",
-      "etherscanStatusPending": "Validating on Etherscan …",
+      "etherscanStatusPending": "Validating on {{blockExplorerName}} …",
       "sourcifyStatusSuccess": "Verified on Sourcify",
-      "etherscanStatusSuccess": "Verified on Etherscan",
+      "etherscanStatusSuccess": "Verified on {{blockExplorerName}}",
       "sourcifyStatusWarning": "Partially verified on Sourcify",
       "sourcifyStatusCritical": "Not verified on Sourcify",
-      "etherscanStatusCritical": "Not verified on Etherscan",
+      "etherscanStatusCritical": "Not verified on {{blockExplorerName}}",
```
Fabricevladimir: Please revert these three as the corresponding ticket has not yet been picked up on our board.

Fabricevladimir (commented):
Thanks, left a couple of comments and questions

src/components/proposalList/index.tsx
selankon: Before it was

```ts
            id: pluginAddress + '_' + proposal.id,
```

Now is just using the same as the previously calculated props

https://github.com/aragon/app/blob/98425858ee7d911c22eb4013f9000d39f1c7fd2d/src/components/proposalList/index.tsx#L168-L190

Fabricevladimir:
Also there seems to be a styling issue with this as the dropdown trigger is partly cropped
Screenshot 2023-11-16 091745

src/components/proposalList/index.tsx
Fabricevladimir: Ah, didn't catch that the props were being spread. 👍🏾

src/containers/votingTerminal/gaslessVotingTerminal.tsx
```diff
@@ -260,14 +265,21 @@ export const GaslessVotingTerminal: React.FC<CommitteeVotingTerminalProps> = ({
...
+
   return (
     <>
       <Container>
         <Header>
           <Title>{t('votingTerminal.vocdoni.title')}</Title>
           <Summary>{t('votingTerminal.vocdoni.desc')}</Summary>
         </Header>
-        <Accordion type={'multiple'}>
+        <AccordionMultiple defaultValue={defaultAccordion}>
```
Fabricevladimir, line 281: Something is up with the Accordion styling:
- The accordion title is replicated inside the accordion itself once it is opened.
- the corners shouldn't be rounded
- the dropdown menu trigger button should be removed since there isn't an actual dropdown in te header
image

src/context/proposalTransaction.tsx
selankon: Removed from there and added the `invalidateProposalQueries();` also inside the gasless voting modal

a970b111

src/locales/en/common.json
selankon: Is this what you need? 08bab2e4

selankon:
> Also there seems to be a styling issue with this as the dropdown trigger is partly cropped Screenshot 2023-11-16 091745

image

Much better now 0c77e134

src/containers/votingTerminal/gaslessVotingTerminal.tsx
selankon: To fix this i just created a custom accordion item with different styles to achieve what are you asking on the PR comment.

https://github.com/aragon/app/commit/1d29d2c3e13647767436da4617771c4ebba79e2f

The only problem is that the voting terminal border is still rounded, may I have to add a attribute for the voting terminal specifying that the borders haven't to be rounded?

image

Fabricevladimir (approved):

PR #1137, opened 2023-11-17
Feat: App-2545 Backend: Add token value API and integrate

jamesej:
## Description

This PR connects the App to a new Backend token value API which uses Covalent for any token data/pricing calls, falling back if this fails for some reason to Coingecko. It also caches data for an hour.

Task: [APP-2545](https://aragonassociation.atlassian.net/browse/APP-2545)

## Type of change

## Checklist:

[APP-2545]: https://aragonassociation.atlassian.net/browse/APP-2545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/token/token-service.ts
```diff
@@ -70,11 +71,65 @@ class TokenService {
       ? TOP_ETH_SYMBOL_ADDRESSES[symbol.toLowerCase()]
       : address;
 
-    const token =
-      processedNetwork === 'base' || processedNetwork === 'base-goerli'
-        ? this.fetchCovalentToken(processedNetwork, processedAddress)
-        : this.fetchCoingeckoToken(processedNetwork, processedAddress);
```
cgero-eth, line 76: I would remove these two methods as we're not using them anymore. Same for the `CovalentToken` and `CoingeckoToken` types

src/services/token/token-service.ts
```diff
@@ -70,11 +71,65 @@ class TokenService {
...
+    }
+  `;
 
+  private fetchTokenFromBackend = async (
+    network: SupportedNetworks,
+    tokenAddress: string
+  ) => {
+    let token: Token | null = null;
+    try {
+      const {token: resp} = await request(
```
cgero-eth, line 107: Can we add a type that represent the response returned by the backend?

cgero-eth (commented):
Thanks for the update @jamesej!

src/services/token/token-service.ts
jamesej: Don't think it's necessary as this is the only place it would ever be used and we convert the incoming data straight into our local type `Token`. Since it's Typescript it won't even validate the incoming data is in the right format. We'd effectively be repeating ourselves to no purpose, needing the code to be changed in two places rather than one if the incoming data structure changed.

src/services/token/token-service.ts
jamesej: That's done, thanks

cgero-eth (approved):

PR #1138, opened 2023-11-21
Fix: Flash of empty state even when data is available 

RakeshUP:
## Description

- Empty state is shown momentarily on the Finance and Community page even when there's some data
- The fix was to show a loading state till all the data from multiple subgraph/RPC queries were loaded

Task: [APP-2389](https://aragonassociation.atlassian.net/browse/APP-2389)

## Type of change

## Checklist:

[APP-2389]: https://aragonassociation.atlassian.net/browse/APP-2389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useDaoMembers.tsx
+++ b/src/hooks/useDaoMembers.tsx
@@ -143,7 +143,7 @@ export const useDaoMembers = (
   const {
     data: subgraphData = [],
     isError: isSubgraphError,
-    isInitialLoading: isSubgraphLoading,
+    isLoading: isSubgraphLoading,
   } = useMembers(
     {pluginAddress, pluginType},
     {enabled: useSubgraph && enabled}
--- a/src/hooks/useDaoMembers.tsx
+++ b/src/hooks/useDaoMembers.tsx
@@ -139,7 +139,9 @@ export const useDaoMembers = (
     network === 'base' ||
     network === 'base-goerli'
   );
-  const useSubgraph = isTokenBased === true || !covalentSupportedNetwork;
+
+  const useSubgraph =
+    (pluginType != null && !isTokenBased) || !covalentSupportedNetwork;
   const {
     data: subgraphData = [],
     isError: isSubgraphError,
@@ -168,7 +170,7 @@ export const useDaoMembers = (
   const {
     data: graphqlData,
     isError: isGraphqlError,
-    isInitialLoading: isGraphqlLoading,
+    isLoading: isGraphqlLoading,
   } = useTokenHolders(
     {
       network,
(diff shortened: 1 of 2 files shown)


src/hooks/useDaoMembers.tsx
```diff
@@ -143,7 +143,7 @@ export const useDaoMembers = (
   const {
     data: subgraphData = [],
     isError: isSubgraphError,
-    isInitialLoading: isSubgraphLoading,
```
RakeshUP, line 146: @RuggeroCino, is there a specific reason you added `isInitialLoading` specifically instead of the react-query's `isLoading` here?

cgero-eth: @RakeshUP The `isLoading`variable will be set to `true` when the query is not enabled, the `isInitialLoading` should instead be checking that the query is actually enabled ([react-query v4 migration](https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#disabled-queries))

cgero-eth (commented):
Given that the `isLoading` result returns true when a query is not enabled, this is currently causing an infinite loading on the dao members list when not using the subgraph query, i.e. on Polygon or Ethereum token-based dao. Here for instance:
http://localhost:3000/#/daos/polygon/0x72f4c30cb36b8fae1c82175a11586f298654d6eb/community

src/hooks/useDaoMembers.tsx
```diff
@@ -204,7 +206,7 @@ export const useDaoMembers = (
         daoToken,
         memberCount,
       },
-      isLoading: isSubgraphLoading || isGraphqlLoading,
+      isLoading: isSubgraphLoading && isGraphqlLoading,
```
cgero-eth, line 209: This is pretty counterintuitive, the `isLoading` should be set to true when one of the query is loading. There shouldn't be a case where both queries are loading. What's the issue of using the `isInitialLoading` and the `||` operator?

RakeshUP: When using `isInitialLoading` with the `||`, the `isLoading` tends to be `false` and this causes consumers of this hook to display a flash of an empty state or fallback state.

I'm thinking of this line ```isLoading: isSubgraphLoading && isGraphqlLoading``` similar to `Promise.race`. When not using isInitialLoading, the isSubgraphLoading and isGraphqlLoading are true from the beginning, and when one completes loading, the data is returned to the hook consumer.

cgero-eth: Okay I see, we might review this later on, thanks for the explanation! 🙏🏻

cgero-eth (approved):

src/hooks/useDaoMembers.tsx
RakeshUP: Yeah, I know it's not ideal, but there are quite a lot of intermittent states this hook can go into and that's causing this hiccup

PR #1143, opened 2023-11-24
Fix: Plugin update proposal execution flow

sepehr2github:
## Description

- Change update card view from horizontal to vertical
- Update SDK 

Task: [APP-2625](https://aragonassociation.atlassian.net/browse/APP-2625)

## Type of change

## Checklist:

[APP-2625]: https://aragonassociation.atlassian.net/browse/APP-2625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/proposals.ts
```diff
@@ -969,13 +968,13 @@ export function recalculateProposalStatus<
  * @param client - An instance of the `Client` class providing methods for DAO and plugin updates.
  * @returns A boolean indicating whether the proposal contains verified updates for Aragon DAO or plugins.
  */
-export function isVerifiedAragonUpdateProposal(
-  proposalActions: DaoAction[],
+export async function isVerifiedAragonUpdateProposal(
```
Fabricevladimir: Considering this is now an async function, I would highly recommend having a react-query hook in the services for it.

src/containers/defineProposal/update.tsx
```diff
@@ -48,6 +48,10 @@ export const DefineUpdateProposal: React.FC = () => {
   // data fetching
   const {data: dao, isLoading: detailsLoading} = useDaoDetailsQuery();
   const daoAddress = dao?.address as string;
+  const pluginUpdateTypeLabel =
+    dao?.plugins[0].id === 'token-voting.plugin.dao.eth'
+      ? 'Token voting'
+      : 'Multisig';
```
Fabricevladimir, line 54: Do we not have translation keys for this?

src/context/createProposal.tsx
```diff
@@ -363,17 +363,18 @@ const CreateProposalWrapper: React.FC<Props> = ({
...
+          const daoActionsArray =
+            client.encoding.applyUpdateAndPermissionsActionBlock(
+              daoDetails?.address as string,
+              {
+                permissions: action.inputs.permissions,
+                initData: new Uint8Array([]),
+                helpers: action.inputs.helpers,
+                versionTag: action.inputs.versionTag,
+                pluginRepo: pluginAvailableVersions?.address as string,
+                pluginAddress: pluginAddress,
```
Fabricevladimir: These two should be coming from the action inputs

src/context/createProposal.tsx
```diff
@@ -363,17 +363,18 @@ const CreateProposalWrapper: React.FC<Props> = ({
...
-              pluginRepo: pluginAvailableVersions?.address as string,
-              pluginAddress: pluginAddress,
-            }
-          );
+          const daoActionsArray =
+            client.encoding.applyUpdateAndPermissionsActionBlock(
+              daoDetails?.address as string,
+              {
+                permissions: action.inputs.permissions,
+                initData: new Uint8Array([]),
```
Fabricevladimir: `initData` as well.

src/context/createProposal.tsx
```diff
@@ -363,17 +363,18 @@ const CreateProposalWrapper: React.FC<Props> = ({
...
-            {
-              permissions: action.inputs.permissions,
-              initData: new Uint8Array([]),
-              helpers: action.inputs.helpers,
-              versionTag: action.inputs.versionTag,
-              pluginRepo: pluginAvailableVersions?.address as string,
-              pluginAddress: pluginAddress,
-            }
-          );
+          const daoActionsArray =
```
Fabricevladimir: This should be renamed to `pluginUpdateActions` or something since it's the plugin update and not the Dao update

src/hooks/useUpdateVerification.tsx
```diff
@@ -8,32 +7,21 @@ import {useClient} from './useClient';
  * @param address dao address
  * @returns an area of queries the indicates the status of verifications
  */
-export function useUpdateVerification(
-  actions: DaoAction[],
-  daoAddress: string
-  // isPluginUpdateProposal?: boolean,
-  // isOsUpdateProposal?: boolean
-) {
+export function useUpdateVerification(daoAddress: string, proposalId?: string) {
```
Fabricevladimir: Please update the comments since this is no longer a mock and now takes in a different parameter(s). 

I haven't wrapped the whole concept around my head just yet, but doesn't this force you to run both validations even if the propsal is only one type of update proposal?

Also, in general, I recommend using the services structure for these hooks as we're able to manage them easier and have fine grained control concerning when to enable/disable a hook.

src/hooks/useUpdateVerification.tsx
```diff
@@ -8,32 +7,21 @@ import {useClient} from './useClient';
...
   const verificationQueries = [
     {
       queryKey: ['isPluginUpdateProposalValid', daoAddress],
       queryFn: () =>
-        client?.methods.isPluginUpdateValid({
-          daoAddress: daoAddress,
-          actions,
-        }),
-      enabled: Boolean(daoAddress) && Boolean(actions),
+        client?.methods.isPluginUpdateProposal(proposalId as string),
```
Fabricevladimir: Shouldn't this be `isPluginUpdateProposalValid`?

src/hooks/useUpdateVerification.tsx
```diff
@@ -8,32 +7,21 @@ import {useClient} from './useClient';
...
+      enabled: Boolean(daoAddress) && Boolean(proposalId),
```
Fabricevladimir: The query seemingly no longer needs `daoAddress`, I'm not sure it should be part of what is enabling and disabling it.

src/containers/proposalSnapshot/index.tsx
```diff
@@ -41,16 +41,23 @@ type ProposalItemProps = CardProposalProps & {
...
+    }
 
-  let verifiedUpdateProposal = false;
-
-  if (client != null) {
-    verifiedUpdateProposal = isVerifiedAragonUpdateProposal(actions, client);
-  }
-
+    fetchIsVerifiedAragonUpdateProposal();
+  }, [client, proposalId]);
```
Fabricevladimir: That is potentially a whole lot of calls trying to find if the proposal is an update proposal. My proposal (hehehe) is that at the very least convert this to a react-query query so we can make use of caching.

src/components/proposalList/index.tsx
```diff
@@ -59,15 +59,26 @@ type ProposalListItemProps = CardProposalProps & {
...
+      }
+    }
 
-  let verifiedUpdateProposal = false;
-
-  if (client != null) {
-    verifiedUpdateProposal = isVerifiedAragonUpdateProposal(actions, client);
-  }
+    fetchIsVerifiedAragonUpdateProposal();
+  }, [client, proposalId]);
```
Fabricevladimir: See my comment on the proposal snapshot

Fabricevladimir (changes requested):

src/containers/defineProposal/update.tsx
sepehr2github: No, we don't unfortunately. Will check with Selim

Fabricevladimir:
Two other things:
Seems like the `isMember` function isn't working and returning false when that's not the case.
The banner on top of the Proposal Card doesn't show up for the update proposals

A final very nice to have would be after proposal execution, if we know that it is an update proposal, invalidate the queries that fetch the plugin and os versions because right now once we've executed the proposal, the version card on the settings page doesn't get updated until we manually refresh the page or wait the allotted 5 mins for the query to become stale

src/locales/en/common.json
```diff
@@ -1404,12 +1404,12 @@
...
       "sourcifyStatusPending": "Validating on Sourcify …",
-      "etherscanStatusPending": "Validating on Etherscan …",
+      "etherscanStatusPending": "Validating on {{blockExplorerName}} …",
       "sourcifyStatusSuccess": "Verified on Sourcify",
-      "etherscanStatusSuccess": "Verified on Etherscan",
+      "etherscanStatusSuccess": "Verified on {{blockExplorerName}}",
       "sourcifyStatusWarning": "Partially verified on Sourcify",
       "sourcifyStatusCritical": "Not verified on Sourcify",
-      "etherscanStatusCritical": "Not verified on Etherscan",
+      "etherscanStatusCritical": "Not verified on {{blockExplorerName}}",
```
Fabricevladimir, line 1412: This ticket hasn't been picked up yet so there is no variable being passed for these strings.

Also, for whatever reason the title for the update Proposal has `p` tags surrounding it

src/pages/proposal.tsx
```diff
@@ -91,6 +90,8 @@ import {
 import {Action} from 'utils/types';
 import {GaslessVotingTerminal} from '../containers/votingTerminal/gaslessVotingTerminal';
 import {useGaslessHasAlreadyVote} from '../context/useGaslessVoting';
+import {UpdateVerificationCard} from 'containers/updateVerificationCard';
+import {featureFlags} from 'utils/featureFlags';
```
Fabricevladimir: We have a duplicate (thoughh commented out) import on line 59

src/pages/proposal.tsx
```diff
@@ -782,16 +783,9 @@ export const Proposal: React.FC = () => {
           )}
 
           {/* @todo: Add isUpdateProposal check once it's developed */}
```
Fabricevladimir: Can you remove the comment, please

src/context/proposalTransaction.tsx
```diff
@@ -247,6 +247,12 @@ const ProposalTransactionProvider: React.FC<Props> = ({children}) => {
     queryClient.invalidateQueries(currentProposal);
   }, [pluginType, proposalId, queryClient]);
 
+  // if network is unsupported this will be caught when compared to client
+  const queryNetwork = useMemo(
```
Fabricevladimir: Why have this outside and not calculated inside the funciton that needs it?

src/containers/updateVerificationCard/index.tsx
```diff
@@ -1,145 +1,129 @@
...
+  const pluginType =
+    daoDetails?.plugins[0].id === 'token-voting.plugin.dao.eth'
+      ? 'token voting'
+      : 'multisig';
+
+  const [
+    {data: pluginUpdateVerification, isLoading: isPluginUpdateLoading},
+    {data: osUpdateVerification, isLoading: isOSUpdateLoading},
+  ] = useUpdateVerification(proposalId as string);
+
```
Fabricevladimir: Not a huge deal, but ths is going to run for every proposal regardless of whether it's a os-update or default proposal because we don't have access to the options object to enable this only after the `useIsUpdateProposal` hook returns true.

src/containers/updateVerificationCard/Status.tsx
```diff
@@ -30,22 +41,60 @@ const Icon: React.FC<{mode: StatusProps['mode']}> = ({mode}) => {
   }
 };
 
-export const Status: React.FC<StatusProps> = ({mode, label}) => {
+export const Status: React.FC<StatusProps> = ({
```
Fabricevladimir: Can you lowercase the file name?

src/pages/proposal.tsx
```diff
@@ -782,16 +783,9 @@ export const Proposal: React.FC = () => {
...
-          {/* {proposal &&
+          {proposal &&
```
Fabricevladimir, line 784: Don't forget to hide the verification card when the proposal is neither pending nor active.

src/locales/en/common.json
sepehr2github: could you point me to the update Proposal title file?

Fabricevladimir (approved):

PR #1182, opened 2023-12-21
DRAFT: Quick fix for L2 ENS issue

cgero-eth:


cgero-eth:
Closing as not needed anymore, proper fix will be done in APP-2661

PR #1196, opened 2024-01-16
Feature: APP-2701 - DAO list advanced filters

Fabricevladimir:
## Description

- Implements the DaoFilterModal

TODO: Update translation keys for some networks

Task: [APP-2701](https://aragonassociation.atlassian.net/browse/APP-2701)

## Type of change

## Checklist:

[APP-2701]: https://aragonassociation.atlassian.net/browse/APP-2701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/containers/daoExplorer/daoExplorer.tsx
```diff
@@ -31,87 +30,124 @@ const followedDaoToDao = (dao: NavigationDao): IDao => ({
...
-          iconRight={
-            isFetchingNextPage ? <Spinner size="xs" /> : <IconChevronDown />
-          }
-          bgWhite={true}
-          mode="ghost"
-          onClick={() => fetchNextPage()}
-        />
-      )}
-    </Container>
+    <>
```
cgero-eth: The `<>` fragment can be removed, right?

src/containers/daoExplorer/daoExplorer.tsx
```diff
@@ -31,87 +30,124 @@ const followedDaoToDao = (dao: NavigationDao): IDao => ({
...
+        isLoading={isLoading}
+        quickFilter={filters.quickFilter}
+        governanceIds={filters.governanceIds}
+        onClearFilters={resetFilters}
+        onQuickFilterChanged={setQuickFilter}
+        onNetworkFiltersChanged={setNetworkFilters}
+        onGovernanceFiltersChanged={setGovernanceIdFilters}
+        onClose={() => {
+          setShowAdvancedFilters(false);
+        }}
```
cgero-eth: Just a note, ideally we should have less properties passed between components. This could be achieved here by passing only the following props:
```
isOpen={isModalOpen}
filters={filters}
onChange={handleFilterChange}
onClose={handleModalClose}
```
Given that react-query caches the query results, we could safely use the `useDaos` hook both here and on the `DaoFilterModal` component

src/containers/daoFilterModal/data.ts
```diff
@@ -0,0 +1,78 @@
...
+type NetworkFilter = {
+  label: string;
+  value: SupportedNetworks;
+  testnet?: boolean;
+};
+
+export const networkFilters: Array<NetworkFilter> = [
+  {
+    label: 'explore.modal.filterDAOs.label.ethereum',
+    value: 'ethereum',
```
cgero-eth: Can't we build this array based on the `CHAIN_METADATA` definitions? In that way we don't have to maintain multiple lists when adding/removing chains

src/containers/daoFilterModal/index.tsx
```diff
@@ -0,0 +1,338 @@
...
+  isOpen: boolean;
+  isLoading: boolean;
+  onClose: () => void;
+  onClearFilters: () => void;
+  onQuickFilterChanged: (value: QuickFilterValue) => void;
+  onNetworkFiltersChanged: (value: SupportedNetworks[] | undefined) => void;
+  onGovernanceFiltersChanged: (value: string[] | undefined) => void;
+};
+
+const DaoFilterModal: React.FC<DaoFilterModalProps> = ({
```
cgero-eth, line 42: Deselecting a governance filter does not work on first click when there's only one governance filter selected

src/containers/daoExplorer/daoExplorer.tsx
Fabricevladimir: The Modal is on the same level as the Container

src/containers/daoExplorer/daoExplorer.tsx
Fabricevladimir: Concerning the props, yes, but this also going to mean that the individual components are going to have to be aware of the way they're being controlled since they have to fire specific actions.

src/containers/daoExplorer/daoExplorer.tsx
cgero-eth: Oh I see, why?

src/containers/daoExplorer/daoExplorer.tsx
cgero-eth: Which is okay imho, actually the logic of building the filter object fits more the dao-filter component than the dao-explorer one

src/containers/daoExplorer/daoExplorer.tsx
Fabricevladimir: No real reason now actually lol 😆

src/containers/daoExplorer/daoExplorer.tsx
Fabricevladimir: Yes, and no. The Dao explorer itself is going to have to make use of the same logic anyway because of the quick filters.

About the `useDaos`: valid point, but for now there will be a whole lot of duplicated code because you also have to take into account `useFollowedDaos` and the OR-ing of the results. Not sure it's worth it for two props that are needed in both parent and child components.

src/containers/daoFilterModal/data.ts
Fabricevladimir: True, good catch!

src/containers/daoExplorer/daoExplorer.tsx
```diff
@@ -33,59 +28,70 @@ const followedDaoToDao = (dao: NavigationDao): IDao => ({
...
+  const isLoading = showFollowedDaos ? followedDaosLoading : daoListLoading;
 
-  const displayFilters = isConnected && followedDaoList.length > 0;
-
-  useEffect(() => {
-    if (followedStatus === 'success' && followedDaoList.length === 0) {
-      setFilter('all');
-    }
-  }, [followedStatus, followedDaoList]);
+  console.log(filteredDaoList);
```
cgero-eth:
```suggestion
```

cgero-eth (approved):
I applied the fix on my base branch, apart from the conflicts the PR good to me, thanks @Fabricevladimir!

PR #1240, opened 2024-02-22
feat: Implements fetching of Vocdoni DAO member information through census3

emmdim:
## Description

Implements fetching of DAO member information through census3.

## Type of change

## Checklist:


src/services/vocdoni-census3/census3-queries.ts
```diff
@@ -0,0 +1,119 @@
...
+    },
+    ...options,
+  }));
+
+  return useQueries({queries});
+};
+
+interface useCensus3VotingPowerProps {
+  page?: number;
+  // censusId?: string | null;
```
Fabricevladimir: Please remove any commented out code

src/services/vocdoni-census3/census3-queries.ts
```diff
@@ -0,0 +1,119 @@
+import {useQueries, useQuery, UseQueryOptions} from '@tanstack/react-query';
```
Fabricevladimir: Can you structure the vocdoni-census3 to match our current structure, with a dedicated api interface file, queries folder and one query per file? This way things are consistent. We also have a few hooks we're in the process of refactoring to match the new structure

src/services/vocdoni-census3/census3-queries.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Get member balance from vocdoni census3. It accepts a census id or a token id to retrieve the voting power
+ * @param holders list of members to get the balance
+ * @param censusId
+ * @param tokenId
+ */
+export const useCensus3VotingPower = (
+  holders: TokenDaoMember[],
+  censusId?: string | null,
+  tokenId?: string,
```
Fabricevladimir: Can you define params in an interface stored in the dedicated api file for vocdoni-census3 queries? Same for the other two queries.

```suggestion
IFetchCensus3VotingPowerParams
```

src/services/aragon-sdk/queries/use-past-voting-power.ts
```diff
@@ -23,10 +28,11 @@ const useGaslessPastVotingPower = () => {
   const getGaslessPastVotingPower = useCallback(
     async (address: string) => {
       if (!censusId) return BigNumber.from(0);
-      const votingPower = await getCensus3VotingPower(
+      const votingPower = await getCensus3VotingPowerByCensusId(
+        // todo(kon): use query
```
Fabricevladimir: Yes, lol

src/hooks/useCensus3DaoMembers.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+export const useCensus3DaoMembers = ({
+  holders,
+  pluginAddress,
+  pluginType,
+  options,
+}: UseCensus3DaoMembersProps): HookData<DaoMembersData> => {
+  const enable = options?.enabled || false;
+  const countOnly = options?.countOnly || false;
+  const {id: proposalId} = useParams();
```
Fabricevladimir, line 34: Doesn't this mean that this hook can only be used in what is currently the Proposal details page? Why not have the proposal id passed in as one of the params?

src/hooks/useCensus3DaoMembers.tsx
```diff
@@ -0,0 +1,156 @@
...
+}
+
+export const useCensus3DaoMembers = ({
+  holders,
+  pluginAddress,
+  pluginType,
+  options,
+}: UseCensus3DaoMembersProps): HookData<DaoMembersData> => {
+  const enable = options?.enabled || false;
+  const countOnly = options?.countOnly || false;
```
Fabricevladimir:
```suggestion
  const enable = options?.enabled ?? false;
  const countOnly = options?.countOnly ?? false;
```

src/hooks/useCensus3DaoMembers.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const {id: proposalId} = useParams();
+  const {data: daoToken} = useDaoToken(pluginAddress);
+  const {address} = useWallet();
+
+  // If is not a wrapped token and not on a proposal context we can still get the token holders amount
+  const enableCensus3Token = enable && !proposalId;
+  const {data: census3Token} = useCensus3Token(daoToken?.address ?? '', {
+    enabled: enable && !!(daoToken?.address ?? false) && enableCensus3Token,
```
Fabricevladimir:
```suggestion
    enabled: enable && !!daoToken?.address && enableCensus3Token,
```

src/hooks/useDaoMembers.tsx
```diff
@@ -40,7 +39,7 @@ const compareAddresses = (addressA?: string | null, addressB?: string | null) =>
 export const isTokenDaoMember = (member: DaoMember): member is TokenDaoMember =>
   'balance' in member;
 
-const sortDaoMembers =
+export const sortDaoMembers =
```
Fabricevladimir, line 47: Can you add documentation to this function since it's now being exported?

src/hooks/useCensus3DaoMembers.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const filteredMembers = !searchTerm
+    ? sortedData
+    : sortedData?.filter(
+        member =>
+          member?.address.toLowerCase().includes(searchTerm.toLowerCase())
+      );
+
+  let memberCount = 0;
+  if (enableCensus3Token && census3Token) {
+    memberCount = census3Token!.size;
```
Fabricevladimir:
```suggestion
    memberCount = census3Token.size;
```

Fabricevladimir (commented):
Thanks! Left a few comments. Unfortunately, I was unable to see the members on a DAO with exisiting vanilla erc20 token, and the community section was missing on the Settings page as well. I assume it's probably the same query failing in both places. Here's the DAO for reference: https://ipfs.eth.aragon.network/ipfs/QmXQnFypiXWJTG6dHjtpoeSWJvz8BRHz6TvawpTfXf6NQZ/#/daos/goerli/gaslessnowrap.dao.eth/dashboard

src/services/vocdoni-census3/census3-queries.ts
selankon: b3e453f2

src/services/vocdoni-census3/census3-queries.ts
selankon: bc54f345

src/services/aragon-sdk/queries/use-past-voting-power.ts
selankon: ea75f092 should be enough

src/services/vocdoni-census3/census3-queries.ts
selankon: 72f34322

src/hooks/useCensus3DaoMembers.tsx
selankon: This hook can be used either inside a proposal or either outside. If its inside (proposalId is not undefined), it uses the census3 Id to get the voting power of the members. 

If no proposal Id, uses the token address to get the voting powers or the members.

src/hooks/useCensus3DaoMembers.tsx
selankon: 8ca032bc

src/hooks/useDaoMembers.tsx
selankon: 718d22de

src/services/aragon-sdk/queries/use-past-voting-power.ts
```diff
@@ -4,50 +4,41 @@ import {useCallback} from 'react';
...
+    enable: pluginType === GaslessPluginName,
+  });
 
   return useQuery(
     aragonSdkQueryKeys.pastVotingPower(params),
     () =>
       censusId !== null
-        ? getGaslessPastVotingPower(params.address)
+        ? BigNumber.from(
+            fetchVotingPowerByCensusId({
```
Fabricevladimir, line 36: Does this call not need to be awaited?

Fabricevladimir (commented):
Thanks, left a single question

src/services/aragon-sdk/queries/use-past-voting-power.ts
emmdim: @Fabricevladimir should it be awaited if defined inside the useQuery function?

Fabricevladimir: I believe so, yes. Right now, we're passing a promise to the BigNumber constructor. Regardless it's actually quite helpful for debugging purposes to return a variable assigned with the awaited call value like so: 
```ts
  return useQuery(..., async () => {
     if(censusId) {
        const vp = await fetchVotingPowerByCensusId({...})
        return BigNumber.From(vp);
     }
  ...
  });
```

selankon: You are true, the call was needed to be awaited! Good catch! I also modified the code as you suggested: 676d7866

Fabricevladimir (approved):
Thanks! I took a look with this DAO: https://ipfs.eth.aragon.network/ipfs/QmbqQZBtX44aXTobtWFaiXyuWx5D9gaeJvLsoUAYkmJ9qr/#/daos/sepolia/0x4ad5882ce4b917b00e8e81b112f3530fe638778a/dashboard

I noticed a couple of things. The member list no longer hangs :white_check_mark:, but the census3 API is getting battered because there seems to be an infinite loop in the call to get the holders. I assume this is intended because we're scanning the transactions on the fly?

This is the recurring call: https://census3-stg.vocdoni.net/api/strategies/85/holders?pageSize=-1
The token has few holders but a lot of transactions apparently 🙈 

Finally, I noticed that the membership reloads every time you access the page. It's seems as if a new query is being made each time and the react-query cached result is not being used.

My recommendation is to merge this as is (assuming the interval call _is_ intentional) and push up a fix for the members not being cached in a follow up PR.

PR #1259, opened 2024-03-01
feat: APP-2930 - Sentry setup

cristianizzo:
## Description

- Implemented sentry -  error tracking
- Implemented pino - logger 
- Init Sentry only when user accept the terms
- Remove Elastic APM + configuration
- update readme

Task: [APP-2930](https://aragonassociation.atlassian.net/browse/APP-2930)

## Type of change

## Checklist:

[APP-2930]: https://aragonassociation.atlassian.net/browse/APP-2930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/webapp-deploy-dev.yml
```diff
@@ -1,288 +0,0 @@
-name: WebApp Deploy DEV (no Fleek)
```
cgero-eth, line 1: Is the Devops team aligned with the removal of this workflow & and `app-dev` environment?

src/hooks/useMonitoring.tsx
```diff
@@ -1,12 +0,0 @@
-import {usePrivacyContext} from 'context/privacyContext';
-import {useEffect} from 'react';
-import {monitoring} from 'services/monitoring';
-
-export const useMonitoring = () => {
-  const {preferences} = usePrivacyContext();
-
-  useEffect(
-    () => monitoring.enableMonitoring(preferences?.analytics),
```
cgero-eth, line 9: Currently the `useMonitoring` hook enables the monitoring service when the user accepts the cookies. I think we should do something similar here, to only initialise `Sentry` if the user accepts the cookies & analytics

cgero-eth (commented):
Thanks for the PR @cristianizzo!

.github/workflows/webapp-deploy-dev.yml
chilcano: The [webapp-deploy-dev](https://github.com/aragon/app/blob/develop/.github/workflows/webapp-deploy-dev.yml) workflow takes the changes from `develop` branch and publish a docker image. Once done, ArgoCD takes that docker image and deploys in Kubernetes instead of deploy it on Fleek. This is the url: https://app-dev.aragon.org
The same github workflow generate other docker image taking changes in branches `DOPS-xyz`, this is deployed on K8s using this URL: https://app-test.aragon.org/

Both URLs (https://app-dev.aragon.org/ and https://app-test.aragon.org/) are 2 new environments created some months ago only for testing purposes, where QA can run some tests without Fleek dependencies and where other developers can introduce changes in a isolated scenario.

src/hooks/useMonitoring.tsx
chilcano: Sentry, Google Analytic, Elastic APM, Plausible or any monitoring tool embedded in SPA (application running in the browser) will be blocked by the browser and/or Ad Blockers, that happens although cookies are accepted.
We have 2 options to avoid that:
1. Enable and activate CORS - sometimes is not effective
2. Deploy a reverse-proxy - it is effective

cgero-eth: Yes, ad-blockers block such events and we can investigate on making sure to send them.
In any case, we should not initialise Sentry and start sending events if users do no accept our cookie banner

src/services/monitoring.ts
```diff
@@ -1,30 +1,15 @@
-import {init as initApm, ApmBase, AgentConfigOptions} from '@elastic/apm-rum';
+import {initSentry} from 'services/sentry';
```
cgero-eth: Why not combining the two services `monitoring` and `sentry` into one? We could just have the monitoring service which initialises Sentry

src/services/logger.ts
```diff
@@ -0,0 +1,36 @@
...
+  warn = 'warn',
+  error = 'error',
+}
+
+const pinoLogger = pino({
+  level: (import.meta.env.VITE_APP_LOG_LEVEL as LogLevel) || LogLevel.warn,
+  browser: {
+    asObject: true,
+  },
+} as LoggerOptions);
```
cgero-eth: I would suggest to remove the logger functionality for this PR. I'm not sure if we should introduce a new dependency only for better logs on the console. Let's keep this out of scope for now and discuss if we want to introduce it or not

src/hooks/useMonitoring.tsx
```diff
@@ -1,12 +1,19 @@
...
-    [preferences?.analytics]
-  );
+  useEffect(() => {
+    const enableMonitoring = async () => {
+      if (preferences?.analytics !== undefined) {
+        await monitoring.enableMonitoring(preferences.analytics);
+      }
+    };
+    enableMonitoring().catch((error: Error) => {
```
cgero-eth: The monitoring service already checks the `enabled` property and enables the service only if set to true
```suggestion
    monitoring.enableMonitoring().catch((error: Error) => {
```

src/index.tsx
```diff
@@ -1,3 +1,4 @@
+import 'services/logger';
```
cgero-eth: This import is not needed, right?
```suggestion
```

src/services/sentry.ts
```diff
@@ -0,0 +1,59 @@
+// import * as Sentry from '@sentry/react';
```
cgero-eth: Commented-out code should not be pushed, let's remove this and only add these functionality when needed

cristianizzo: ah yep i forgot to remove, i did some test so sentry only start when terms are accepted

cgero-eth (approved):
LGTM, thanks @cristianizzo!

PR #1263, opened 2024-03-04
feat: APP-2933 - Check token allowance on wrap tokens flow

cgero-eth:
## Description

- Check token allowance during the wrap tokens flow

Note:
For such direct calls to smart contracts we should be using the [useReadContract](https://wagmi.sh/react/api/hooks/useReadContract) hook from Wagmi that is not available on the current installed version. I created a ticket ([APP-2949](https://aragonassociation.atlassian.net/browse/APP-2949)) to update Wagmi and Viem to v2.

Task: [APP-2933](https://aragonassociation.atlassian.net/browse/APP-2933)

## Type of change

## Checklist:

[APP-2933]: https://aragonassociation.atlassian.net/browse/APP-2933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APP-2949]: https://aragonassociation.atlassian.net/browse/APP-2949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/context/govTokensWrapping.tsx
+++ b/src/context/govTokensWrapping.tsx
@@ -60,7 +60,7 @@ const GovTokensWrappingProvider: FC<{children: ReactNode}> = ({children}) => {
   const {data: daoTokenData, isLoading: isTokenDataLoading} = useDaoToken(
     daoDetails?.plugins?.[0]?.instanceAddress || ''
   );
-  const underlyingToken = (daoTokenData as Erc20WrapperTokenDetails)
+  const underlyingToken = (daoTokenData as Erc20WrapperTokenDetails | undefined)
     ?.underlyingToken;
   const [daoTokenBalance, setDaoTokenBalance] = useState('');
 
@@ -80,7 +80,7 @@ const GovTokensWrappingProvider: FC<{children: ReactNode}> = ({children}) => {
 
   const {data: tokenAllowance} = useTokenAllowance(
     {
-      token: underlyingToken?.address,
+      token: underlyingToken?.address as string,
       owner: userAddress as string,
       spender: wrappedDaoToken?.address as string,
     },
--- a/src/context/govTokensWrapping.tsx
+++ b/src/context/govTokensWrapping.tsx
@@ -26,10 +26,11 @@ import React, {
 } from 'react';
 import {useForm, useWatch} from 'react-hook-form';
 import {generatePath, useLocation, useNavigate} from 'react-router-dom';
+import {useTokenAllowance} from 'services/aragon-sdk/queries/use-token-allowance';
 import {CHAIN_METADATA} from 'utils/constants';
 import {toDisplayEns} from 'utils/library';
 import {Community} from 'utils/paths';
-import {fetchBalance, getAllowance} from 'utils/tokens';
+import {fetchBalance} from 'utils/tokens';
 import {TokensWrappingFormData} from 'utils/types';
 import {useQueryClient} from 'wagmi';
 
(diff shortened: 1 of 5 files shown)


src/context/govTokensWrapping.tsx
```diff
@@ -77,6 +78,20 @@ const GovTokensWrappingProvider: FC<{children: ReactNode}> = ({children}) => {
...
+    {
+      token: underlyingToken?.address as string,
+      owner: userAddress as string,
+      spender: wrappedDaoToken?.address as string,
+    },
+    {
+      enabled:
+        underlyingToken != null &&
+        userAddress != null &&
+        wrappedDaoToken != null,
```
Fabricevladimir, line 91: Considering these three correspond with the required params, I would suggest handling the corresponding enabling/disabling inside the hook itself.

src/context/govTokensWrapping.tsx
```diff
@@ -357,6 +372,22 @@ const GovTokensWrappingProvider: FC<{children: ReactNode}> = ({children}) => {
     form.resetField('amount');
   }, [form, mode]);
 
+  useEffect(() => {
+    if (wrappedDaoToken == null || amount === '' || tokenAllowance == null) {
+      return;
+    }
+
+    const wrapAmount = BigInt(
+      ethers.utils.parseUnits(amount, wrappedDaoToken.decimals).toString()
```
Fabricevladimir, line 381: All fine, but if possible, in the spirit of reducing tech debt, let's use the viem utility function instead

Fabricevladimir (commented):
Thanks, Ruggero

src/context/govTokensWrapping.tsx
cgero-eth: Mm.. and create two interfaces, one with optional params for the hook and one with required params for the `getAllowance` call? I'd avoid that, or you have something else in mind?

Fabricevladimir: I mean to have the disabled check inside the the hook like we usually do. Something to the extent of this:
```javascript
  if (underlyingToken == null || userAddress == null || wrappedDaoToken == null) {
    options.enabled = false;
  }
```
This way we don't have to duplicate the check for whether the params are available everywhere we use that hook.

cgero-eth: Yeah I saw we're using this in some hooks and I don't like it that much tbh. For instance, to do it here we need to:

1. Add default empty object value to options parameter
2. Update the `IFetchTokenAllowanceParams` interface to mark all params as optional
3. Update the call to `getAllowance` to add the non-null assertion as `getAllowance(params.token!, params.owner!, params.spender!, provider).

It might make sense to add the non-null assertions on the react-query hook and avoid the casting on the hook usage (e.g. `userAddress as string`) but how to make this consistent across the app? Should we always set the parameters of a react-query call (e.g. `IFetchTokenAllowanceParams`) to optional and always handle the undefined check in the react-query hooks? How can we then differentiate between optional and not-optional params?

Fabricevladimir: Looks like we've a topic for a tech talk. I don't like having the params as optional since you lose context as to what is _actually_ optional, but having the assertions inside the query hook is more preferable to me because all that logic is handled there in only one place. It's much more convenient for using the hook. As for the optionality of the params themselves, we can keep things not-optional & optional and just cast the params when passing them into the hook. The hook defines its parameters and the user conforms to that either by casting or setting a default value.

Fabricevladimir (approved):

src/context/govTokensWrapping.tsx
cgero-eth: If we don't make the parameters optional, the assertions inside the query hook will look useless and might be removed by devs as redundant checks, because we're basically checking that a parameter that has a required type is not undefined 😅

Fabricevladimir: Also true lol. This is why I was defining optional params as `type|undefined`. Type wise completely useless, but that's how I was differentiating. At this point I guess we really just need to align and stay consistent

cgero-eth: Sounds good, let's discuss this topic during the next tech talk 👌🏻

PR #1267, opened 2024-03-05
fix: Ensure useDelegatee is disabled for non-wrapped tokens

selankon:
## Description

Now there is a moment where the `useDelegattee` hook is enabled when the token is not wrapped. This happen when the DAO details are not loaded yet, and the `useGaslessGovernanceEnabled` hook have is `true` by default. 

This PR changes the default value from `true` to `false` avoiding to enter into the `useDelegatee` query if is gasless plugin and the settings are not fully loaded yet. 

It also refactors the `useGaslessGovernanceEnabled` to use the react query cache to access to the Dao information, avoiding to pass down the plugin address, so now is easier to  be called: 

https://github.com/aragon/app/pull/1267/files#diff-81eebd27e361315ca8a90bd3f581c0b162f34580b468fd601430044c906dede5L40-R38

```diff
-  const {data: daoDetails} = useDaoDetailsQuery();
-  const {isGovernanceEnabled} = useGaslessGovernanceEnabled({
-    pluginAddress: daoDetails?.plugins[0].instanceAddress as string,
-    pluginType: daoDetails?.plugins[0].id as PluginTypes,
-  });
+  const {isGovernanceEnabled} = useGaslessGovernanceEnabled();
```

## Type of change

## Checklist:

### DIFF
--- a/src/hooks/useGaslessGovernanceEnabled.tsx
+++ b/src/hooks/useGaslessGovernanceEnabled.tsx
@@ -1,26 +1,27 @@
 import {GaslessPluginName, PluginTypes} from './usePluginClient';
 import {useVotingSettings} from '../services/aragon-sdk/queries/use-voting-settings';
 import {GaslessPluginVotingSettings} from '@vocdoni/gasless-voting';
+import {useDaoDetailsQuery} from './useDaoDetails';
 
-export const useGaslessGovernanceEnabled = ({
-  pluginType,
-  pluginAddress,
-}: {
-  pluginType?: PluginTypes;
-  pluginAddress?: string;
-}) => {
+export const useGaslessGovernanceEnabled = () => {
+  const {data: daoDetails} = useDaoDetailsQuery();
+
+  const pluginAddress = daoDetails?.plugins[0].instanceAddress;
+  const pluginType = daoDetails?.plugins[0].id as PluginTypes;
+
   const {data: votingSettings} = useVotingSettings({
-    pluginAddress: pluginAddress,
-    pluginType: pluginType,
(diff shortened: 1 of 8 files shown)


src/hooks/useGaslessGovernanceEnabled.tsx
```diff
@@ -1,26 +1,27 @@
...
   const isGasless = pluginType === GaslessPluginName;
+
   let isGovernanceEnabled = true;
 
   if (isGasless) {
     isGovernanceEnabled =
       (votingSettings as GaslessPluginVotingSettings)?.hasGovernanceEnabled ??
-      true;
+      false; // Await until loading is finished to ensure that governance is enabled
   }
```
Fabricevladimir:
```suggestion
   const isGovernanceEnabled =
    pluginType === GaslessPluginName
      ? !!(votingSettings as GaslessPluginVotingSettings)?.hasGovernanceEnabled
      : true;
```

Fabricevladimir (commented):
Thanks for the refactor! Left a quick suggestion. Let me know what you think

src/hooks/useGaslessGovernanceEnabled.tsx
selankon: For me its ok and the code is enough clear imo. 

I try to use ternary operators only when is absolutely needed, but in this case, the code wins on readability. 

Thxs, committing the changes!

selankon: 89621675

Fabricevladimir (approved):

PR #1301, opened 2024-03-27
chore: APP-2953 - Migrate to react-query v5

thekidnamedkd:
## Description

Begin migration of app to react-query v5

Task: [APP-2953](https://aragonassociation.atlassian.net/browse/APP-2953)

## Type of change

## Checklist:

[APP-2953]: https://aragonassociation.atlassian.net/browse/APP-2953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/containers/daoExplorer/daoExplorer.tsx
```diff
@@ -99,21 +99,20 @@ export const DaoExplorer = () => {
...
     () =>
-      followedDaosResult.data?.pages.flatMap(page =>
-        page.data.map(followedDaoToDao)
-      ) ?? [],
-    [followedDaosResult.data]
+      followedDaosResult.data?.pages
+        .flatMap(page => page.data)
+        .map(followedDaoToDao),
+    [followedDaosResult]
   );
```
thekidnamedkd, line 106: this has a type issue now where infiniteQueries data comes back with pages.

src/services/aragon-sdk/mutations/use-delegate-tokens.ts
```diff
@@ -15,16 +15,20 @@ const delegateTokens = async (
...
-    AsyncGenerator<DelegateTokensStepValue>,
-    unknown,
-    IDelegateTokensParams
+  options?: Omit<
+    UseMutationOptions<
+      AsyncGenerator<DelegateTokensStepValue>,
+      unknown,
+      IDelegateTokensParams
+    >,
+    'mutationKey'
```
cgero-eth: Do we need to omit the `mutationKey` parameter? Looks like it's optional anyway

src/services/aragon-backend/queries/use-token-holders.ts
```diff
@@ -45,12 +45,12 @@ const fetchTokenHolders = async (
 
 export const useTokenHolders = (
   params: IFetchTokenHoldersParams,
-  options: UseQueryOptions<TokenHoldersResponse> = {}
+  options: Omit<UseQueryOptions<TokenHoldersResponse>, 'queryKey'> = {}
```
cgero-eth: If I remember correctly we opted to have the `options` parameter as optional without an empty object as default value, correct? Could you update all the queries to follow this pattern?

src/services/aragon-sdk/queries/use-member.ts
```diff
@@ -196,7 +196,10 @@ const fetchMemberDAOs = async (
 
 export const useMember = (
   params: IFetchMemberParams,
-  options: UseQueryOptions<TokenVotingMember | null> = {}
+  options: Omit<
+    UseQueryOptions<TokenVotingMember | null>,
+    'queryKey' | 'queryFn'
```
cgero-eth: Why do we need to omit the `queryFn` parameter here?

src/services/cache.ts
```diff
@@ -22,6 +22,11 @@ import {SmartContract, VerifiedContracts} from 'utils/types';
 export const FOLLOWED_DAOS_KEY = 'favoriteDaos';
 export const PENDING_DAOS_KEY = 'pendingDaos';
 
+export type PageWithDaos = {
+  data: NavigationDao[];
+  total: number;
+};
```
cgero-eth: Aren't `PageWithDaos` and `FollowedDaosResultWithTotal` the same?

src/containers/proposalSnapshot/index.tsx
```diff
@@ -96,7 +96,7 @@ const ProposalSnapshot: React.FC<Props> = ({
       );
     });
 
-  if (proposalsAreLoading || proposalCountIsLoading) {
+  if (proposalsAreLoading && proposalCountIsLoading && !data) {
```
cgero-eth: Why this change? It sounds reasonable to render a loading on proposal or proposal-count loading

src/services/aragon-backend/domain/paginated-response.ts
```diff
@@ -1,6 +1,20 @@
+import {IDao} from 'services/aragon-backend/domain/dao';
 export interface IPaginatedResponse<TItem> {
-  data: TItem[];
+  pages: TItem[];
```
cgero-eth: This interface shouldn't be changed, it represent the data we're getting from the backend. What's the issue in using `IPaginatedResponse<IDao>`?

src/services/aragon-backend/queries/use-daos.ts
```diff
@@ -63,24 +65,27 @@ const fetchDaos = async (
 };
 
 export const useDaos = (
-  params: IFetchDaosParams,
-  options: UseInfiniteQueryOptions<IPaginatedResponse<IDao>> = {}
+  initialParams: IFetchDaosParams,
```
cgero-eth: Let's keep the same names for infinite and normal queries
```suggestion
  params: IFetchDaosParams,
```

src/services/aragon-backend/queries/use-daos.ts
```diff
@@ -63,24 +65,27 @@ const fetchDaos = async (
...
-    aragonBackendQueryKeys.daos(params),
-    ({pageParam}) => fetchDaos({...params, ...pageParam}),
-    {
-      ...options,
-      getNextPageParam: (lastPage: IPaginatedResponse<IDao>) => {
-        const {skip, total, take} = lastPage;
-        const hasNextPage = skip + take < total;
+  return useInfiniteQuery({
+    queryKey: aragonBackendQueryKeys.daos(initialParams),
+    queryFn: ({pageParam = initialParams}) =>
```
cgero-eth: A default value shouldn't be needed, react-query should set this as the `initialPageParam` property
```suggestion
    queryFn: ({pageParam}) =>
```

src/hooks/useFollowedDaos.tsx
```diff
@@ -43,7 +43,6 @@ export const useFollowedDaosQuery = (
   return useQuery<NavigationDao[]>({
     queryKey: ['followedDaos'],
     queryFn: useCallback(() => getFollowedDaosFromCache({skip}), [skip]),
-    select: addAvatarToDaos,
```
cgero-eth, line 46: Why removing this?

src/hooks/useFollowedDaos.tsx
```diff
@@ -243,18 +253,22 @@ export const useRemoveFollowedDaoMutation = (
...
-/**
- * Augment DAOs by resolving the IPFS CID for each DAO's avatar.
- * @param data raw fetched data for the cached DAOs.
- * @returns list of DAOs augmented with the resolved IPFS CID avatars
- */
+// /**
+//  * Augment DAOs by resolving the IPFS CID for each DAO's avatar.
+//  * @param data raw fetched data for the cached DAOs.
+//  * @returns list of DAOs augmented with the resolved IPFS CID avatars
+//  */
```
cgero-eth: Double comments? 🤔

src/hooks/useFollowedDaos.tsx
```diff
@@ -267,11 +281,11 @@ function augmentCachedDaos(
...
-/**
- * Add resolved IPFS CID for each DAO's avatar to the metadata.
- * @param daos array of `NavigationDao` objects representing the DAOs to be processed.
- * @returns array of augmented NavigationDao objects with resolved avatar IPFS CIDs.
- */
+// /**
+//  * Add resolved IPFS CID for each DAO's avatar to the metadata.
+//  * @param daos array of `NavigationDao` objects representing the DAOs to be processed.
+//  * @returns array of augmented NavigationDao objects with resolved avatar IPFS CIDs.
+//  */
```
cgero-eth: Same

src/pages/dashboard.tsx
```diff
@@ -354,7 +387,7 @@ const DashboardContent: React.FC<DashboardContentProps> = ({
   const proposalCount = proposals.length;
   const transactionCount = transfers.length;
 
-  if (isTokensLoading || isLoading) {
+  if (isTokensLoading && isLoading && !data) {
```
cgero-eth: Same here, why changing this logic?

src/pages/dashboard.tsx
```diff
@@ -142,6 +151,37 @@ export const Dashboard: React.FC = () => {
...
   }, [liveDao, daoCreationState, pendingDao]);
 
+  useEffect(() => {
+    // Wait until all data is loaded
+    if (!isLoading) {
+      // Check if there is no liveDao and no pendingDao
+      if (
+        !liveDao &&
+        !pendingDao &&
+        pendingDaoFetched &&
```
cgero-eth: Why is this now needed?

thekidnamedkd: It is currently here: https://github.com/aragon/app/blob/c3dcfb2af2328d41f5f7b9d3048ea166413d76a9/src/pages/dashboard.tsx#L312-L319

This was causing a redirect prematurely in the feature branch after changes with RQ5 seemed to create race conditions where the `(!liveDao && !pendingDao)` check were not "tolerant" enough.

Additionally, navigate() is generally not recommended in the render so moved to a useEffect and added some extra conditions to check against the status of the query itself. Mostly we need to see that it tried to fetch because there are nested async conditions that enable the query with pluginClient for example. 

Per @Fabricevladimir, it is needed generally for UX if we don't get a valid data response with bad addresses out of the params.

src/services/aragon-backend/domain/dao.ts
```diff
@@ -8,6 +8,11 @@ export interface IDaoStats {
   members: number;
 }
 
+export interface IDaoPages {
+  pages: IDao[];
+  pageParams: [];
+}
+
```
cgero-eth: This is no longer used, right?

src/pages/dashboard.tsx
```diff
@@ -456,8 +479,13 @@ const MobileDashboardContent: React.FC<DashboardContentProps> = ({
...
 }) => {
   const {transfers, totalAssetValue, isTokensLoading} = useDaoVault();
+  const {data, isLoading} = useProposals({
+    daoAddressOrEns,
+    pluginType,
+    pluginAddress,
+  });
 
-  if (isTokensLoading) {
+  if (isTokensLoading && isLoading && !data) {
```
cgero-eth: Can we revert these changes as well?

src/pages/proposal.tsx
```diff
@@ -107,7 +107,10 @@ export const Proposal: React.FC = () => {
 
   const {dao, id: proposalId} = useParams();
 
-  const {data: daoDetails, isLoading: detailsAreLoading} = useDaoDetailsQuery();
+  const {
+    data: daoDetails,
+    // isLoading: detailsAreLoading
```
cgero-eth: Can we revert these changes?

thekidnamedkd: what about something like this @cgero-eth?

```ts
/*************************************************
   *                     Render                    *
  const isLoading = paramsAreLoading || proposalIsLoading || detailsAreLoading;

  if (isLoading) {
    return <Loading />;
  }

  // the last check is to make sure Typescript narrows the type properly
  const hasInvalidProposal = proposalError || (proposalIsFetched && !proposal);

  if (hasInvalidProposal) {
    navigate(NotFound, {replace: true, state: {invalidProposal: proposalId}});
    return null;
  }

  if (proposal) {
    // Store voting terminal props
    const votingTerminalProps: VotingTerminalProps = {
      ...mappedProps,
      title: isMultisigProposal(proposal)
        ? t('votingTerminal.multisig.title')
        : isTokenVotingPlugin
        ? t('votingTerminal.title')
        : undefined, // Title will be shown on the GaslessVotingTerminal
      status: proposalStatus,
      pluginType,
      
      (.....cont)
```

Typescript needs to narrow to proposal to avoid null and undefined and the 404 redirects too quickly if just let it check for `proposal == null`

cgero-eth: Mm.. could we instead do something like:
```
if (proposalError || (!isLoading && proposal == null)) {
  // redirect to 404
}

if (isLoading || proposal == null) {
  // return loader
}

[proposal is not null here]
```

cgero-eth: @thekidnamedkd If we invert the error/loading order, we can avoid adding the `if (proposal)` check, right?

cgero-eth (approved):
LGTM ✅

PR #1333, opened 2024-04-22
fix: APP-2956 - Refine session detection for terminated sessions on dAppConnect Actions flow 

thekidnamedkd:
## Description

Fixes race conditions that prevented connected dApps from correctly triggering 'Start adding actions' view

Task: [APP-2956](https://aragonassociation.atlassian.net/browse/APP-2956)

## Type of change

## Checklist:

[APP-2956]: https://aragonassociation.atlassian.net/browse/APP-2956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/walletConnect/dAppValidationModal/index.tsx
+++ b/src/containers/walletConnect/dAppValidationModal/index.tsx
@@ -41,7 +41,6 @@ const DAppValidationModal: React.FC<Props> = props => {
   const [connectionStatus, setConnectionStatus] = useState<ConnectionState>(
     ConnectionState.WAITING
   );
-
   const {wcConnect, validateURI, sessions} = useWalletConnectContext();
 
   const {control} = useFormContext();
@@ -76,8 +75,9 @@ const DAppValidationModal: React.FC<Props> = props => {
     return t('labels.paste');
   }, [connectionStatus, t, uri]);
 
-  const currentSession = sessions.find(
-    ({pairingTopic}) => pairingTopic === sessionTopic
+  const currentSession = useMemo(
+    () => sessions.find(({pairingTopic}) => pairingTopic === sessionTopic),
+    [sessions, sessionTopic]
   );
 
   /*************************************************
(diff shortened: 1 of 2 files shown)


src/containers/walletConnect/dAppValidationModal/index.tsx
```diff
@@ -131,14 +143,16 @@ const DAppValidationModal: React.FC<Props> = props => {
...
 
-  // Reset the connection state if the session has been terminated on the dApp
+  // Reset the connection state if the session has been terminated on the dApp before `Adding actions` view is closed
   useEffect(() => {
-    const isSuccess = connectionStatus === ConnectionState.SUCCESS;
-
-    if (isSuccess && currentSession == null) {
+    if (
+      connectionStatus === ConnectionState.SUCCESS &&
+      latestKnownSession.current &&
```
cgero-eth: What is exactly happening that prevents the application from going to the action-listening view? Introducing `ref` is usually not a good solution

thekidnamedkd: I had some reservations about this approach too. But `isSuccess` is much faster than `currentSession` lookup with find which initializes as `null` so resetConnectedState() runs and short circuits the succesful connect. We never hit the alert and ctaHandler callback

cgero-eth: Okay I see, thanks for the explanation 👌🏻 It might happen that the wallet-connect sessions are not immediately updated, therefore the connection is set to success but the `currentSession` variable is not set yet. 
If that's the case, we can probably simplify this a bit by introducing a `sessionFound` ref like:
```
// Use a React ref to avoid resetting the WalletConnect connection when connection is successful but WalletConnect session is not found yet
const sessionFound = useRef(false);

useEffect(() => {
  sessionFound.current = currentSession != null;
}, [currentSession]);
```
And here:
```
if (isSuccess && sessionFound.current === true && currentSession == null) {
  resetConnectionState();
}
```
What do you think?

cgero-eth (approved):
LGTM, thanks for the fix @thekidnamedkd!

PR #1341, opened 2024-05-08
fix: APP-3119 - Cannot Edit DAO Settings of New DAOs

Barukimang:
## Description

Found that the edit settings proposal flow, and after debugging that, the edit settings transaction would crash because at DAO creation the value `ipfs://undefined` for avatar is given in case no logo is added (this is the case since we have the new transaction flow).

Before that, the value was for avatar was `undefined`, which I again added in case no logo is added at DAO creation.

Edit settings proposals succeed again when changing metadata such as the DAO name, and not editing the DAO logo, see e.g. the proposal [here](http://localhost:5173/#/daos/sepolia/0x3593a8099a96b7caa7b8163962b537ebc3dde786/governance/proposals/0x75de0acf44e2c36e247f13eb2b27bc4135214b61_0x0).

Update: also added a fix for the edge case when a DAO is removing its logo which resulted in a failed create edit proposal transaction before.

Task: [APP-3119](https://aragonassociation.atlassian.net/browse/APP-3119)

## Type of change

## Checklist:

[APP-3119]: https://aragonassociation.atlassian.net/browse/APP-3119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/basic.setup.ts
+++ b/e2e/basic.setup.ts
@@ -5,7 +5,7 @@ import {
 } from '@synthetixio/synpress';
 import 'dotenv/config';
 
-export const LOCALHOST_URL = 'http://localhost:5173/';
+export const LOCALHOST_URL = 'http://localhost:5173';
 
 const SEED_PHRASE = process.env.METAMASK_SEED_PHRASE!;
 const PASSWORD = process.env.METAMASK_PASSWORD!;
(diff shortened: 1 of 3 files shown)


thekidnamedkd (approved):
LGTM! Make sense from the stack trace photos. It expects to return an empty string as fallback ('' aka undefined) not a null. 🚀

PR #1342, opened 2024-05-08
fix: Apply minor fixes for Gasless plugin

emmdim:
## Description

- [x] Fixes missing string upon rejecting a vote transaction
- [x] Fixes ->  "The list of voters on proposals doesn’t have a label showing their true voting option"
- [x] Add a census3 scan check (in addition to subgraph check) to DAO pending screen to not allow user to advance until initial scan has completed.

Minor fixes

- [x] Hide select option after votes
- [x] Fix see holders button is redirecting to mint page
- [x] Prevent empty strategy id
- [x] Add isLoading token on the useCensus3DaoMembers hook

## Type of change

### DIFF
--- a/src/utils/proposals.ts
+++ b/src/utils/proposals.ts
@@ -512,11 +512,6 @@ export function getLiveProposalTerminalProps(
     };
 
     // voters
-    // voters =
-    //   proposal.voters?.map(voter => {
-    //     return {wallet: voter, src: voter} as VoterType;
-    //   }) ?? [];
-
     voters =
       proposal.votes?.map(voter => {
         return {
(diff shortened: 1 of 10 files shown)


src/utils/proposals.ts
```diff
@@ -512,9 +512,19 @@ export function getLiveProposalTerminalProps(
     };
 
     // voters
+    // voters =
+    //   proposal.voters?.map(voter => {
+    //     return {wallet: voter, src: voter} as VoterType;
+    //   }) ?? [];
+
```
cgero-eth:
```suggestion
```

cgero-eth (commented):
Thanks for the fixes @emmdim!

cgero-eth (approved):

PR #1343, opened 2024-05-08
Update to latest Polygon logo

evanaronson:
They've updated their logo since we launched with them last year

## Type of change

## Checklist:

- [X] I have tested the updated logo locally and it looks just fine - transparency in the png all good.

### DIFF
--- a/src/utils/constants/chains.ts
+++ b/src/utils/constants/chains.ts
@@ -132,7 +132,7 @@ export const CHAIN_METADATA: Record<SupportedNetworks, ChainData> = {
     id: 137,
     name: i18n.t('explore.modal.filterDAOs.label.polygon'),
     domain: 'L2 Blockchain',
-    logo: 'https://assets.coingecko.com/coins/images/4713/large/matic-token-icon.png?1624446912',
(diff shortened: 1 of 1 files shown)


src/utils/constants/chains.ts
```diff
@@ -132,7 +132,7 @@ export const CHAIN_METADATA: Record<SupportedNetworks, ChainData> = {
     id: 137,
     name: i18n.t('explore.modal.filterDAOs.label.polygon'),
     domain: 'L2 Blockchain',
-    logo: 'https://assets.coingecko.com/coins/images/4713/large/matic-token-icon.png?1624446912',
+    logo: 'https://assets.coingecko.com/coins/images/4713/large/polygon.png',
```
thekidnamedkd, line 135: are we sure this is ~LARGE~ enough?

thekidnamedkd (approved):
LGTM! TYSM! 🚀

src/utils/constants/chains.ts
evanaronson: It's the same size as the previous one. Looks fine in app.

PR #1357, opened 2024-05-24
fix: APP-3244 - Fix proposal filters on governance page

Barukimang:
## Description

Filters working again, test on [this DAO](http://localhost:5173/#/daos/polygon/0xcd0c49d4de6caf086350549753cfc4f4accb1031/governance) eg. Also see images below 👇 

Image 1
image

Image 2
image

Image 3
image

Task: [APP-3244](https://aragonassociation.atlassian.net/browse/APP-3244)

## Type of change

## Checklist:

[APP-3244]: https://aragonassociation.atlassian.net/browse/APP-3244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/pages/governance.tsx
+++ b/src/pages/governance.tsx
@@ -46,15 +46,6 @@ export const Governance: React.FC = () => {
     status: filter !== 'All' ? filter : undefined,
   });
 
-  const filteredProposals = () => {
-    if (filter === 'All') {
-      return data?.pages.flat() ?? [];
-    }
-    return (data?.pages.flat() ?? []).filter(
-      proposal => proposal.status === filter
-    );
-  };
-
   const noProposals =
     isFetched && data?.pages[0].length === 0 && filter === 'All';
 
@@ -171,7 +162,7 @@ export const Governance: React.FC = () => {
             toDisplayEns(daoDetails?.ensDomain) ||
             (daoDetails?.address as string)
           }
-          proposals={filteredProposals()}
+          proposals={data?.pages.flat() ?? []}
           pluginAddress={pluginAddress as string}
           pluginType={pluginType as PluginTypes}
           isLoading={proposalsLoading}
--- a/src/pages/governance.tsx
+++ b/src/pages/governance.tsx
@@ -46,6 +46,15 @@ export const Governance: React.FC = () => {
     status: filter !== 'All' ? filter : undefined,
   });
 
+  const filteredProposals = () => {
+    if (filter === 'All') {
+      return data?.pages.flat() ?? [];
+    }
+    return (data?.pages.flat() ?? []).filter(
+      proposal => proposal.status === filter
+    );
+  };
+
   const noProposals =
     isFetched && data?.pages[0].length === 0 && filter === 'All';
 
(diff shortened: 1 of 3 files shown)


src/pages/governance.tsx
```diff
@@ -46,6 +46,15 @@ export const Governance: React.FC = () => {
...
   });
 
+  const filteredProposals = () => {
+    if (filter === 'All') {
+      return data?.pages.flat() ?? [];
+    }
+    return (data?.pages.flat() ?? []).filter(
+      proposal => proposal.status === filter
+    );
+  };
```
cgero-eth: Previously we used to apply the proposal status filter directly to Subgraph, in that way Subgraph only gave us the succeeded, active, pending.. proposals. Looks like the recent changes introduced for the IPFS fixes removed this filter. This solution applies the filter on the client side, which work for some cases but fails when, for instance, we fetch only the first 20 proposals from Subgraph, all of these have a "pending" state and we apply a "succeeded" filter on the client. In this case, we'll show no proposals when we might have some on Subgraph.
To properly fix this, we should update the `useProposals` hook to apply the correct filter on the Subgraph query depending on the selected proposal state (this has been done on the SDK through the `computeProposalsStatusFilter` functions for the [multisig](https://github.com/aragon/sdk/blob/develop/modules/client/src/multisig/internal/utils.ts#L162) and [tokenVoting](https://github.com/aragon/sdk/blob/develop/modules/client/src/tokenVoting/internal/utils.ts#L339) plugins)

Barukimang: Submitted 2 commits @cgero-eth 👇 removing the 'client based fix' and adding a 'sdk based fix', which is working. Found some issues however, which might already have been there but not noticed:

-Token Voting proposal filters:
1. Succeeded filter: does not work since the `or` condition is failing the query, have not been able to debug:


```
where = {
        or: [{approvalReached: true, endDate_lt: now}, {earlyExecutable: true}],
      };
```
      

2. Defeated filter: does not work since `potentiallyExecutable` cannot be queried, whether it is true or false:

```
where = {
        potentiallyExecutable: false,
        endDate_lt: now,
        executed: false,
      };
```

-Multisig proposal filters:
1. Succeeded filter: there is no `Succeeded` status for Multisig proposals, but `Approved` instead. Also does not work since the `or` condition is failing the query, have not been able to debug:

```
where = {
        or: [
          {approvalReached: true, endDate_lt: now, isSignaling: false},
          {approvalReached: true, isSignaling: true},
        ],
      };
```
      

2. Defeated filter: also brings `Approved` proposals:

```
where = {
        endDate_lt: now,
        executed: false,
      };
```

So again, think these have been existing issues (confirmed them on an old web preview), can make a separate ticket for it, wdyt?

cgero-eth: Thanks for the changes, yes given that these are existing issues we can move them into another ticket (found a Discord message related to it => https://discord.com/channels/672466989217873929/950422827893420153/1187351860462506024) 👍🏻

cgero-eth (approved):
LGTM, thanks @Barukimang!

PR #1384, opened 2024-07-08
fix(APP-3092): Truncate DAO links on settings page

Barukimang:
## Description

Before (find DAO [here](http://localhost:5173/#/daos/polygon/0xbf3460589097676656343a9e44d5a83e54e081ad/settings))
image

After
image

Task: [APP-3092](https://aragonassociation.atlassian.net/browse/APP-3092)

## Type of change

## Checklist:

[APP-3092]: https://aragonassociation.atlassian.net/browse/APP-3092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/@aragon/ods-old/components/link/link.tsx
+++ b/src/@aragon/ods-old/components/link/link.tsx
@@ -81,7 +81,7 @@ const Label = styled.span.attrs({
 })``;
 
 const Description = styled.p.attrs({
-  className: 'ft-text-sm text-neutral-600 truncate',
+  className: 'ft-text-sm text-neutral-600 truncate max-w-md',
 })``;
 
 const variants = {
--- a/src/@aragon/ods-old/components/link/link.tsx
+++ b/src/@aragon/ods-old/components/link/link.tsx
@@ -81,7 +81,7 @@ const Label = styled.span.attrs({
 })``;
 
 const Description = styled.p.attrs({
-  className: 'ft-text-sm text-neutral-600 truncate max-w-md',
+  className: 'ft-text-sm text-neutral-600 truncate',
 })``;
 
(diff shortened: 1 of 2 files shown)


src/@aragon/ods-old/components/link/link.tsx
```diff
@@ -81,7 +81,7 @@ const Label = styled.span.attrs({
 })``;
 
 const Description = styled.p.attrs({
-  className: 'ft-text-sm text-neutral-600 truncate',
+  className: 'ft-text-sm text-neutral-600 truncate max-w-md',
```
cgero-eth: Similar as the other PR, we shouldn't set a max-width globally for every link of the application.
Usually, when a text does not truncate inside a flexbox, the solution is to set a `minWidth: 0` on the flexbox container, you can read more here https://css-tricks.com/flexbox-truncated-text/#aa-what-we-want

Barukimang: Thanks for the comment @cgero-eth! Removed the global `Link` style and added local styling on the flexbox container keeping the `Link` element. Wanted to solve it with Tailwind instead of classic CSS, so tried with `min-w-0` which didn't work, so used `max-w-md` instead. See image below:


Image
image

Barukimang:
Added breaking point so to only repair at xl screens where this was an issue. It is fixed now for all screen sizes, also bit more square sizes (see image below):

Image
image

cgero-eth (approved):
LGTM, thanks @Barukimang!

